PR #6, opened 2021-12-07
Remove core dep

connorwstein:


go.mod
```diff
@@ -1,12 +1,10 @@
-module github.com/smartcontractkit/solana-integration
+module github.com/smartcontractkit/chainlink-solana
 
 go 1.17
 
 require (
-	github.com/ethereum/go-ethereum v1.10.11
 	github.com/gagliardetto/binary v0.5.0
 	github.com/gagliardetto/solana-go v1.0.2
-	github.com/smartcontractkit/chainlink v1.0.1-0.20211130190746-c1f78221a5e7
```
RyanRHall, line 9: :+1:

pkg/solana/relay.go
```diff
@@ -57,6 +58,11 @@ func (r relayer) Healthy() error {
 	return nil
 }
 
+type Transmitter interface {
+	Sign(msg []byte) ([]byte, error)
+	PublicKey() solana.PublicKey
+}
+
```
RyanRHall, line 65: :+1:

RyanRHall (approved):

aalu1418 (approved):
LGTM 🚀

pkg/solana/relay.go
krebernisak: Should be `Signer` as it just signs and has an id, doesn't transmit anything.

Np now, we'll tweak things like this at a later time.

pkg/solana/relay.go
```diff
@@ -65,31 +71,25 @@ type OCR2Spec struct {
...
 
-	// on-chain program + 2x state accounts (state + transmissions)
-	ProgramID       solana.PublicKey
-	StateID         solana.PublicKey
-	TransmissionsID solana.PublicKey
+	// on-chain program + 2x state accounts (state + transmissions) + validator program
+	ProgramID          solana.PublicKey
+	StateID            solana.PublicKey
+	ValidatorProgramID solana.PublicKey
+	TransmissionsID    solana.PublicKey
```
krebernisak, line 78: Isn't this available from the `Transmitter` interface as `t.PublicKey()`?

pkg/solana/transmitter.go
```diff
@@ -78,15 +73,17 @@ func (c ContractTracker) Transmit(
...
+		return err
 	}
-	if _, err = tx.Sign(pkGetter); err != nil {
-		return errors.Wrap(err, "error on transaction sign")
+	finalSigBytes, err := c.Transmitter.Sign(msgToSign)
+	if err != nil {
+		return err
 	}
+	var finalSig [64]byte
+	copy(finalSig[:], finalSigBytes)
```
krebernisak, line 85: What's the purpose of this final copy?

pkg/solana/relay.go
aalu1418: @krebernisak not sure i understand your question...
`TransmissionsID` refers to the account where the transmissions state is stored...`Transmitter` is the account sending the report tx.

krebernisak: Ah, yes...

krebernisak: Was thinking about the `Transmitter` interface name, and then this confused me. Thx!

pkg/solana/relay.go
krebernisak: Cosmos SDK interface example: https://github.com/cosmos/cosmos-sdk/blob/779c7e2e174b1dff4f71329a5662c4b8806d36dd/crypto/keyring/keyring.go#L109

PR #10, opened 2021-12-08
Onchain monitoring 

topliceanu:


cmd/monitoring/main.go
```diff
@@ -0,0 +1,66 @@
...
+	go func() {
+		if err := http.ListenAndServe(cfg.Http.Address, nil); err != nil {
+			log.Fatalf("failed to start http server with address %s: error %v", cfg.Http.Address, err)
+		}
+	}()
+
+	client := rpc.New(cfg.Solana.RPCEndpoint)
+
+	schemaRegistry := monitoring.NewSchemaRegistry(cfg.SchemaRegistry)
+	trSchema, err := schemaRegistry.EnsureSchema("transmission", monitoring.TransmissionAvroSchema)
```
pedrovsg, line 35: subject should be topic name + value https://github.com/smartcontractkit/gokit/blob/ca2df4d6266ca7a27cb8ba92a29a247d150778c9/kafka/schema_registry.go#L29

monitoring.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENV PATH /usr/local/go/bin:$PATH
+
+# Compile binary
+
+WORKDIR /solana-monitoring
+RUN go build -o ./monitoring ./cmd/monitoring/*.go
+
+# Production image
+
+FROM ubuntu:20.04
```
pedrovsg, line 30: consider using alpine

cmd/monitoring/main.go
topliceanu: Got it. Will update!

pkg/monitoring/config.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if value, isPresent := os.LookupEnv("SCHEMA_REGISTRY_PASSWORD"); isPresent {
+		cfg.SchemaRegistry.Password = value
+	}
+
+	var feeds = []jsonFeedConfig{}
+	if feedsFilePath == "" && feedsRDDURL == "" {
+		return cfg, fmt.Errorf("feeds configuration missing, either '-feeds.file_path' or '-feeds.rdd_url' must be set")
+	} else if feedsRDDURL != "" {
+		res, err := http.Get(feedsRDDURL)
```
pedrovsg, line 140: we shouldn't use default http cause does not have a timeout.

cmd/monitoring/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	monitor := monitoring.NewMultiFeedMonitor(
+		cfg.Solana,
+		trReader, stReader,
+		trSchema, stSchema,
+		producer,
+		cfg.Feeds,
+		monitoring.DefaultMetrics,
+	)
+	go monitor.Start(ctx)
```
samsondav, line 60: Can main exit before the goroutine terminates?

monitoring.Dockerfile
samsondav: Why use alpine?

samsondav (approved):

pkg/monitoring/config.go
topliceanu: Agreed

cmd/monitoring/main.go
topliceanu: It can. I was thinking a WaitGroup could help. Wdyt?

monitoring.Dockerfile
topliceanu: Main reason would be size. The image now is ~100mb. 
I made a note for this, maybe for mainnet. For now, I think it's fine to stick to ubuntu!

cmd/monitoring/README.md
```diff
@@ -0,0 +1,75 @@
+# Solana Onchain monitoring
+
+
+## Run locally
+
+The monitor requires a schema registry, a kafka instance and a zookeper.
+You can use the docker-compose.yml file in `smartcontractkit/gokit` to start these up.
```
krebernisak, line 7: `smartcontractkit/gokit` is still internal, so perhaps we could copy the `docker-compose.yml` over here under `/ops`, or add an option (implement requirements) to our Pulumi local env, or just make the `gokit` public.

pkg/monitoring/poller.go
```diff
@@ -0,0 +1,70 @@
...
+		account,
+		reader,
+		fetchInterval,
+		bufferCapacity,
+		make(chan interface{}, bufferCapacity),
+	}
+}
+
+type solanaPollerImpl struct {
+	account        solana.PublicKey
```
krebernisak, line 32: This is the only part that makes this `Poller` implementation Solana specific, but we could easily generalize it. Consider just using string for account identifier:

```suggestion
type poller struct {
	account        string
```

pkg/monitoring/poller.go
```diff
@@ -0,0 +1,70 @@
...
+	bufferCapacity uint32
+	updates        chan interface{}
+}
+
+// Start should be executed as a goroutine
+func (s *solanaPollerImpl) Start(ctx context.Context) {
+	// Fetch initial data
+	data, err := s.reader.Read(ctx, s.account)
+	if err != nil {
+		log.Printf("failed initial read of the account contents for address '%s': %v", s.account.String(), err)
```
krebernisak, line 44: Please use the relay `Logger` interface instead directly `log.Printf`

pkg/monitoring/producer.go
```diff
@@ -0,0 +1,65 @@
+package monitoring
```
krebernisak, line 1: This and many files in this implementation are currently or are close to being Solana agnostic.

As this monitoring service will be useful across networks and across protocols (contracts) I'd suggest as a first step we try to decouple the two best we can ATM.

- `pkg/monitoring` - generic monitoring code
- `pkg/monitoring/solana/...` - solana specific implementations and details

Then we iterate on the monitoring interfaces couple of times until we get a fit, and we can raise the generic part to the Relay SDK level, which all relays use to implement and bootstrap their monitoring serivce.

monitoring.Dockerfile
pedrovsg: Security is also a reason, alpine is a minimal distribution with less dependencies.

pkg/monitoring/config.go
topliceanu: Updated!

cmd/monitoring/main.go
topliceanu: Updated

cmd/monitoring/README.md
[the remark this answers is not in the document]
topliceanu: I went for copying docker-compose from gokit. First, it's only for development. Second, gokit is not going to be made public. Third, I don't know pulumi!

pkg/monitoring/account_reader.go
```diff
@@ -0,0 +1,69 @@
...
+	State       pkgSolana.State
+	BlockNumber uint64
+}
+
+func (s *stReader) Read(ctx context.Context, stateAccount solana.PublicKey) (interface{}, error) {
+	ctx, cancel := context.WithTimeout(ctx, solanaAccountReadTimeout)
+	defer cancel()
+	res, err := s.client.GetAccountInfo(ctx, stateAccount)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch state account: %w", err)
```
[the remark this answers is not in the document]
topliceanu: I was under the impression that `%w` is the recommended way to warp errors in golang. The package you're referring to - [errors](https://pkg.go.dev/github.com/pkg/errors#section-readme) - is not in the stdlib and I'm reluctant to add is as a dependency! Why it better?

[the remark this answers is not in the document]
krebernisak: Just for convenience and consistency in usage and generating error reports. I know it's used throughout CL core repo: https://github.com/smartcontractkit/chainlink/search?q=errors.Wrap

pkg/monitoring/account_reader.go
```diff
@@ -0,0 +1,69 @@
...
+
+type StateEnvelope struct {
+	State       pkgSolana.State
+	BlockNumber uint64
+}
+
+func (s *stReader) Read(ctx context.Context, stateAccount solana.PublicKey) (interface{}, error) {
+	ctx, cancel := context.WithTimeout(ctx, solanaAccountReadTimeout)
+	defer cancel()
+	res, err := s.client.GetAccountInfo(ctx, stateAccount)
```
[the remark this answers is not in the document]
topliceanu: Done

pkg/monitoring/poller.go
topliceanu: Updated

cmd/monitoring/main.go
```diff
@@ -0,0 +1,128 @@
...
+		monitor.Start(bgCtx, wg)
+	}()
+
+	osSignalsCh := make(chan os.Signal, 1)
+	signal.Notify(osSignalsCh, syscall.SIGINT, syscall.SIGTERM)
+	sig := <-osSignalsCh
+	log.Infof("received signal '%v'. Stopping", sig)
+
+	cancelBgCtx()
+	if err := server.Shutdown(bgCtx); err != nil && err != context.Canceled {
```
kaleofduty, line 97: I'd suggest using https://pkg.go.dev/errors#Is here

pkg/monitoring/account_reader.go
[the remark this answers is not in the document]
topliceanu: Got it! Will do

cmd/monitoring/main.go
topliceanu: Ah good point. I forgot about that! Updated!

PR #14, opened 2021-12-10
Contracts: Update account memory layout

archseer:
`repr(packed)` will soon be swapped for `repr(C)` and we need to ensure there's no padding gaps in the types.


pkg/solana/config_tracker.go
```diff
@@ -18,27 +18,31 @@ func (c ContractTracker) LatestConfigDetails(ctx context.Context) (changedInBloc
...
-// LatestConfig returns the latest configuration.
-func (c ContractTracker) LatestConfig(ctx context.Context, changedInBlock uint64) (types.ContractConfig, error) {
-	err := c.fetchState(ctx)
-
+func configFromState(state State) types.ContractConfig {
 	pubKeys := []types.OnchainPublicKey{}
 	accounts := []types.Account{}
-	for _, o := range c.state.Oracles.Data() {
+	for _, o := range state.Oracles.Data() {
+		o := o //  https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
```
archseer, line 25: @aalu1418 this was the fix

pkg/solana/types.go
```diff
@@ -27,11 +27,13 @@ const (
 // State is the struct representing the contract state
 type State struct {
 	AccountDiscriminator [8]byte // first 8 bytes of the SHA256 of the account’s Rust ident, https://docs.rs/anchor-lang/0.18.2/anchor_lang/attr.account.html
+	Version              uint8
 	Nonce                uint8
+	Padding0             uint16
+	Padding1             uint32
```
krebernisak, line 33: Consider something like this, as the paddings are unused nd just reserve memory:
```suggestion
	_ /* _padding0 */    uint16
	_ /* _padding0 */    uint32
```

pkg/solana/types_test.go
```diff
@@ -37,46 +37,41 @@ func TestState_Decode(t *testing.T) {
 
 var mockRawState = []byte{
```
krebernisak, line 38: Consider declaring the config digest for this mocked state so we can reference it in tests from one place:

```go
var mockState :=   struct {
        Raw                          []byte
        ConfigDigestHex    string
    }{
        Raw:                          []byte{...},
        ConfigDigestHex:    "00039ad1eaa13649831b4cc02f5437bf556910037c6833f87249edbeceab2828",
    }
```

Or even declare a type if we'll have more than one test example.

pkg/solana/types.go
```diff
@@ -61,16 +71,17 @@ type Config struct {
 	BillingAccessController   solana.PublicKey
 	MinAnswer                 bin.Int128
 	MaxAnswer                 bin.Int128
-	Decimals                  uint8
 	Description               [32]byte
+	Decimals                  uint8
 	F                         uint8
+	Round                     uint8
+	Padding0                  uint8
```
krebernisak, line 78:
```suggestion
	_ /* _padding0 */         uint8
```

krebernisak (approved):
Great to see contracts cleanly separated, and awesome job finding the bug 💯

pkg/solana/types.go
archseer: I'm not sure this will work since the `binary` encoding library can only see public fields

krebernisak: [The docs](https://golang.google.cn/pkg/encoding/binary) say it should possibly be ok:

> When reading into structs, the field data for fields with blank (_) field names is skipped; i.e., blank field names may be used for padding. When reading into a struct, all non-blank fields must be exported or Read may panic.

> When writing structs, zero values are written for fields with blank (_) field names.

aalu1418: We're using a binary library that is used in `solana-go` https://github.com/gagliardetto/binary
In this case, it will skip `_` parameters

aalu1418 (approved):

PR #20, opened 2021-12-10
Gauntlet inspection commands

RodrigoAD:
Added inspect aggregator command. 
Inspect responses is indicated only. We should come back when we have a contract example with valid responses

### DIFF
--- a/gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
+++ b/gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
@@ -1,0 +1,103 @@
+import { Result } from '@chainlink/gauntlet-core'
+import { SolanaCommand, TransactionResponse } from '@chainlink/gauntlet-solana'
+import { PublicKey } from '@solana/web3.js'
+import { CONTRACT_LIST, getContract } from '../../../../lib/contracts'
+import { getRDD } from '../../../../lib/rdd'
+import {
+  inspect,
+  Inspection,
+  makeInspection,
+  toComparableNumber,
+  toComparablePubKey,
+} from '../../../../core/inspection'
+
+type Input = {
+  description: string
+  decimals: string | number
+  minAnswer: string | number
+  maxAnswer: string | number
+  transmitters: string[]
+  billingAccessController: string
+  requesterAccessController: string
+  link: string
+}
+
+export default class OCR2Inspect extends SolanaCommand {
+  static id = 'ocr2:inspect'
+  static category = CONTRACT_LIST.OCR_2
+
+  makeInput = (userInput): Input => {
+    if (userInput) return userInput as Input
+    const rdd = getRDD(this.flags.rdd)
+    const info = rdd.contracts[this.flags.state]
+    const aggregatorOperators: string[] = info.oracles.map((o) => o.operator)
+    const transmitters = aggregatorOperators.map((operator) => rdd.operators[operator].nodeAddress[0])
+    const billingAccessController =
+      this.flags.billingAccessController || process.env.BILLING_ADMIN_ACCESS_CONTROLLER_ADDRESS
+    const requesterAccessController =
+      this.flags.requesterAccessController || process.env.REQUESTER_ADMIN_ACCESS_CONTROLLER_ADDRESS
+    const link = this.flags.link || process.env.LINK_ADDRESS
+    return {
+      description: info.name,
+      decimals: info.decimals,
+      minAnswer: info.minSubmissionValue,
+      maxAnswer: info.maxSubmissionValue,
+      transmitters,
+      billingAccessController,
+      requesterAccessController,
+      link,
+    }
+  }
+
+  constructor(flags, args) {
+    super(flags, args)
+    this.require(!!this.flags.state, 'Please provide flags with "state""')
+  }
+
+  execute = async () => {
+    const ocr2 = getContract(CONTRACT_LIST.OCR_2, '')
+    const program = this.loadProgram(ocr2.idl, ocr2.programId.publicKey.toString())
+
+    const state = new PublicKey(this.flags.state)
+    const input = this.makeInput(this.flags.input)
+    const data = await program.account.state.fetch(state)
+
+    const inspections: Inspection[] = [
+      makeInspection(toComparableNumber(data.config.minAnswer), toComparableNumber(input.minAnswer), 'Min Answer'),
+      makeInspection(toComparableNumber(data.config.maxAnswer), toComparableNumber(input.maxAnswer), 'Max Answer'),
+      makeInspection(toComparableNumber(data.config.decimals), toComparableNumber(input.decimals), 'Decimals'),
+      makeInspection(
+        // Description comes with some empty bytes
+        Buffer.from(data.config.description.filter((v) => v !== 0)).toString(),
+        input.description,
+        'Description',
+      ),
+      makeInspection(
+        toComparablePubKey(data.config.requesterAccessController),
+        toComparablePubKey(input.requesterAccessController),
+        'Requester access controller',
+      ),
+      makeInspection(
+        toComparablePubKey(data.config.billingAccessController),
+        toComparablePubKey(input.billingAccessController),
+        'Requester access controller',
+      ),
+      makeInspection(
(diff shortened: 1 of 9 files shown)


gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -0,0 +1,103 @@
...
+        'Requester access controller',
+      ),
+      makeInspection(
+        data.oracles.xs.slice(0, data.oracles.len).map(({ transmitter }) => toComparablePubKey(transmitter)),
+        input.transmitters.map(toComparablePubKey),
+        'Transmitters',
+      ),
+    ]
+
+    const successfulInspection = inspect(inspections)
```
RodrigoAD, line 92: `isSuccessfulInspection` better

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.require(!!this.flags.state, 'Please provide flags with "state""')
+  }
+
+  execute = async () => {
+    const ocr2 = getContract(CONTRACT_LIST.OCR_2, '')
+    const program = this.loadProgram(ocr2.idl, ocr2.programId.publicKey.toString())
+
+    const state = new PublicKey(this.flags.state)
+    const input = this.makeInput(this.flags.input)
+    const data = await program.account.state.fetch(state)
```
topliceanu, line 63: Can't this fetch generate an error? Where is it handled? 
Also, do we maybe need to retry in case of connection errors? Or do you just invoke gauntlet again

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const inspections: Inspection[] = [
+      makeInspection(toComparableNumber(data.config.minAnswer), toComparableNumber(input.minAnswer), 'Min Answer'),
```
topliceanu, line 66: So this logic compares what's in the RDD with what's in the program's configuration onchain, right? 
I suggest you make these variable names more readable, eg. `data -> onChainState`, `input -> rddFeedConfig`. Wdyt?

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -0,0 +1,103 @@
...
+    return {
+      description: info.name,
+      decimals: info.decimals,
+      minAnswer: info.minSubmissionValue,
+      maxAnswer: info.maxSubmissionValue,
+      transmitters,
+      billingAccessController,
+      requesterAccessController,
+      link,
+    }
```
topliceanu, line 49: This is a subset of the config stored onchain, which is also present in the RDD, right? 
I'm asking because there is a lot more stuff onchain that we could be checking on. Eg. https://github.com/smartcontractkit/offchain-reporting/blob/d6773dfde7b48d814d7d8dec1a0e85ad8d6eb01a/lib/offchainreporting2/internal/config/offchainreporting2_offchain_config.proto or https://github.com/smartcontractkit/offchain-reporting/blob/d6773dfde7b48d814d7d8dec1a0e85ad8d6eb01a/lib/offchainreporting2/reportingplugin/median/offchainreporting2_median_config.proto

gauntlet/packages/gauntlet-solana-contracts/src/core/assert.ts
```diff
@@ -0,0 +1,18 @@
...
+  return isEqual(value, expected)
+}
+
+export const isEqual = (a, b) => new String(a).toString() === new String(b).toString()
+
+export function isEqualArrays(a, b) {
+  a = a.map((x) => JSON.stringify(x))
+  b = b.map((y) => JSON.stringify(y))
+
+  const intersection = a.filter((x) => b.includes(x))
```
topliceanu, line 12: Nice! TIL

gauntlet/packages/gauntlet-solana-contracts/src/core/assert.ts
```diff
@@ -0,0 +1,18 @@
...
+  const difference = arrayDifference(a, b)
+  return a.length === b.length && intersection.length === a.length && difference.length === 0
+}
+
+export const arrayDifference = (a: any[], b: any[]) =>
+  a.filter((x) => !b.includes(x)).concat(b.filter((x) => !a.includes(x)))
```
topliceanu, line 18: Wow! This is not very readable!

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
RodrigoAD: That's correct, but it compares it with the input provided. By default it falls to the values from RDD, but is not necessary.
`data -> onChainState` I think makes sense

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
RodrigoAD: Yes you're right, I'll add the offchain config validation

gauntlet/packages/gauntlet-solana-contracts/src/core/assert.ts
RodrigoAD: haha you are right, is not. Copied from old gauntlet, let me make it more readable

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
RodrigoAD: Anything that errors during a Gauntlet execution is caught at a higher level. You'd need to run it again in that case

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspectResponses.ts
```diff
@@ -0,0 +1,46 @@
...
+    const ocr2 = getContract(CONTRACT_LIST.OCR_2, '')
+    const program = this.loadProgram(ocr2.idl, ocr2.programId.publicKey.toString())
+
+    const state = new PublicKey(this.flags.state)
+    const input = this.makeInput(this.flags.input)
+    const data = await program.account.state.fetch(state)
+
+    const inspections: Inspection[] = []
+
+    const successfulInspection = inspect(inspections)
```
aalu1418, line 35: correct me if i'm wrong, but `ocr2:inspect:responses` doesn't look like it does anything?

for EVM i think responses are usually fetched from events, but Solana doesn't have the same thing...
additionally, when the report is put on-chain it only has the median answer (no set of responses from each node)

RodrigoAD: Yeah, for now I left it indicated only until we have some feed running, so I have an example with valid transmissions

topliceanu (approved):
lgtm

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspectResponses.ts
aalu1418: If you want something preliminary, you could try using the state from: https://github.com/smartcontractkit/chainlink-solana/blob/feature/relay/pkg/solana/contract_test.go#L70

it's from one of my local test runs

aalu1418: up to you, otherwise it might be worth leaving a TODO in a comment :)

RodrigoAD: Created a story for it. Will leave it for later

PR #26, opened 2021-12-13
Gauntlet fix offchain serialize

RodrigoAD:
Old JSON descriptor was generating a different result than libocr. Now is autogenerated from the `.proto` description


gauntlet/packages/gauntlet-solana-contracts/src/core/proto/ocr2Proto.ts
```diff
@@ -0,0 +1,45 @@
+// Autogenerated from ocr2Proto.proto with Protobuf.toJSON()
+export const descriptor = {
```
RodrigoAD, line 2: Autogenerated descriptor with https://github.com/smartcontractkit/chainlink-solana/blob/8bb813d26e1ab55e54fbf6b938c33417685c86cd/gauntlet/packages/gauntlet-solana-contracts/src/core/proto/ocr2Proto.proto as source
JSON descriptor is still used as TS transpilation does not allow to include `.proto` files, that will be needed for the binary

krebernisak: Let's add a `yarn` script that will do the auto-generation and write the descriptor to the file so there are no manual steps involved.

gauntlet/packages/gauntlet-solana-contracts/src/core/proto/ocr2Proto.proto
```diff
@@ -0,0 +1,37 @@
+syntax="proto3";
```
krebernisak, line 1: Where is this file from? It will also be required for the monitoring service AFAIK cc @topliceanu

Let's move this to the root e.g. `./proto/ocr2.proto`

gauntlet/packages/gauntlet-solana-contracts/src/core/proto/ocr2Proto.ts
```diff
@@ -0,0 +1,45 @@
+// Autogenerated from ocr2Proto.proto with Protobuf.toJSON()
```
krebernisak, line 1: Let's rename the file as `ocr2.proto.ts` and move it from `./core` as these are contracts/ocr2 specific, not part of core lib.

gauntlet/packages/gauntlet-solana-contracts/src/core/proto/ocr2Proto.proto
RodrigoAD: From libocr. Monitoring uses the same, but does not include: https://github.com/smartcontractkit/chainlink-solana/pull/26/files#diff-832dbc2305017d852ec2b4160c3b9d4f18443f44fd1696e6d8dced401c949722R31 
which gauntlet needs

krebernisak: Yeah, let's just move it to root and monitoring can use directly while Gauntlet derives/generates the `ts` descriptors by using the local script.

PR #33, opened 2021-12-14
Fix: transmission decoding

aalu1418:
Previously, timestamp was a u32, but the latest program uses a u64. This PR fixes the offchain side for proper answer + timestamp parsing

Bonus:
* updated ocr2 schema for gauntlet related to upstream contract updates
* fix transmitter context + error logging
  * go func context should be separate from `transmit` context. otherwise the end of the function cancels the go func call

Currently testing... ✅ works!


krebernisak (approved):
LGTM! 👍

PR #39, opened 2021-12-15
Add configsetsimplified

topliceanu:


pkg/monitoring/config.go
```diff
@@ -30,13 +30,15 @@ type SolanaConfig struct {
...
-	Topic            string `json:"topic,omitempty"`
+	Brokers                  string `json:"brokers,omitempty"`
+	ClientID                 string `json:"client_id,omitempty"`
+	SecurityProtocol         string `json:"security_protocol,omitempty"`
+	SaslMechanism            string `json:"sasl_mechanism,omitempty"`
+	SaslUsername             string `json:"sasl_username,omitempty"`
+	SaslPassword             string `json:"sasl_password,omitempty"`
+	TransmissionTopic        string `json:"kafka_transmission_topic,omitempty"`
+	ConfigSetTopic           string `json:"kafka_config_set_topic,omitempty"`
+	ConfigSetSimplifiedTopic string `json:"kafka_config_set_simplified_topic,omitempty"`
```
topliceanu, line 41: you don't need the `kafka_` prefix here. These field will be in an object under the `kafka` key anyway! See https://github.com/smartcontractkit/chainlink-solana/blob/feature/relay/pkg/monitoring/config.go#L17

pkg/monitoring/config.go
```diff
@@ -85,7 +87,9 @@ func ParseConfig(ctx context.Context) (Config, error) {
 	flag.StringVar(&cfg.Solana.NetworkID, "solana.network_id", "", "")
 	flag.StringVar(&cfg.Solana.ChainID, "solana.chain_id", "", "")
 
-	flag.StringVar(&cfg.Kafka.Topic, "kafka.topic", "", "")
+	flag.StringVar(&cfg.Kafka.ConfigSetTopic, "kafka.config.set.topic", "", "")
+	flag.StringVar(&cfg.Kafka.ConfigSetSimplifiedTopic, "kafka.config.set.simplified.topic", "", "")
+	flag.StringVar(&cfg.Kafka.TransmissionTopic, "kafka.transmission.topic", "", "")
```
topliceanu, line 92: Use underscores like I described above!

pkg/monitoring/feed_monitor.go
```diff
@@ -66,70 +68,98 @@ func (f *feedMonitor) Start(ctx context.Context) {
...
-			f.log.Errorw("failed to publish to kafka", "message", mapping, "error", err)
-			continue
-		}
-		// Publish metrics to prometheus
-		switch typed := update.(type) {
+			err = f.processState(typed)
+			if err != nil {
+				break
+			}
+			err = f.processConfigSetSimplified(typed)
```
topliceanu, line 78: Just a not here, in this code, if `processState()` fails, then you don't do `processConfigSetSimplified()` or export to prometheus. This shouldn't be the case, each of these three pieces of functionality should be attempted in parallel. 

Let's do this in a later update!

pkg/monitoring/mapping_test.go
```diff
@@ -219,10 +221,52 @@ func TestMapping(t *testing.T) {
...
+		require.Equal(t, configSetSimplified["delta_round"], uint64ToBeBytes(offchainConfig.DeltaRoundNanoseconds))
+		require.Equal(t, configSetSimplified["delta_grace"], uint64ToBeBytes(offchainConfig.DeltaGraceNanoseconds))
+		require.Equal(t, configSetSimplified["delta_stage"], uint64ToBeBytes(offchainConfig.DeltaStageNanoseconds))
+		require.Equal(t, configSetSimplified["r_max"], int64(offchainConfig.RMax))
+		require.Equal(t, configSetSimplified["f"], int32(state.Config.F))
+		require.Equal(t, configSetSimplified["signers"], jsonMarshalToString(t, extractSigners(state.Oracles)))
+		require.Equal(t, configSetSimplified["transmitters"], jsonMarshalToString(t, extractTransmitters(state.Oracles)))
+		require.Equal(t, configSetSimplified["s"], jsonMarshalToString(t, offchainConfig.S))
+		require.Equal(t, configSetSimplified["oracles"], string(oracles))
+		require.Equal(t, configSetSimplified["feed_state_account"], strings.Replace(jsonMarshalToString(t, feedConfig.StateAccount.Bytes()), "\"", "", -1))
```
topliceanu, line 257: You're checking various json-encoded strings in this assertion block. How about using this https://pkg.go.dev/github.com/stretchr/testify/require?utm_source=godoc#JSONEq instead of checking strings. It should catch more cases

pkg/monitoring/schemas.go
```diff
@@ -176,9 +201,15 @@ func init() {
 		panic(fmt.Errorf("failed to parse Avro schema for the latest transmission: %w", err))
 	}
 
+	configSetSimplifiedCodec, err = goavro.NewCodec(ConfigSetSimplifiedAvroSchema)
+	if err != nil {
+		panic(fmt.Errorf("failed to parse Avro schema for the latest configSimplified: %w", err))
```
topliceanu, line 206:
```suggestion
		panic(fmt.Errorf("failed to parse Avro schema for the latest configSetSimplified: %w", err))
```

PR #43, opened 2021-12-15
Fix validator state decoding bug (impacts transmit)

### LINKED ISSUE #22: Add deploying validator to local testing environment
Using a validator is not required for operation (as far as i know), but it would be good to include in our local e2e test

aalu1418:
closes #22 

Brings validator back into local testing env

✅ Verify that the local testing env transmits with the validator components


pkg/solana/types.go
```diff
@@ -137,6 +137,7 @@ type AccessController struct {
 
 // Validator state
 type Validator struct {
+	AccountDiscriminator     [8]byte // first 8 bytes of the SHA256 of the account’s Rust ident, https://docs.rs/anchor-lang/0.18.2/anchor_lang/attr.account.html
```
aalu1418, line 140: This was the fix!

State decoding of the validator was incorrect (missing the AccountDiscriminator) so it was using the wrong value in `transmit`.

se3000 (approved):

archseer (approved):

topliceanu (approved):

PR #45, opened 2021-12-15
23179 remove formatStaged script

tateexon:
The script this is referencing no longer exists in the project


RodrigoAD (approved):
Thanks @tateexon

PR #49, opened 2021-12-16
Gauntlet no programid required

RodrigoAD:


gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -36,7 +38,27 @@ const getContractSchema = (name: CONTRACT_LIST, version: string) => {
...
+    [CONTRACT_LIST.ACCESS_CONTROLLER]: 'ACCESS_CONTROLLER',
+    [CONTRACT_LIST.OCR_2]: 'OCR2',
+    [CONTRACT_LIST.DEVIATION_FLAGGING_VALIDATOR]: 'DEVIATION_FLAGGING_VALIDATOR',
+  }
+
+  try {
+    return new PublicKey(process.env[envNames[name]]!)
+  } catch (e) {
+    logger.warn(`No program id set for program ${name}`)
+    return
```
krebernisak, line 62: Why swallow the error here instead of letting it go to bubble up?

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -36,7 +38,27 @@ const getContractSchema = (name: CONTRACT_LIST, version: string) => {
...
 
-const getProgramId = (name: CONTRACT_LIST, version: string) => {
-  const rawPK = io.readJSON(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/programId', `${name}`))
-  return Keypair.fromSecretKey(Uint8Array.from(rawPK))
+const getProgramKeypair = (name: CONTRACT_LIST, version: string): Keypair | undefined => {
+  try {
+    const rawPK = io.readJSON(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/programId', `${name}`))
+    return Keypair.fromSecretKey(Uint8Array.from(rawPK))
+  } catch (e) {
+    logger.warn(`No program id keypair set for program ${name}`)
```
krebernisak, line 46: Why swallow the error here instead of letting it go to bubble up?

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setValidatorConfig.ts
```diff
@@ -37,7 +37,8 @@ export default class SetValidatorConfig extends SolanaCommand {
 
   execute = async () => {
     const ocr2 = getContract(CONTRACT_LIST.OCR_2, '')
-    const address = ocr2.programId.publicKey.toString()
+    this.require(!!ocr2.programId, 'OCR 2 Program ID is necessary. Set it with "OCR2" env var')
+    const address = ocr2.programId!.toString()
     const program = this.loadProgram(ocr2.idl, address)
```
krebernisak, line 42: These 3-4 lines are repeated 7-8 times in this PR. Please consider extracting it in a separate fn.

gauntlet/packages/gauntlet-solana-contracts/networks/.env.devnet
```diff
@@ -1,2 +1,5 @@
 NODE_URL=https://api.devnet.solana.com
 
+OCR2=GA6oqeZii6WLgkQbR3NKaP1gsBUPeUvMAqakHiQt8mBX
```
krebernisak, line 3: At least add a comment above these are program IDs, but I'd rather those env vars be more descriptive perhaps with the `PROGRAM_ID_` prefix.

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
RodrigoAD: You're right

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
RodrigoAD: Keypair of the program is only necessary for deployments. For interacting we just need the pub key, which we can get from `.env`

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setValidatorConfig.ts
RodrigoAD: Yep

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -1,14 +1,15 @@
...
 
 export type Contract = {
   id: CONTRACT_LIST
-  bytecode: Buffer
+  bytecode: Buffer | undefined
   version: string
   idl: any
-  programId: Keypair
+  programId: PublicKey | undefined
+  programKeypair: Keypair | undefined
```
krebernisak, line 12:
```suggestion
  programId?: PublicKey
  programKeypair?: Keypair
```

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -1,14 +1,15 @@
...
-import { Keypair } from '@solana/web3.js'
+import { io, logger } from '@chainlink/gauntlet-core/dist/utils'
+import { Keypair, PublicKey } from '@solana/web3.js'
 import { readFileSync } from 'fs'
 import { join } from 'path'
 
 export type Contract = {
   id: CONTRACT_LIST
-  bytecode: Buffer
+  bytecode: Buffer | undefined
```
krebernisak, line 8:
```suggestion
  bytecode?: Buffer
```

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
krebernisak: Yeah, but IMHO we should use it only when we need and in that case, the error should be thrown.

So if we use it (call that fn) when we don't actually need it, and then swallow the err because we didn't actually need it so not interested, that points that we should restructure the code slightly to not actually call it if we don't need it (not deploying).

Minor thing, feel free to keep it as is and move forward for now.

RodrigoAD: That makes sense. Split the function into 2. If a deployment version of the contract is needed, they need ask for it explicitly

krebernisak (approved):
LGTM!

PR #52, opened 2021-12-17
Monitoring improvements

topliceanu:
- [x] add tests for the Schema registry
- [x] refactor config
- [x] drop flags and keep only the env vars
- [x] add configurable poll interval and timeout for reading from solana and rdd
- [x] add test for encoding and empty state and transmission objects in avro + fixes
- [x] remove all keys from structures to get go's type system to help out
- [x] add base 58 encoding 'feed_state_account'
- [x] add exporter abstraction for kafka and prometheus.
- [ ] ~~remove dependency on `core/logging`~~ Not atm
- [ ] ~~add module to follow weiwatcher changes on feeds and restart~~ Doing this in a separate PR


pkg/monitoring/poller.go
```diff
@@ -51,10 +54,16 @@ func (s *solanaPollerImpl) Start(ctx context.Context) {
...
 
 	for {
-		timer := time.NewTimer(s.fetchInterval)
+		timer := time.NewTimer(s.pollInterval)
 		select {
 		case <-timer.C:
-			data, err := s.reader.Read(ctx, s.account)
+			var data interface{}
+			var err error
+			func() {
```
pedrovsg, line 62: Why this pattern?

pkg/monitoring/mapping.go
```diff
@@ -26,6 +26,20 @@ func MakeConfigSetMapping(
...
 		return nil, fmt.Errorf("failed to parse ReportingPluginConfig from OffchainConfig: %w", err)
 	}
+	var sharedSecredEncryptions map[string]interface{}
+	if offchainConfig.SharedSecretEncryptions != nil {
+		sharedSecredEncryptions = map[string]interface{}{
+			"diffie_hellman_point": offchainConfig.SharedSecretEncryptions.DiffieHellmanPoint,
+			"shared_secret_hash":   offchainConfig.SharedSecretEncryptions.SharedSecretHash,
+			"encryptions":          offchainConfig.SharedSecretEncryptions.Encryptions,
+		}
+	} else {
```
pedrovsg, line 36: can create as empty and fill if != nil

pkg/monitoring/config/feeds.go
```diff
@@ -0,0 +1,82 @@
...
+func populateFeeds(cfg *Config) error {
+	feeds := []jsonFeedConfig{}
+	if cfg.Feeds.URL != "" {
+		rddCtx, cancel := context.WithTimeout(context.Background(), cfg.Feeds.RddReadTimeout)
+		defer cancel()
+		readFeedsReq, err := http.NewRequestWithContext(rddCtx, http.MethodGet, cfg.Feeds.URL, nil)
+		if err != nil {
+			return fmt.Errorf("unable to build a request to the RDD URL '%s': %w", cfg.Feeds.URL, err)
+		}
+		httpClient := &http.Client{}
```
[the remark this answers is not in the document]
topliceanu: There is a timeout set on the request. See line [16](https://github.com/smartcontractkit/chainlink-solana/pull/52/files#diff-0e036670d4454b42804b87cb7dbccf8fabfa1eaa23c047d169fb1eeb625dd4b1R16)

pkg/monitoring/poller.go
topliceanu: So I can `defer cancel()` which ensures it will be executed even if Read panics!

pkg/monitoring/mapping.go
topliceanu: Done

topliceanu:
@pedrovsg @johnnymugs I rebased on develop and this is now ready for another review

pkg/monitoring/config/feeds.go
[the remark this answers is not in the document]
pedrovsg: I see the ctx timeout, but is a good pratice to add it to the http client.

pkg/monitoring/poller.go
pedrovsg: Defer cancel should always run since you call it before the read. Don't understand the difference of wrapping into a func

johnnymugs (approved):

PR #55, opened 2021-12-17
Update confirmation on state reading

aalu1418:
Originally state reading confirmation level was not set which defaults to `finalized`

This causes an issue where it can take 25 seconds for a tx to be finalized and the change reflected in the state. This will cause libocr to trigger sending extra transactions in an effort to transmit the report and the txs will reject with stale report errors.

This PR changes the confirmation level to `confirmed` which in local testing eliminated the stale report errors in transmitting

### DIFF
--- a/pkg/solana/contract.go
+++ b/pkg/solana/contract.go
@@ -16,6 +16,7 @@ import (
 
 var (
 	configVersion uint8 = 1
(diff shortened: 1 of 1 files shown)


krebernisak (approved):
Confirmed 🚀

PR #56, opened 2021-12-17
Improve program IDs handling

RodrigoAD:
- [x] Improve program IDs handling (use helper scripts)
- [x] Add `localnet` static keys for faster setup (change keys location & regenerate)
- [x] Add program IDs for `localnet`


contracts/programs/access-controller/src/lib.rs
```diff
@@ -5,6 +5,13 @@ use std::mem;
 
 use arrayvec::arrayvec;
 
+#[cfg(feature = "mainnet")]
+declare_id!("DzzjdPWNfwHZmzPVxnmqkkMJraYQQRCpgFZajqkqmU6G");
```
archseer, line 9: @RodrigoAD I pushed a change to allow specifying network via a cargo feature:

`anchor build -- --features mainnet` for mainnet. If no flag specified it will build for localnet instead.

archseer: Can you plug in the `devnet` ids?

aalu1418 (approved):
🎉

contracts/programs/access-controller/src/lib.rs
krebernisak: Let's not include any devnet or mainnet program IDs, for now.

PR #61, opened 2021-12-20
go mod tidy

krebernisak:


aalu1418 (commented):
These packages look like they are from the e2e QA testing...is that something we want included in the relay pkg? (it  would be nice to keep the relay module as light weight as possible)

krebernisak:
@aalu1418 I've just run `go mod tidy` here and did not add anything extra.

I don't know a solution to the problem you are suggesting, so I've opened this issue to track:

- https://github.com/smartcontractkit/chainlink-solana/issues/62

aalu1418 (approved):
🚀 approving for now, and we can revisit with the issue at a later point!

PR #74, opened 2021-12-24
Contracts: Allow storing more historical data & refactor validator+proxy into store

archseer:


pkg/monitoring/mapping.go
```diff
@@ -97,8 +97,6 @@ func MakeConfigSetMapping(
 				"billing": map[string]interface{}{
 					"observation_payment": int64(state.Config.Billing.ObservationPayment),
 				},
-				"validator":          state.Config.Validator[:],
-				"flagging_threshold": int64(state.Config.FlaggingThreshold),
```
archseer, line 101: @topliceanu The validator argument was removed, but flagging_threshold was moved over to the transmissions account. I'm also moving over decimals + description in a follow-up, is that ok or should changes be made so that monitoring also reads from the feed account?

pkg/solana/contract.go
```diff
@@ -75,6 +77,24 @@ func (c *ContractTracker) fetchState(ctx context.Context) error {
...
+		Encoding:   "base64",
+		Commitment: rpcCommitment,
+		DataSlice: &rpc.DataSlice{
+			Offset: &offset,
+			Length: &length,
+		},
+	})
+	if err != nil {
+		return err
+	}
```
archseer, line 94: @aalu1418 this is now fetched together with the state

pkg/solana/contract.go
```diff
@@ -168,13 +198,27 @@ func GetLatestTransmission(ctx context.Context, client *rpc.Client, account sola
...
+	err = binary.Read(buf, binary.LittleEndian, &timestamp)
+	if err != nil {
+		return Answer{}, 0, err
+	}
+
+	// TODO: we could use ag_binary.Int128 instead
+	err = binary.Read(buf, binary.LittleEndian, &raw)
+	if err != nil {
+		return Answer{}, 0, err
+	}
```
archseer, line 214: We should be able to just use `ag_binary` to decode these into a struct rather than manual bit fiddling.

tests/e2e/smoke/common.go
```diff
@@ -133,13 +133,13 @@ func CreateOCR2Jobs(
...
-	validator contracts.OCRv2DeviationFlaggingValidator) error {
+	store contracts.OCRv2Store) error {
 	relayConfig := map[string]string{
 		"nodeEndpointHTTP":   "http://sol:8899",
 		"nodeEndpointWS":     "ws://sol:8900",
 		"stateID":            ocr2.Address(),
-		"transmissionsID":    ocr2.TransmissionsAddr(),
-		"validatorProgramID": validator.ProgramAddress(),
+		"transmissionsID":    store.TransmissionsAddress(),
+		"validatorProgramID": store.ProgramAddress(),
```
archseer, line 142: @aalu1418 I didn't rename this spec param so there's less breakage, but maybe we should do that anyway?

pkg/solana/contract.go
```diff
@@ -121,36 +141,46 @@ func GetState(ctx context.Context, client *rpc.Client, account solana.PublicKey)
...
 	}
-	cursor := binary.LittleEndian.Uint32(c)
+	err = binary.Read(buf, binary.LittleEndian, &cursor)
+	if err != nil {
+		return Answer{}, 0, errCursorLength
+	}
+
 	if cursor == 0 { // handle array wrap
-		cursor = TransmissionsSize
+		cursor = liveLength
```
archseer, line 177: Feeds can now be allocated of any length so this is fetched from the header

pkg/monitoring/mapping.go
topliceanu: I don't know what the feed account is? Is it the same as the transmissions account? 
Yeah, it looks like all of that might need to be update after your changes are ins

aalu1418 (approved):
🚀 can be merged with CI passing (smoke test won't pass until new image is created) will be working on local testing env in separate PR

PR #75, opened 2021-12-27
simplify vec index removes

drbh:
small changes to simplify removing items from a vector at specified index


drbh:
@archseer please see small changes

archseer:
Not sure why the rust actions aren't running, but it's not a Vec but a custom arrayvec implementation. Retain isn't implemented. https://github.com/smartcontractkit/chainlink-solana/blob/develop/contracts/crates/arrayvec/src/lib.rs

drbh:
ahh sorry about that, I didnt realize it was an `arrayvec` thanks for the information!

PR #87, opened 2022-01-07
fix: store.latest() needs to account for wraparound

archseer:


contracts/programs/store/src/state.rs
```diff
@@ -117,7 +117,11 @@ impl<'a> Feed<'a> {
             return None;
         }
 
-        Some(self.live[self.header.live_cursor as usize - 1])
+        let len = self.header.live_length;
+        // Handle wraparound
+        let i = ((self.header.live_cursor + len.saturating_sub(1)) % len) % len;
```
aalu1418, line 122: Curious: why take the modulus twice? (doesn't that just generate the same number?)

aalu1418 (approved):
🚀 Approving to unblock and get the changes in

PR #88, opened 2022-01-07
Move decimals/description from the aggregator to feed

archseer:


aalu1418 (approved):
🚀 lgtm, will continue to validate changes with local testing in a separate PR

PR #89, opened 2022-01-07
fix: get local testing working, fix memory leak

aalu1418:
Main: fix memory leak by removing WS connection
* temporary solution: don't confirm transaction, use simulate to catch any errors that might occur before transmission
* long term solution (todo): use polling to confirm tx #90 

Secondary: small doc updates + remove extra modulus in contract

Tested locally for 2+ hours: ✅ contract fix, ✅ removed WS, ✅ memory leak

corresponding core implementation: https://github.com/smartcontractkit/chainlink-solana/pull/89


pkg/solana/transmitter.go
```diff
@@ -81,18 +80,18 @@ func (c *ContractTracker) Transmit(
...
-			true, // skip preflight
+			false, // use preflight as first check
 			rpc.CommitmentConfirmed,
 		)
 
 		if err != nil {
 			c.lggr.Errorf("error on Transmit.SendAndConfirmTransaction: %s", err.Error())
 		}
+		// TODO: poll rpc for tx confirmation (WS connection unreliable)
+		// TODO: see if SendTransactionWithOpts returns quickly enough to be used outside of goroutine
```
krebernisak, line 94: We should not count on it being quick enough, as this is a side-effect that depends on other systems and infrastructure (connection).

pkg/solana/transmitter.go
```diff
@@ -81,18 +80,18 @@ func (c *ContractTracker) Transmit(
...
 			tx,
-			true, // skip preflight
+			false, // use preflight as first check
 			rpc.CommitmentConfirmed,
 		)
 
 		if err != nil {
 			c.lggr.Errorf("error on Transmit.SendAndConfirmTransaction: %s", err.Error())
 		}
+		// TODO: poll rpc for tx confirmation (WS connection unreliable)
```
krebernisak, line 93: Was there an agreement we are dropping WS support for the long term?

I'd prefer we actually have two implementations, both simpler but more resource-intensive RPC polling, and WS subscription so we can make incremental improvements to it and potentially switch again once stable enough.

aalu1418: Sort of, there was consensus for dropping it for now. 

I agree that the RPC polling is more resource intensive, but the WS implementation in the solana-go module was the cause of a memory leak which would take down the core node. Building a stable WS implementation would take a lot of time and effort, so it was shelved for now.

krebernisak: I'd just suggest we declare a TxConfirmer as an interface and then we implement the RPC TxConfirmer + some docs about what were the latest encountered issues using the WS TxConfirmer (no need to do any work on it now). That way we have a defined path of switching implementations if/when WS implementation becomes stable.

aalu1418: that sounds like a good idea, but probably better addressed in a separate PR?
it will also likely tie into some tx management work for terra as well

krebernisak: Yup, no need to slow down this PR whose main purpose is to quickly fix the memory leak and get the network stable.

I'll open a GH Issue to track this suggestion for the future. And yeah, we should also define types like `TxCofirmer` on the Relay SDK level and use (implement) them for specific relays.

aalu1418: might be easier to add comments to this open issue for adding a tx confirmer: https://github.com/smartcontractkit/chainlink-solana/issues/90

krebernisak (approved):
LGTM!

PR #92, opened 2022-01-10
Add CI pipeline step to run the race detector against the tests

topliceanu:


krebernisak (commented):
Nice catch! Could you please add the same step to the Terra relay workflow, so we also run the race detector there?

krebernisak:
@SergeyDNazarov is this a flaky test (e2e/smoke)?

https://github.com/smartcontractkit/chainlink-solana/runs/4761612631?check_suite_focus=true

This PR (monitoring service fix) should not influence the ocr2 smoke test, right?

aalu1418:
> @skudasov is this a flaky test (e2e/smoke)?
> 
> https://github.com/smartcontractkit/chainlink-solana/runs/4761612631?check_suite_focus=true
> 
> This PR (monitoring service fix) should not influence the ocr2 smoke test, right?

Not flaky tests, it's because the contracts changed, and I haven't been able to get an updated core image. So the test will continue to fail until the core version in the e2e tests are updated.

krebernisak (approved):

PR #96, opened 2022-01-11
Close account instructions

archseer:
Needs to be backported as 0.1.3 so we can close down and reclaim existing accounts


krebernisak (approved):
Feels it needs a rebase, but other than that looks good!

PR #97, opened 2022-01-11
miscellaneous gauntlet deployment fixes

aalu1418:


gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/initialize.ts
```diff
@@ -111,6 +111,7 @@ export default class Initialize extends SolanaCommand {
       STATE ACCOUNTS:
         - State: ${state.publicKey}
         - Transmissions: ${transmissions}
+        - StoreAuthority: ${storeAuthority.toString()}
```
aalu1418, line 114: Used in `store:set_writer`, printing for ease of use

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setPayees.ts
```diff
@@ -70,7 +70,13 @@ export default class SetPayees extends SolanaCommand {
...
           input.operators.map(async ({ payee }) => {
             try {
-              const info = await token.getAccountInfo(new PublicKey(payee))
+              const to = await Token.getAssociatedTokenAddress(
+                ASSOCIATED_TOKEN_PROGRAM_ID,
+                TOKEN_PROGRAM_ID,
+                token.publicKey,
+                new PublicKey(payee),
+              )
+              const info = await token.getAccountInfo(to)
```
aalu1418, line 79: This could be a misinterpretation on my end, but it seems like the `payee` parameter should be a normal solana address, and then the associated token address should be calculated before checking if exists

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/createFeed.ts
```diff
@@ -54,7 +54,7 @@ export default class CreateFeed extends SolanaCommand {
 
     const granularity = new BN(input.granularity)
     const liveLength = new BN(input.liveLength)
-    const length = new BN(8096)
+    const length = new BN(this.flags.length || 8096)
```
aalu1418, line 57: Ability to configure length as requested by @archseer

archseer: We should also validate that liveLength <= length

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/createFeed.ts
```diff
@@ -69,8 +69,13 @@ export default class CreateFeed extends SolanaCommand {
     console.log(`
       - Decimals: ${decimals}
       - Description: ${description}
+      - Live Length: ${liveLength.toNumber()}
+      - Granularity (historical): ${granularity.toNumber()}
+      - Total Length: ${length.toNumber()}
```
archseer, line 74: You could also compute the historical length (length - liveLength)

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setPayees.ts
krebernisak: If `to` is the associated token address, what is the `info.address` then?

Ok, they are the same by looking at [this client code](https://github.com/solana-labs/solana-program-library/blob/583afbd35f8ad16bf844386183c2cfd5cbd6fac3/token/js/client/token.js#L779).

Conclusion, the `const info = await token.getAccountInfo(to)` LOC is unnecessary.

krebernisak: Additionally, what if node op did not prepare his associated token address? 

The deployment will fail and be blocked, right?. So what should we do in that case?
Should we create the associated accounts (fund) on their behalf like described in this [Example: Transferring tokens to another user, with sender-funding](https://spl.solana.com/token#example-transferring-tokens-to-another-user-with-sender-funding)?

Why did this work before this change where we didn't check for associated acc? 

1. did we/they pre-create all associated token accounts?
2. or does the contract creates/funds them on the first transfer?
3. or we didn't even test withdrawals?

cc @archseer @RodrigoAD

RodrigoAD: Payee address needs to be a token associated address. NodeOps created it in advance on past deployments.
You can allow sender-funding with [this option](https://github.com/smartcontractkit/chainlink-solana/pull/97/files#diff-03d47de09b3b7833573e594fa7b2b9045fdc42a5fe36a86e03cc85966dd689b4R15)
If the NodeOp didn't prepare the payee address, we prepare it for him, and run the command again.

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/createFeed.ts
aalu1418: yup! this is validated: https://github.com/smartcontractkit/chainlink-solana/pull/97/files#diff-ecea88c93003d6366639eac590fafe34bb4ee31afc3bc09f342ffeb528f79c67R64

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setPayees.ts
aalu1418: ok, seems like I was interpreting things incorrectly. the `adminAddress` from RDD should be the associatedTokenAccount. which makes these changes incorrect, i will remove them 😄

RodrigoAD (approved):
LGTM

PR #105, opened 2022-01-13
Gauntlet transfer ownership

RodrigoAD:


krebernisak (approved):

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/acceptOwnership.ts
```diff
@@ -0,0 +1,49 @@
...
+  execute = async () => {
+    const store = getContract(CONTRACT_LIST.STORE, '')
+    const storeAddress = store.programId.toString()
+    const storeProgram = this.loadProgram(store.idl, storeAddress)
+
+    const owner = this.wallet.payer
+
+    const storeState = new PublicKey(this.flags.state)
+
+    await prompt(`Accepting ownership of store state (${storeState.toString()}). Continue?`)
```
aalu1418, line 28: small nit pick, but would be nice to validate that the proposed owner onchain is the same as the proposed owner. but that can be tackled at a later point too 😄

RodrigoAD: The tx will revert if it's not like that, but is a nice to have. Leaving it for next iteration

PR #106, opened 2022-01-14
Ownership commands for every program

RodrigoAD:


gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ownership/acceptOwnership.ts
```diff
@@ -0,0 +1,54 @@
...
+      const owner = this.wallet.payer
+
+      const state = new PublicKey(this.flags.state)
+
+      await prompt(`Accepting ownership of ${contractId} state (${state.toString()}). Continue?`)
+
+      const tx = await program.rpc.acceptOwnership({
+        accounts: {
+          // Store contract expects an store account instead of a state acc
+          ...(contractId === CONTRACT_LIST.STORE && { store: state }),
```
RodrigoAD, line 35: @archseer Seems like the store contract expects a different account name than the rest. Should we have same interface on every program?
(https://github.com/smartcontractkit/chainlink-solana/blob/develop/contracts/programs/store/src/lib.rs#L133)

archseer: Yeah we could do that, I kept the name different initially since there were some calls where both `store` and ocr2 `state` had to be passed but in that case it's probably better to use `store_state`/`ocr2_state`.

archseer: https://github.com/smartcontractkit/chainlink-solana/issues/113

archseer (approved):

PR #110, opened 2022-01-14
fix raising flag

skudasov:


contracts/programs/store/src/lib.rs
```diff
@@ -169,7 +169,7 @@ pub mod store {
         };
         let feed_address = ctx.accounts.feed.key();
 
-        if is_valid {
+        if !is_valid {
```
krebernisak, line 172: This seems right, but we'd need a test for this.

krebernisak: GH Issue:

- https://github.com/smartcontractkit/chainlink-solana/issues/112

krebernisak (approved):

PR #114, opened 2022-01-15
Improve store_program validation

archseer:
The Program<T> account type provided by Anchor will check that the provided address is executable (and that the id matches the expected program_id). This can help catch issues where a regular account is provided accidentally.

### DIFF
--- a/contracts/programs/ocr2/src/context.rs
+++ b/contracts/programs/ocr2/src/context.rs
@@ -6,7 +6,7 @@ use anchor_spl::token::{Mint, Token, TokenAccount, Transfer};
 use crate::state::State;
 
 use access_controller::AccessController;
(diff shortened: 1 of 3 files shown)


contracts/programs/ocr2/src/context.rs
```diff
@@ -79,8 +79,7 @@ pub struct Transmit<'info> {
     #[account(mut, address = state.load()?.transmissions)]
     pub transmissions: Account<'info, Transmissions>,
 
-    #[account(address = store::ID)]
-    pub store_program: AccountInfo<'info>,
```
archseer, line 83: This is the bulk of the change. The rest just adds the `Store` type and renames the old `Store` type to `State`.

PR #125, opened 2022-01-20
Example v2

archseer:
With return data now enabled on devnet I've brought back the code but disabled the code for mainnet builds


aalu1418 (approved):
lgtm!

PR #140, opened 2022-01-25
Update job specs examples

krebernisak:


examples/spec/ocr2-oracle.spec.toml
```diff
@@ -78,3 +73,8 @@ nodeEndpointHTTP   = "http:..."
 ocr2ProgramID      = "<insert solana ocr2 program ID>"
 transmissionsID    = "<insert solana ocr2 transmissions account>"
 storeProgramID     = "<insert solana ocr2 store account>"
+usePreflight       = false
+commitment         = "confirmed"
+pollingInterval    = "1s"
+pollingCtxTimeout  = "2s"
+staleTimeout       = "1m"
```
aalu1418, line 80: these are optional parameters, there are defaults included in the relay...not sure if we should have them here or not

aalu1418: if you wanted to include these, then should they be included in the bootstrap spec too?

aalu1418 (commented):
Overall looks good to me!

i've never tested using env var parameters since it usually requires rebooting a node to test that it is set. But i believe it does work like that

examples/spec/ocr2-oracle.spec.toml
krebernisak: Yup! 

1. Are all relevant for the bootstrap node?
2. Are default values listed or should I fish for default values to document?

aalu1418: 1. the `preflight` is the only one that is not since it's for transmissions (although it will just default to false anyway)
2. those are the default values! (the default `pollingCtxTimeout` is calculated via double the `pollingInterval`)

PR #141, opened 2022-01-25
Feature: ability to locally test swapping out node versions for select CL nodes

aalu1418:


krebernisak (approved):
LGTM!

We'll need to find a way to define the CL image version on the repo level, not file (3) level.

PR #146, opened 2022-01-26
SOM: use promauto

pedrovsg:


krebernisak:
@pedrovsg what is the status of this PR?

PR #149, opened 2022-01-27
Feed layout v2

archseer:
[Do not merge, do not squash]

Adds the current slot to rounds and some padding for future data.


pkg/solana/contract.go
```diff
@@ -309,16 +309,27 @@ func GetLatestTransmission(ctx context.Context, client *rpc.Client, account sola
 	if len(t) != int(transmissionLen) { // validate length
 		return Answer{}, 0, errTransmissionLength
 	}
+	buf = bytes.NewReader(t)
```
archseer, line 312: @aalu1418 still need to adjust this so that we read the version from the header, then if v1, read using the old offsets (24 byte transmission lengths etc), if v2 then use the following.

Maybe we can finally use https://github.com/gagliardetto/binary with a Decoder?

pkg/solana/contract.go
```diff
@@ -305,32 +294,27 @@ func GetLatestTransmission(ctx context.Context, client *rpc.Client, account sola
...
+			return Answer{}, 0, errors.Wrap(err, "failed to decode v1 transmission")
+		}
 
-	buf = bytes.NewReader(t)
-	err = binary.Read(buf, binary.LittleEndian, &timestamp)
-	if err != nil {
-		return Answer{}, 0, err
+		return Answer{
+			Data:      t.Answer.BigInt(),
+			Timestamp: uint32(t.Timestamp), // TODO: not good typing conversion
```
aalu1418, line 306: Just to flag, that there should be a better way to handle this....

contracts/tests/ocr2.spec.ts
```diff
@@ -320,6 +320,48 @@ describe('ocr2', async () => {
...
+    } catch (err) {
+      // Translate IDL error
+      const idlErrors = anchor.parseIdlErrors(program.idl);
+      let translatedErr = ProgramError.parse(err, idlErrors);
+      if (translatedErr === null) {
+        throw err;
+      }
+      throw translatedErr;
+    }
+  });
```
se3000, line 363: Should we check that the latest previous answer is still readable?

se3000 (approved):

PR #160, opened 2022-01-31
Mix in the state account address into the config digest, additional tweaks

archseer:
e2e tests will fail because a new core image needs to be built


kaleofduty:
Do we need to include the transmissions account as well?

What other accounts are there in the system which we are not including?

archseer:
Going to include this in #194

Issue #170, opened 2022-02-01
Small config digester improvements


aalu1418:
This can be replaced with a constant from libocr
https://github.com/smartcontractkit/chainlink-solana/blob/41bc04a342d73aeb30454278ba5f8c36a314f443/pkg/solana/config_digester.go#L15

This is a little brittle:
https://github.com/smartcontractkit/chainlink-solana/blob/41bc04a342d73aeb30454278ba5f8c36a314f443/pkg/solana/config_digester.go#L87
Suggestion: Use `binary.BigEndian.PutUint16`

Issue #169, opened 2022-02-01
Potential to leak goroutines in the relay transmitter


aalu1418:
https://github.com/smartcontractkit/chainlink-solana/blob/41bc04a342d73aeb30454278ba5f8c36a314f443/pkg/solana/transmitter.go#L87

There should be a timeout associated with the context.

PR #171, opened 2022-02-01
Small relay fixes + improvements

aalu1418:
- [x] use libocr for digest prefix #170 
- [x] use PutUint16 for digest prefix #170 
- [x] implement timeout for transmit context #169 
- [x] document relayConfig variables in a md file (see comment below)

To be addressed in separate PR:
- add golangci-lint parameters from core #173 
- fix potential nil pointer risk #174

### DIFF
--- a/examples/spec/ocr2-bootstrap.spec.toml
+++ b/examples/spec/ocr2-bootstrap.spec.toml
@@ -14,6 +14,7 @@ ocr2ProgramID      = "<insert solana ocr2 program ID>"
 transmissionsID    = "<insert solana ocr2 transmissions account>"
 storeProgramID     = "<insert solana ocr2 store account>"
 commitment         = "confirmed" # optional, defaults to "confirmed"
+txTimeout          = "1m"        # optional, defaults to "1m"
 pollingInterval    = "1s"        # optional, defaults to "1s"
 pollingCtxTimeout  = "2s"        # optional, defaults to `2x ${pollingInterval}`
 staleTimeout       = "1m"        # optional, defaults to "1m"
--- a/pkg/solana/client.go
+++ b/pkg/solana/client.go
@@ -49,7 +49,7 @@ func NewClient(spec OCR2Spec, logger Logger) *Client {
 	// parse context lenght, if errors, use 2x poll interval
 	ctxInterval, err := time.ParseDuration(spec.PollingCtxTimeout)
 	if err != nil {
-		logger.Warnf("could not parse polling context duration ('%s') using default 2x polling interval (%s)", spec.PollingCtxTimeout, 2 * pollInterval)
+		logger.Warnf("could not parse polling context duration ('%s') using default 2x polling interval (%s)", spec.PollingCtxTimeout, 2*pollInterval)
 		ctxInterval = 2 * pollInterval
 	}
 
--- a/pkg/solana/client.go
+++ b/pkg/solana/client.go
@@ -13,7 +13,7 @@ type Client struct {
 	rpc             *rpc.Client
 	skipPreflight   bool // to enable or disable preflight checks
 	commitment      rpc.CommitmentType
-	txDuration       time.Duration
+	txTimeout       time.Duration
 	pollingInterval time.Duration
 	contextDuration time.Duration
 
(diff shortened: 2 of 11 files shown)


pkg/solana/transmitter.go
```diff
@@ -83,8 +83,10 @@ func (c *ContractTracker) Transmit(
 
 	// Send transaction, and wait for confirmation:
 	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), c.client.txTimeout)
+		defer cancel()
 		txSig, err := c.client.rpc.SendTransactionWithOpts(
-			context.Background(), // does not use libocr transmit context
+			ctx, // does not use libocr transmit context
```
aalu1418, line 89: Add a 1m default context cancel to prevent leaky goroutines

pkg/solana/client.go
```diff
@@ -52,8 +53,16 @@ func NewClient(spec OCR2Spec, logger Logger) *Client {
 		ctxInterval = 2 * pollInterval
 	}
 
+	// parse tx context, if errors use defaultStaleTimeout
```
connorwstein, line 56: a broader nit: right now the validation is split between core and chainlink-solana, I think we should just move it all into core since the earlier we get the raw strings into types the better. OCR2Spec can just be assumed to be a validated spec on the relay side

pkg/solana/transmitter.go
connorwstein: oh one potential issue I just noticed - is it possible SendTransactionWithOpts when it errors will return a nil txSig? If so that Debugf log will panic on txSig.String(). Probably want to early return on the error case

aalu1418: hmmm it returns this: https://github.com/gagliardetto/solana-go/blob/d741961be7bd2b53d2a46e93518a6d573bb61f16/nativetypes.go#L82

but i'm not sure if it will be nil...i'll have to do some more digging

connorwstein: nah you're safe then, arrays can't be nil. If it was a slice then it could be an issue

pkg/solana/client.go
krebernisak: I'm also not fond of the split but would like to have most network-specific code here instead of in Core. That doesn't mean that we can not execute earlier in Core, but execute code that is defined here.

On the other hand, for the short term, all/most of these args should probably be moved as Solana chain configuration in core to follow Ethereum & Terra.

examples/spec/ocr2-bootstrap.spec.toml
```diff
@@ -14,6 +14,7 @@ ocr2ProgramID      = "<insert solana ocr2 program ID>"
 transmissionsID    = "<insert solana ocr2 transmissions account>"
 storeProgramID     = "<insert solana ocr2 store account>"
 commitment         = "confirmed" # optional, defaults to "confirmed"
+txTimeout          = "1m"        # optional, defaults to "1m"
```
krebernisak, line 17: Do you mind using this PR to just add human-readable descriptions to a README file, or some other `.md` file under `./docs` about the specific purpose of these knobs? 

Something like Connor did for Terra in a Notion doc describing NodeOps setup, but we'll also move those docs to the Terra repo as a single source of truth.

aalu1418: can do! i'll add a few more items to my todo list at the top

pkg/solana/client.go
aalu1418: i don't mind either way (handling the validation in core or external to it)

And i agree that this needs to be refactored into a chain config, but that's a bit out of scope for this PR. I book marked it here to revisit later on: #176

krebernisak: Of course, I didn't mean this PR but as a next first step to bring it up to date with Terra :)

pkg/solana/client.go
```diff
@@ -52,8 +53,16 @@ func NewClient(spec OCR2Spec, logger Logger) *Client {
...
+	txTimeout, err := time.ParseDuration(spec.TxTimeout)
+	if err != nil {
+		logger.Warnf("could not parse tx context duration using default 1m")
```
jmank88, line 59: Maybe include the err or string that failed to parse?

krebernisak (approved):
Awesome docs 🙌

pkg/solana/client.go
aalu1418: good idea, i also implemented it for parsing the other 2 timeouts as well

PR #172, opened 2022-02-01
fetch contract bytecode from github release

akhilchainani:


gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -39,18 +39,25 @@ export const getContract = (name: CONTRACT_LIST, version: string): Contract => (
...
 })
 
-// TODO: Get it from GH Releases
-const getContractCode = (name: CONTRACT_LIST, version: string) => {
+const getContractCode = async (name: CONTRACT_LIST, version: string): Promise<Buffer> => {
   try {
-    return readFileSync(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/bin', `${name}.so`))
+    if (version === 'local') {
+      return readFileSync(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/bin', `${name}.so`))
+    } else {
```
krebernisak, line 52: nit: explicit `else` clause not required because first `if` returns

gauntlet/packages/gauntlet-solana-contracts/src/lib/contracts.ts
```diff
@@ -39,18 +39,25 @@ export const getContract = (name: CONTRACT_LIST, version: string): Contract => (
...
+  bytecode: await getContractCode(name, version),
 })
 
-// TODO: Get it from GH Releases
-const getContractCode = (name: CONTRACT_LIST, version: string) => {
+const getContractCode = async (name: CONTRACT_LIST, version: string): Promise<Buffer> => {
   try {
-    return readFileSync(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/bin', `${name}.so`))
+    if (version === 'local') {
+      return readFileSync(join(process.cwd(), 'packages/gauntlet-solana-contracts/artifacts/bin', `${name}.so`))
```
krebernisak, line 51: @RodrigoAD should we rather point to the artifact source from here which gets built under `./contracts/artifacts/...`?

RodrigoAD: We don't track the schemas artifacts in `./contracts`

krebernisak: We'll need a solution to handle binary & schema loading both for our local code and vendor (multisig).

archseer: If the IDL is stored on chain then we don't need to have a local copy, `anchor.Program.at(program_id)` is enough to construct an instance. But then we must ensure the IDL is always up to date.

PR #175, opened 2022-02-02
Unify program_id, build with link time optimizations

archseer:
[don't squash]

A bunch of small changes that don't have any impact on the relay or memory layout.

- `localnet`, `devnet` and `mainnet` use the same program_id so we only need to build the contracts once.
- Enable LTO in release mode (when running anchor build). Saves a tiny bit of space.
- `state.config.(pending_)offchain_config` was moved up a level to `state.(pending_)offchain_config`. The resulting memory layout is the same since the offchain config is now positioned right after the config, but this avoids
  ```rust
  let state = &mut *ctx.accounts.state.load_mut()?;
  let config = &mut state.config;
  ```
  which can be accidentally optimized by the compiler into a copy, blowing up the size of the stack frames. Without this fix LTO emits several warnings.


contracts/Anchor.toml
```diff
@@ -21,6 +21,6 @@ test = "yarn run test"
 # TODO: add pubkeys
 
 [programs.localnet]
-ocr2 = "CF13pnKGJ1WJZeEgVAtFdUi4MMndXm9hneiHs8azUaZt"
-access_controller = "2F5NEkMnCRkmahEAcQfTQcZv1xtGgrWFfjENtTwHLuKg"
-store = "A7Jh2nb1hZHwqEofm4N8SXbKTj82rx7KUfjParQXUyMQ"
+ocr2 = "HW3ipKzeeduJq6f1NqRCw4doknMeWkfrM4WxobtG3o5v"
```
krebernisak, line 24: Do we want to do this all the way to localnet? How are we going to share keys for CI localnet deployments which are currently here: https://github.com/smartcontractkit/chainlink-solana/tree/restructure-config/contracts/artifacts/localnet

Is it possible to do the initial deployment just by pointing to the pub key?

contracts/examples/hello-world/tests/hello-world.ts
```diff
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import { Program, BN } from '@project-serum/anchor';
 import { HelloWorld } from '../target/types/hello_world';
 
-const CHAINLINK_PROGRAM_ID = "A7Jh2nb1hZHwqEofm4N8SXbKTj82rx7KUfjParQXUyMQ";
+const CHAINLINK_PROGRAM_ID = "CaH12fwNTKJAG8PxEvo9R96Zc2j8qNHZaFj8ZW49yZNT";
```
krebernisak, line 6: Already defined in `contracts/examples/hello-world/client.js`. Let's use it from there.

contracts/examples/hello-world/tests/hello-world.ts
```diff
@@ -23,7 +23,7 @@ describe('hello-world', () => {
     const storeProgram = new Program(storeIdl, CHAINLINK_PROGRAM_ID, provider);
 
     let acIdl = JSON.parse(fs.readFileSync('../../target/idl/access_controller.json'));    
-    const accessControllerProgram = new Program(acIdl, "2F5NEkMnCRkmahEAcQfTQcZv1xtGgrWFfjENtTwHLuKg", provider);
+    const accessControllerProgram = new Program(acIdl, "DzzjdPWNfwHZmzPVxnmqkkMJraYQQRCpgFZajqkqmU6G", provider);
```
krebernisak, line 26: We could also extract this as a declared constant.

contracts/Anchor.toml
archseer: localnet works specially in that Anchor inserts the programs into the genesis block at certain addresses (without needing the key). It's convenient because you can keep using the same addresses in tests. As you can see it's passing on the CI without the private keypairs

contracts/examples/hello-world/tests/hello-world.ts
archseer: The test file doesn't use the client.js yet. I'm going to add a proper ta client in a follow-up that will clean up both files

contracts/Anchor.toml
krebernisak: That's nice, thanks for the explanation. Then additionally just delete those published localnet keys in this PR.

contracts/examples/hello-world/tests/hello-world.ts
se3000: +1

se3000 (approved):

PR #177, opened 2022-02-02
fix: startup panic if solana endpoint is not valid

aalu1418:
Problem: node on startup will panic if endpoint is not valid and there are existing jobs

Cause:
* times are not initialized to zero: https://github.com/smartcontractkit/chainlink-solana/blob/5b4d7cdb0ba245654894036369a35898f8284751/pkg/solana/contract.go#L88
* which allows the state + answer to be read because the time comparison will pass (even though the state has not been updated) https://github.com/smartcontractkit/chainlink-solana/blob/5b4d7cdb0ba245654894036369a35898f8284751/pkg/solana/contract.go#L172
* then causes a panic when libocr tries using the data returned from LatestTransmission

Fix:
* initialize time to 0 which will throw an error for an uninitialized state (and will keep throwing an error to indicate an issue with polling and the endpoint)

Testing: ✅ 
* tested locally: start up network + nodes + feed => shut down a node => stop local network => start up the stopped node
  * previously would panic at this step
  * now shows error logs related to polling and failed to fetch state/data
  * => start up local network again => all nodes join again ✅

### DIFF
--- a/pkg/solana/contract.go
+++ b/pkg/solana/contract.go
@@ -80,8 +80,6 @@ func NewTracker(spec OCR2Spec, client *Client, transmitter TransmissionSigner, l
 		lggr:            lggr,
 		stateLock:       &sync.RWMutex{},
 		ansLock:         &sync.RWMutex{},
-		stateTime:       time.Time{},
-		ansTime:         time.Time{},
 		staleTimeout:    staleTimeout,
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/solana/contract.go
```diff
@@ -80,8 +80,8 @@ func NewTracker(spec OCR2Spec, client *Client, transmitter TransmissionSigner, l
 		lggr:            lggr,
 		stateLock:       &sync.RWMutex{},
 		ansLock:         &sync.RWMutex{},
-		stateTime:       time.Now(),
-		ansTime:         time.Now(),
+		stateTime:       time.Time{},
+		ansTime:         time.Time{},
```
jmank88, line 84: Alternatively, could omit to get zero implicitly.
```suggestion
```

aalu1418: ah yeah that's true. good point

jmank88 (approved):

PR #192, opened 2022-02-08
Pin geth

connorwstein:


go.sum
```diff
@@ -1,4 +1,5 @@
 bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
+bazil.org/fuse v0.0.0-20200407214033-5883e5a4b512/go.mod h1:FbcW6z/2VytnFDhZfumh8Ss8zxHE6qpMP5sHTRe0EaM=
```
connorwstein: seems like we havent run a go mod tidy in a while here

jmank88 (approved):

go.sum
connorwstein: ah yeah _I_ was running w/o that

jmank88 (approved):

PR #204, opened 2022-02-11
Signal on contract config

connorwstein:
- Signal on contractReady to start the libocr job once we have config. Libocr assumes that the contract is already configured and config is available through LatestConfigDetails when its started
- Added separate objects for readability and isolation (in particular reducing scope of TransmissionSigner)
- See corresponding change https://github.com/smartcontractkit/chainlink/pull/5892


pkg/solana/contract_cache.go
```diff
@@ -111,6 +115,14 @@ func (c *ContractTracker) PollState() {
 				err := c.fetchState(ctx)
 				if err != nil {
 					c.lggr.Errorf("error in PollState.fetchState %s", err)
+				} else {
+					// We have successfully read state from the contract.
+					// Signal that we are ready to start libocr.
```
connorwstein, line 120: One thing to note is that, the transmission fetch needs to be able to handle empty transmissions without erroring regardless, as we can't produce transmissions until libocr is started so there will inevitably be a period of time where that is empty on first contract deploy

connorwstein:
closing in favor of simpler https://github.com/smartcontractkit/chainlink-solana/pull/303

PR #207, opened 2022-02-14
Update e2e to work with latest develop

archseer:


archseer:
@skudasov Can we make the new `ocr2` contract interface for Solana and Terra to be simply `SetConfig`/`Configure`, which will internally build a proposal through various methods and call accept on it? Exposing `SetOracles`, `SetPayees` `SetOffChainConfig` etc seems like a leaky abstraction because this depends on chain specific things.

If it's a single "configure this ocr2 aggregator" interface then the exact implementation can vary per chain, and it doesn't require changes in `integrations-framework` every time we change the contract.

skudasov:
> @skudasov Can we make the new `ocr2` contract interface for Solana and Terra to be simply `SetConfig`/`Configure`, which will internally build a proposal through various methods and call accept on it? Exposing `SetOracles`, `SetPayees` `SetOffChainConfig` etc seems like a leaky abstraction because this depends on chain specific things.
> 
> If it's a single "configure this ocr2 aggregator" interface then the exact implementation can vary per chain, and it doesn't require changes in `integrations-framework` every time we change the contract.

Sure, in fact, we don't have any interface limitations from `integrations-framework` atm.

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/setValidatorConfig.ts
```diff
@@ -81,16 +69,16 @@ export default class SetValidatorConfig extends SolanaCommand {
...
     const txhash = await this.sendTx(tx, [this.wallet.payer], contract.idl)
     logger.success(`Validator config on tx ${txhash}`)
 
     return {
       responses: [
         {
-          tx: this.wrapResponse(txhash, this.flags.state),
-          contract: this.flags.state,
+          tx: this.wrapResponse(txhash, this.flags.feed),
+          contract: this.flags.feed,
```
archseer, line 81: Not sure about this

PR #209, opened 2022-02-15
Run in anchor shell

gheorghestrimtu:


archseer:
This seems to be what `./scripts/anchor-shell.sh` does?

gheorghestrimtu:
I would use this like so
`	err := utils.RunInAnchorShell([]string{"./scripts/programs-keys-gen.sh", "localnet"}, "")
	Expect(err).ShouldNot(HaveOccurred())
	err = utils.RunInAnchorShell([]string{"anchor", "build"}, "contracts")
	Expect(err).ShouldNot(HaveOccurred())
`
Before all tests are run

skudasov (approved):

PR #208, opened 2022-02-15
Proposal and config commands

RodrigoAD:

PR #211, opened 2022-02-15
refactor command args and flags

momentmaker:


archseer:
Should probably base this on #208

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/offchainConfig/begin.ts
```diff
@@ -0,0 +1,71 @@
+import { Result } from '@chainlink/gauntlet-core'
```
archseer, line 1: These files add back commands that were removed in #208, should remove them and update the new commands instead.

archseer:
I'm integrating your changes into #208

PR #214, opened 2022-02-16
fix: Use correct observer count in the report

archseer:
len(observers) was always 32, but we want to know the actual number
of observers, not the array size.


aalu1418 (approved):
🎉

PR #221, opened 2022-02-18
handling multiple build environments

aalu1418:
Keep using a single set of program IDs in the program files
* Use scripts to modify and build for local testing and staging environment


scripts/lib.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
archseer, line 1: Use `set -euxo pipefail` on all bash scripts: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

scripts/setup-staging.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# note call from repo root
```
archseer, line 3: You can make it work relative to the script by using `${BASH_SOURCE%/*}` http://mywiki.wooledge.org/BashFAQ/028

aalu1418:
i'm pretty sure the `e2e_tests_custom_cl_image / Build custom CL image / Build Chainlink Image (push)` test in CI will fail because there were some breaking changes introduced in core related to creating bootstrap jobs. (long story short, bootstrap jobs are created with the `bootstrap` job type now rather than a flag in a spec)

archseer:
You need a rebase, @skudasov fixed this in https://github.com/smartcontractkit/chainlink-solana/pull/226

archseer (approved):

PR #223, opened 2022-02-21
Upgrade to anchor 0.22

archseer:
The main change is the error types, the name will now be output in the transaction log instead of just the code.


aalu1418 (approved):
🚀

PR #237, opened 2022-02-25
scale juels to gjuels for node payment

aalu1418:
Issue: unit mismatch between juels + gjuels

Solution: scale to juels value to gjuels to match rest of program

Tested locally ✅ 
Tested on staging env - tbd

### DIFF
--- a/contracts/programs/ocr2/src/lib.rs
+++ b/contracts/programs/ocr2/src/lib.rs
@@ -710,8 +710,8 @@ fn transmit_impl<'info>(ctx: Context<Transmit<'info>>, data: &[u8]) -> Result<()
     let reimbursement_gjuels =
         calculate_reimbursement_gjuels(report.juels_per_lamport, signature_count)?; // gjuels
     let amount_gjuels =
-        reimbursement_gjuels + u64::from(state.config.billing.transmission_payment_gjuels);
-    state.oracles[oracle_idx].payment_gjuels += amount_gjuels;
+        reimbursement_gjuels.saturating_add(u64::from(state.config.billing.transmission_payment_gjuels));
+    state.oracles[oracle_idx].payment_gjuels = state.oracles[oracle_idx].payment_gjuels.saturating_add(amount_gjuels);
 
     emit!(event::NewTransmission {
         round_id: state.config.latest_aggregator_round_id,
@@ -790,13 +790,22 @@ impl Report {
 }
 
 fn calculate_reimbursement_gjuels(juels_per_lamport: u64, _signature_count: usize) -> Result<u64> {
-    const SIGNERS: u64 = 1;
+    const SIGNERS: u128 = 1;
+    const GIGA: u128 = 10u128.pow(9);
     let fees = Fees::get()?;
-    let lamports_per_signature = fees.fee_calculator.lamports_per_signature;
+    let lamports_per_signature: u128 = fees.fee_calculator.lamports_per_signature.into();
     let lamports = lamports_per_signature * SIGNERS;
-    let juels = lamports * juels_per_lamport;
-    let gjuels = juels / 1000000000; // return value as gjuels
-    Ok(gjuels)
+    let juels = lamports * u128::from(juels_per_lamport);
+    let gjuels = juels / GIGA; // return value as gjuels
+
+    // convert from u128 to u64 with staturating logic to max u64
+    let convert = gjuels.try_into();
+    let gjuels_u64: u64 = match convert {
+        Ok(convert) => convert,
+        Err(_error) => u64::MAX,
+    };
+
+    Ok(gjuels_u64)
 }
 
 fn calculate_owed_payment_gjuels(
--- a/contracts/programs/ocr2/src/lib.rs
+++ b/contracts/programs/ocr2/src/lib.rs
@@ -703,7 +703,7 @@ fn transmit_impl<'info>(ctx: Context<Transmit<'info>>, data: &[u8]) -> Result<()
     state.config.latest_transmitter = ctx.accounts.transmitter.key();
 
     // calculate and pay reimbursement
-    let reimbursement = calculate_reimbursement(report.juels_per_lamport, signature_count)?;
+    let reimbursement = calculate_reimbursement(report.juels_per_lamport, signature_count)?; // gjuels
     let amount = reimbursement + u64::from(state.config.billing.transmission_payment_gjuels);
     state.oracles[oracle_idx].payment += amount;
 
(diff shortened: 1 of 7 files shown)


contracts/programs/ocr2/src/lib.rs
```diff
@@ -703,7 +703,7 @@ fn transmit_impl<'info>(ctx: Context<Transmit<'info>>, data: &[u8]) -> Result<()
     state.config.latest_transmitter = ctx.accounts.transmitter.key();
 
     // calculate and pay reimbursement
-    let reimbursement = calculate_reimbursement(report.juels_per_lamport, signature_count)?;
+    let reimbursement = calculate_reimbursement(report.juels_per_lamport, signature_count)?; // gjuels
```
kaleofduty, line 706: Can we rename `reimbursement` to `reimbursement_gjuels`?
Similarly, `calculate_reimbursement` -> `calculate_reimbursement_gjuels`?

contracts/programs/ocr2/src/lib.rs
```diff
@@ -789,7 +789,8 @@ fn calculate_reimbursement(juels_per_lamport: u64, _signature_count: usize) -> R
     let lamports_per_signature = fees.fee_calculator.lamports_per_signature;
     let lamports = lamports_per_signature * SIGNERS;
     let juels = lamports * juels_per_lamport;
-    Ok(juels)
+    let gjuels = juels / 1000000000; // return value as gjuels
```
kaleofduty, line 792: Should we be concerned about overflows in this code?

kaleofduty (changes requested):

contracts/programs/ocr2/src/lib.rs
kaleofduty: (Ditto for other variables in the code that store a value with a unit)

contracts/programs/ocr2/src/lib.rs
archseer: Division can't overflow but it will panic if `juels == 0`. Can use `.checked_div(GIGA).ok_or(ErrorCode::Overflow)?`

archseer: `let juels = lamports * juels_per_lamport;` could use a check, but we enable program-wide overflow checks anyway which will abort: https://github.com/smartcontractkit/chainlink-solana/blob/3813f358930e37874f2b6a0bce14e0ddeec0e8df/contracts/Cargo.toml#L11

archseer (commented):
I'd prefer the relay used `gjuels_per_lamport` but that would require a new core image

LGTM other than the minor comments

contracts/programs/ocr2/src/lib.rs
aalu1418: 👍🏼

contracts/programs/ocr2/src/lib.rs
```diff
@@ -783,30 +789,44 @@ impl Report {
...
-    Ok(juels)
+    let juels = lamports * u128::from(juels_per_lamport);
+    let gjuels = juels / GIGA; // return value as gjuels
+
+    // convert from u128 to u64 with staturating logic to max u64
+    let convert = gjuels.try_into();
+    let gjuels_u64: u64 = match convert {
+        Ok(convert) => convert,
+        Err(_error) => u64::MAX,
+    };
```
aalu1418, line 806: convert back to u64 or if it doesn't fix, use the max u64 value

contracts/programs/ocr2/src/lib.rs
aalu1418: also will division ever overflow? (assuming they are u128 values) i can't figure out how that would be possible

contracts/programs/ocr2/src/lib.rs
```diff
@@ -783,30 +789,44 @@ impl Report {
...
+
+    Ok(gjuels_u64)
```
archseer, line 808: Can be simplified into:

```suggestion
    Ok(gjuels.try_into().unwrap_or(u64::MAX))
```

contracts/programs/ocr2/src/lib.rs
```diff
@@ -783,30 +789,44 @@ impl Report {
...
 }
 
-fn calculate_reimbursement(juels_per_lamport: u64, _signature_count: usize) -> Result<u64> {
-    const SIGNERS: u64 = 1;
+fn calculate_reimbursement_gjuels(juels_per_lamport: u64, _signature_count: usize) -> Result<u64> {
+    const SIGNERS: u128 = 1;
+    const GIGA: u128 = 10u128.pow(9);
     let fees = Fees::get()?;
-    let lamports_per_signature = fees.fee_calculator.lamports_per_signature;
+    let lamports_per_signature: u128 = fees.fee_calculator.lamports_per_signature.into();
```
archseer, line 796:
```suggestion
    let lamports_per_signature = u128::from(fees.fee_calculator.lamports_per_signature);
```

contracts/programs/ocr2/src/lib.rs
```diff
@@ -783,30 +789,44 @@ impl Report {
...
     let lamports = lamports_per_signature * SIGNERS;
-    let juels = lamports * juels_per_lamport;
-    Ok(juels)
+    let juels = lamports * u128::from(juels_per_lamport);
```
archseer, line 798: Actually, `lamports_per_signature` is guaranteed to be a small value (`1 * 5000`) so you could instead convert it here:

```rust
let juels = u128::from(lamports) * u128::from(juels_per_lamport);
```

contracts/programs/ocr2/src/lib.rs
```diff
@@ -703,9 +707,11 @@ fn transmit_impl<'info>(ctx: Context<Transmit<'info>>, data: &[u8]) -> Result<()
...
 
     // calculate and pay reimbursement
-    let reimbursement = calculate_reimbursement(report.juels_per_lamport, signature_count)?;
-    let amount = reimbursement + u64::from(state.config.billing.transmission_payment_gjuels);
-    state.oracles[oracle_idx].payment += amount;
+    let reimbursement_gjuels =
+        calculate_reimbursement_gjuels(report.juels_per_lamport, signature_count)?; // gjuels
+    let amount_gjuels =
+        reimbursement_gjuels.saturating_add(u64::from(state.config.billing.transmission_payment_gjuels));
+    state.oracles[oracle_idx].payment_gjuels = state.oracles[oracle_idx].payment_gjuels.saturating_add(amount_gjuels);
```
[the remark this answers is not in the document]
archseer: Do we want to saturate in such cases or is a panic preferable? A payment of `u64::MAX` would be a serious bug

archseer (approved):

PR #245, opened 2022-03-09
Migrating to ServiceCtx interface

pinebit:
Closes https://app.shortcut.com/chainlinklabs/story/32524/update-solana-package-to-use-servicectx

Final integration PR: https://github.com/smartcontractkit/chainlink/pull/6181


krebernisak (approved):

krebernisak (changes requested):
The CI seems to be failing with the latest changes: https://github.com/smartcontractkit/chainlink-solana/runs/5476415912?check_suite_focus=true

pinebit:
> The CI seems to be failing with the latest changes: https://github.com/smartcontractkit/chainlink-solana/runs/5476415912?check_suite_focus=true

I think it will continue to fail until we get https://github.com/smartcontractkit/chainlink/pull/6181 merged and update chainlink package version.

cmd/monitoring/main.go
```diff
@@ -5,12 +5,18 @@ import (
...
 	relayMonitoring "github.com/smartcontractkit/chainlink-relay/pkg/monitoring"
-	"github.com/smartcontractkit/chainlink-solana/pkg/monitoring"
 	"github.com/smartcontractkit/chainlink/core/logger"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/monitoring"
 )
 
 func main() {
-	coreLog := logger.NewLogger()
+	coreLog, closeLggr := logger.NewLogger()
```
jmank88, line 14: @vyzaldysanchez @samsondav Did we consider usages like this when implementing the debug disk logs?

@topliceanu Where does this command run?

cmd/monitoring/main.go
```diff
@@ -5,12 +5,18 @@ import (
...
+	defer func() {
+		if closeLggr != nil {
+			_ = closeLggr()
+		}
+	}()
```
jmank88, line 19: I think it is safe to drop the nil check here, and if the error is ignored anyways then could simplify further to just:
```suggestion
	defer closeLggr()
```

cmd/monitoring/main.go
```diff
@@ -5,12 +5,18 @@ import (
...
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/monitoring"
 )
 
 func main() {
-	coreLog := logger.NewLogger()
+	coreLog, closeLggr := logger.NewLogger()
+	defer func() {
+		if closeLggr != nil {
+			_ = closeLggr()
```
samsondav, line 17: Why do we throw away the error? If this errors we probably want to log it e.g. with `log.Println` or `log.Error`

cmd/monitoring/main.go
pinebit: In this case I get a linter error, that's why I added this check.

jmank88: Hmm ok. Does it accept this?
```suggestion
	defer func() {
		_ = closeLggr()
	}()
```

cmd/monitoring/main.go
pinebit: Fixed

cmd/monitoring/main.go
pinebit: See the comment from @samsondav below, I changed this to log the error, because we should not mute it anyway.

cmd/monitoring/main.go
```diff
@@ -2,20 +2,30 @@ package main
...
 func main() {
-	coreLog := logger.NewLogger()
-	log := logWrapper{coreLog}
+	coreLog, closeLggr := logger.NewLogger()
+	defer func() {
+		if closeLggr != nil {
+			if err := closeLggr(); err != nil {
+				log.Println(fmt.Sprintf("Error while closing Logger: %v", err))
+			}
+		}
```
jmank88, line 22: I think this nil check is still unnecessary, no?
```suggestion
		if err := closeLggr(); err != nil {
			log.Println(fmt.Sprintf("Error while closing Logger: %v", err))
		}
```

jmank88 (approved):

cmd/monitoring/main.go
pinebit: Indeed this was odd check, removed.

jmank88 (approved):

cmd/monitoring/main.go
samsondav: @vyzaldysanchez

pinebit:
@krebernisak do you have any other concerns that would prevent this from being merged?
Can we discuss these in a channel with me, @jmank88 and @samsondav ?
Thanks!

krebernisak (approved):

cmd/monitoring/main.go
vyzaldysanchez: I don't think we considered this kind on usages when we implemented the disk logs. But I consider this will work just fine anyways, even for a one time usage, unless there are more things we would need to consider when the logger gets used like this.

vyzaldysanchez (approved):

krebernisak (approved):

PR #246, opened 2022-03-09
fix: allow negative observations, error if bounds exceeded

aalu1418:


pkg/solana/marshal_signed_int.go
```diff
@@ -0,0 +1,67 @@
+// code from: https://github.com/smartcontractkit/chainlink-terra/blob/develop/pkg/terra/marshal_signed_int.go
+// will eventually be removed and replaced with a generalized version from libocr
```
archseer, line 2: Can we get this into libocr now rather than using this placeholder?

aalu1418: ideally yes, but there's no timeline on a new a libocr release

archseer (approved):

krebernisak (approved):

PR #250, opened 2022-03-14
[WIP: do not merge] chains and nodes refactor

aalu1418:


pkg/solana/db/db.go
```diff
@@ -0,0 +1,61 @@
...
+	Name          string
+	SolanaChainID string `json:"solanaChainId" db:"solana_chain_id"`
+	SolanaURL     string `json:"solanaURL" db:"solana_url"`
+}
+
+type Node struct {
+	ID            int32
+	Name          string
+	SolanaChainID string `json:"solanaChainId" db:"solana_chain_id"`
+	SolanaURL     string `json:"solanaURL" db:"solana_url"`
```
jmank88, line 48: The chain id 'Terra/Solana' prefix is sort of a quirk since that field is a foreign key. Probably ok to just use URL here? Unless there are multiple types?

aalu1418: discussed: keeping the naming for now to be explicit (this is likely to be removed anyway from the DB in future configuration improvement work)

pkg/solana/config/config.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/db"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
+)
+
+// Global terra defaults.
+var defaultConfigSet = configSet{
+	BlockRate:           500 * time.Millisecond, // approximate slot rate
```
aalu1418, line 15: should this be slowed down? this is tied to how often the balance monitor queries for updated balance

pkg/solana/types.go
```diff
@@ -162,21 +162,12 @@ type TransmissionV1 struct {
 // CL Core OCR2 job spec RelayConfig member for Solana
 type RelayConfig struct {
 	// network data
-	NodeEndpointHTTP string `json:"nodeEndpointHTTP"`
+	ChainID  string `json:"chainID"`  // required
+	NodeName string `json:"nodeName"` // optional, defaults to a random node with ChainID
```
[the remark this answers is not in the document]
jmank88: Just the ability to specific a single node by name. That is not available on EVM, and it also conflicts with the new EVM failover, which monitors the health of the nodes in a pool and swaps them in and out behind a single client. We could merge the two in some way of course (e.g. ordered csv of names plus fallback to random), and we may have to eventually :shrug:

pkg/solana/transmitter.go
```diff
@@ -84,21 +84,10 @@ func (c *ContractTracker) Transmit(
...
-			c.client.skipPreflight,
-			c.client.commitment,
-		)
-
-		if err != nil {
-			c.lggr.Errorf("error on Transmit.SendAndConfirmTransaction: %s", err.Error())
-			return
+		// send to tx manager
+		if err := c.txManager.Enqueue(c.StateID.String(), tx); err != nil {
+			c.lggr.Errorf("error on Transmit.txManager.Enqueue: %s", err.Error())
```
jmank88, line 89: Since `Enqueue` is a non-blocking channel send attempt, could this be called syncronously instead of from a spawned goroutine?

go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/smartcontractkit/chainlink-solana
 go 1.17
 
 require (
-	github.com/gagliardetto/binary v0.5.2
-	github.com/gagliardetto/solana-go v1.0.4
+	github.com/gagliardetto/binary v0.6.1
+	github.com/gagliardetto/solana-go v1.4.0
```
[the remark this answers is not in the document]
aalu1418: > binary diff: https://github.com/gagliardetto/binary/compare/v0.5.2...v0.6.1
notes:
* documentation updates (no impact)
* convert BE/LE encoding from functions to vars (minimal impact, if any)
* refactor to Uint128/Int128 implementation
  * ability to encode/decode BE or LE values
  * better nil value/pointer handling
  * test case included!

overall assessment:
* changes look fine, mostly a refactor of Uint128/Int128 that adds capability and further broken down into helper functions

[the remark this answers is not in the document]
aalu1418: > solana-go diff: https://github.com/gagliardetto/solana-go/compare/v1.0.4...v1.4.0

impact:
* very large docs update
* keys changes: splitting a function + PublicKeySlice additional functions (not used)
* message changes: split into separate file + some additional helpers
* token program: additional features (i don't believe this is used)
* instruction registry: refactor + clean up (i don't believe this is used)
* examples: `getLatestBlockhas` is a relevant new one (but this code has no direct impact)
* rpc methods: new functions, cleanup + comments (no impact)
* transactions: clean up + minimal code changes

overall assessment:
* changes also look fine, mostly adding some new capabilities, and moving code around

pkg/solana/types.go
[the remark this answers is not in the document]
krebernisak: Auto-failovers are something that we'll need to do in the short term, and then the question is what's the use case of pointing a specific job to go through a specific RPC node, and how will that work in the context of auto-failovers 🤔

I vote to simplify, remove this option for now, and let's revisit in the context of auto-failovers.

pkg/solana/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+	ctx, cancel := context.WithTimeout(ctx, c.contextDuration)
+	defer cancel()
+	opts.Commitment = c.commitment // overrides passed in value - use defined client commitment type
+	return c.rpc.GetAccountInfoWithOpts(ctx, addr, opts)
+}
+
+func (c *Client) LatestBlockhash() (*rpc.GetLatestBlockhashResult, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), c.contextDuration)
+	defer cancel()
+	return c.rpc.GetLatestBlockhash(ctx, c.commitment)
```
krebernisak, line 101: Why don't we also use the `requestGroup` here to wrap this call?

pkg/solana/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+		log:             log,
+		requestGroup:    &singleflight.Group{},
+	}, nil
+}
+
+func (c *Client) Balance(addr solana.PublicKey) (uint64, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), c.contextDuration)
+	defer cancel()
+
+	v, err, _ := c.requestGroup.Do("GetBalance", func() (interface{}, error) {
```
krebernisak, line 69: Coalescing here depends on the input `addr` parm, so having a general `"GetBalance"` group will group them all, ignoring the arg. 

We should include the arg in the group name as `"GetBalance(${addr})"`.

pkg/solana/types.go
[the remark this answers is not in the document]
aalu1418: 👍🏼 we can remove it from the spec

so my understanding is that it will basically pick a random node/endpoint each time?
or would this change propagate further and only allow one node to be configured per chain?

pkg/solana/config/config.go
aalu1418: ah ok, so better to rename this to BalancePollPeriod. we don't really use the BlockRate parameter at all besides polling for balances (ported from Terra)

pkg/solana/client/client.go
aalu1418: ooooh good point, otherwise it would return incorrectly if you were querying for two different balances

pkg/solana/types.go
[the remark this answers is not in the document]
aalu1418: removed node name: https://github.com/smartcontractkit/chainlink-solana/pull/250/commits/8d722896e06dd07f7be80fd5bb4ade83843ba986

pkg/solana/client/client.go
aalu1418: added: https://github.com/smartcontractkit/chainlink-solana/pull/250/commits/b0b22cdca27331bc30a484ff80a1f706c46ae61e

pkg/solana/client/client.go
aalu1418: fixed: https://github.com/smartcontractkit/chainlink-solana/pull/250/commits/b0b22cdca27331bc30a484ff80a1f706c46ae61e

pkg/solana/config/config.go
aalu1418: renamed parameter, and changed to 5s poll rate: https://github.com/smartcontractkit/chainlink-solana/pull/250/commits/b7c508bd4823b4d7a58c376b1b98d3983ce45101

pkg/solana/client/client.go
```diff
@@ -0,0 +1,152 @@
...
+
+type Client struct {
+	rpc             *rpc.Client
+	skipPreflight   bool // to enable or disable preflight checks
+	commitment      rpc.CommitmentType
+	txTimeout       time.Duration
+	contextDuration time.Duration
+	log             logger.Logger
+
+	// provides a duplicate function call suppression mechanism
```
connorwstein, line 50: is this still needed? Seems like we already have caching at a layer higher and this just makes things harder to reason about

pkg/solana/client/client_test.go
```diff
@@ -0,0 +1,111 @@
...
+		"4uhcVJyU9pJkvQyS88uRDiswHXSCkY3zQawwpjk2NsNY", // testnet
+		"5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d", // mainnet
+		"GH7ome3EiwEr7tu9JuTh2dpYWBJK3z69Xm1ZE3MEE6JC", // localnet (random)
+	}
+	networks := []string{"devnet", "testnet", "mainnet", "localnet"}
+	hashCounter := 0
+
+	getFeeNil := false
+
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
connorwstein, line 35: no way to test against a local node here? Would make the test significantly higher fidelity

pkg/solana/config/config.go
```diff
@@ -0,0 +1,152 @@
...
+		default:
+			c.lggr.Warnf(invalidFallbackMsg, "CommitmentType", str, c.defaults.Commitment, nil)
+			commitment = rpc.CommitmentConfirmed
+		}
+		return commitment
+	}
+	return c.defaults.Commitment
+}
+
+const invalidFallbackMsg = `Invalid value provided for %s, "%s" - falling back to default "%s": %v`
```
connorwstein, line 152: probably better to keep inline bc a) only used in one place b) user has to infer how many fields are required

pkg/solana/transmitter.go
```diff
@@ -82,25 +82,9 @@ func (c *ContractTracker) Transmit(
...
-
-		if err != nil {
-			c.lggr.Errorf("error on Transmit.SendAndConfirmTransaction: %s", err.Error())
-			return
-		}
-		// TODO: poll rpc for tx confirmation (WS connection unreliable)
-		c.lggr.Debugf("tx signature from Transmit.SendAndConfirmTransaction: %s", txSig.String())
-	}()
-	return nil
+	// pass transmit payload to tx manager queue
```
connorwstein, line 85: probably want a log here, I don't see one on the other side

pkg/solana/client/client.go
aalu1418: i think it still is, the caching is only for the state pollling but doesn't include anything like blockhash, balances, etc

pkg/solana/config/config.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
+)
+
+// Global terra defaults.
+var defaultConfigSet = configSet{
+	BalancePollPeriod:   5 * time.Second, // poll period for balance monitoring
+	ConfirmPollPeriod:   time.Second,     // polling for tx confirmation
+	OCR2CachePollPeriod: time.Second,     // cache polling rate
+	OCR2CacheTTL:        time.Minute,     // stale cache deadline
+	TxTimeout:           time.Minute,     // transaction timeout
```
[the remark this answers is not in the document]
aalu1418: should be, solana generates blocks around 400ms per block so we should see some sort of result after a minute

pkg/solana/client/client_test.go
aalu1418: agreed, we currently don't have something to test it with at the moment, but it's a need to also test the tx manager.

will be taking a look at it today or tomorrow

pkg/solana/client/client.go
connorwstein: are there rpc problems if we do lots of blockhash/balance queries though?

aalu1418: we have yet to hit a limit, but i suspect that it could become an issue as we try to scale the number of feeds and decrease the round time.

right now we're querying state (3 calls per second per feed) + transmitting (2 calls blockhash + transmit, per feed) + querying balance (1 call per 5 second, separate)
and in the future we'd probably need to go faster on the state and transmitting calls...so having coalescing on certain calls seems to be useful?

ops/Pulumi.localnet.yaml
```diff
@@ -11,7 +11,7 @@ config:
   solana-env:CL-PORT-START: 6688 # start port to expose lite client port
   solana-env:CL-P2P_PORT-START: 5050 # start port to expose lite client P2P ports
   solana-env:CL-NODE_VERSION: develop.2e585e2dfc3fab13bcad988f1e3dfb9db3495c0e # development version can be found at https://gallery.ecr.aws/chainlink/chainlink
-  solana-env:CL-BUILD_LOCALLY: false # (overrides NODE_VERSION) BUILD_CONTEXT and BUILD_DOCKERFILE must be set if this is enabled
+  solana-env:CL-BUILD_LOCALLY: true # (overrides NODE_VERSION) BUILD_CONTEXT and BUILD_DOCKERFILE must be set if this is enabled
```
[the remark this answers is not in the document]
samsondav: Seems like this has to be removed?>

pkg/solana/client/client.go
connorwstein: the coalescing feels like a crutch which masks the real load were sending and the load the node can handle. Imo we should confirm the exact query load the node can handle, then update our cache refresh period and add explicit rate limiting accordingly

krebernisak: Connor, do you see any potential issues if the client actually coalesces requests?

I do agree that most of the caching, as currently implemented, is on the higher level, and we'll probably coalesce only `blockhash` calls while running feeds. But I don't see any problems if the lower-level client actually contributes to the overall performance optimization of the system in this way, do you?

Not to mention other parts of the system like monitoring, e2e tests, which should ideally all use a single (type, not instance) functional and robust client. Those systems can then execute parallel calls to the client, without worrying about call de/duplication.

In any case, as we have larger performance efforts incoming, I'd suggest we leave it as it was for now and address it if necessary in the follow-up (performance-focused PR).

pkg/solana/config/config.go
```diff
@@ -0,0 +1,150 @@
...
+	"sync"
+	"time"
+
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/db"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
+)
+
+// Global terra defaults.
```
jmank88, line 13:
```suggestion
// Global solana defaults.
```

krebernisak (approved):

jmank88 (approved):

pkg/solana/client/client.go
connorwstein: I don't think there are any implications for correctness with the overall system right now, so fine to leave as is. My only concern with it is just unjustified complexity, I think having a dumb client and caching a layer above is sufficient for our use cases. I think it'd be totally fine to add it to the client if the requests it was handling were stateless (which is I think its main use case), but they are not, so it runs the risk of returning staler than necessary blockhashes/nonces etc.

PR #255, opened 2022-03-17
Additional test cases

aalu1418:


pkg/solana/types.go
```diff
@@ -90,7 +90,7 @@ type Oracles struct {
 }
 
 func (o Oracles) Data() ([]Oracle, error) {
-	if o.Len > MaxOffchainConfigLen {
+	if o.Len > MaxOracles {
```
krebernisak, line 93: Nice catch! 🎉

krebernisak (approved):

pkg/solana/types.go
aalu1418: those test cases coming in handy 😆

PR #256, opened 2022-03-18
asdf and ginkgo updates to bring more in line with terra

tateexon:
- Brings things more in sync with terra.
- I could not figure out how to get the gauntlet.yml, and the e2e yml github action jobs to run nix without have an hour startup for the first command on every run. Tried using caches and such in different ways to no avail. It does just seem to work for the relay.yml so I left that using nix.
- I did fix up the gauntlet and e2e ymls so all we need to do is drop in the nix install and prepend the nix develop call so it should be an easy update once we get nix issues figured out.
- I left us at go 1.17.x for now even though 1.18 is out since we probably don't want to add that extra complexity to testing this close to release

NOTE: I will need someone with permissions to update the required checks to use the updated gauntlet.yml jobs.


.github/workflows/chaos.yml
```diff
@@ -23,19 +23,22 @@ jobs:
       - uses: actions/checkout@v2
       - run: echo $HOME
       - run: echo $PATH
-      - run: |
+      - name: Set Npm Config
+        run: |
           npm config set -- '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
```
krebernisak, line 28: Let's remove all mention of the `NPM_TOKEN`. We shouldn't need it anymore as `gauntlet-core` is now public. We still need to `yarn install` though :)

.github/workflows/chaos.yml
```diff
@@ -23,19 +23,22 @@ jobs:
...
         working-directory: contracts
-      - uses: actions/upload-artifact@master
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@master
         with:
           name: target
           path: contracts/target/deploy
-  build_custom_cl_image:
-    name: Build custom CL image
+  Chaos_Testing_Build_Custom_Chainlink_Image:
```
krebernisak, line 40: It's a nitpick, but I'd prefer `snake_case` here, to be consistent with the style of the workflow file, and then explicit names for all workflow (top lvl) and jobs.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,42 @@
+name: golangci-lint
+
+on:
+  push:
+    branches:
+      - develop
+      - main
+    paths:
+      - '**.go'
```
krebernisak, line 9: Will this only run if '**.go' files are touched?

If that's the case, then we can't enable this as a global check for develop branch. Consider removing this filter, for now, so we enable this check globally.

.github/workflows/chaos.yml
```diff
@@ -11,22 +11,20 @@ on:
...
       image: projectserum/build:v0.22.1
       env:
         RUSTUP_HOME: "/root/.rustup"
         FORCE_COLOR: 1
-        NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
     steps:
       - uses: actions/checkout@v2
       - run: echo $HOME
       - run: echo $PATH
       - name: Set Npm Config
```
krebernisak, line 26: Remove this name here, as other `yarn install` steps don't have an explicit name.

scripts/tool-versions-to-env.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "" > ${ENV_FILE}
+    while read line; do
+        to_echo ${1} "Original line: $line"
+        lineArray=($line)
+        key=${lineArray[0]^^}
+        value=${lineArray[1]}
+        # ignore comments
+        if [[ ${key:0:1} != "#" ]]; 
+        then 
+
```
krebernisak, line 32:
```suggestion
```

scripts/tool-versions-to-env.sh
```diff
@@ -0,0 +1,49 @@
...
+# Second argument is the string to echo
+to_echo() {
+    if [ $1 -eq 1 ]
+    then
+        echo "$2"
+    fi
+}
+
+# First argument is the boolean of 0 = false or 1 = true to echo
+read_file() {
```
krebernisak, line 22:
```suggestion
tools_to_env() {
```

or 

```suggestion
rw_file() {
```

scripts/tool-versions-to-env.sh
```diff
@@ -0,0 +1,49 @@
...
+then
+    # If argument is passed only echo the version number for the specified library to std out
+    read_file 0
+    source ${ENV_FILE}
+    echo "${!1^^}"
+else
+    # If no argument is passed just echo all the read in versions to std out
+    read_file 1
+    echo "Run 'source .env' to load environment variables into your shell"
+fi
```
krebernisak, line 49: Not sure I understand these two options? Do you mind explaining the use cases, and please consider extending improving the in-line comments to be more descriptive.

krebernisak (dismissed):

krebernisak (approved):

PR #296, opened 2022-04-10
Add inspection docs and allow ocr2:inspect to run without RDD flag

kylesmartin:


gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -136,6 +134,38 @@ export default class OCR2Inspect extends SolanaCommand {
...
+          - Transmitter: ${oracle.transmitter}
+          - Proposed Payee: ${oracle.proposedPayee}
+          - From Round ID: ${oracle.fromRoundId}
+          - Payment Gjuels: ${oracle.paymentGjuels}
+      `,
+      )
+    })
+
+    // If rdd input doesn't exist, just print config
+    if (!this.flags.rdd) {
```
RodrigoAD: `input` generation is independent from RDD. It can be generated with RDD, but also with the `--input` flag directly, or in the future through some other way. What we really want to check here, is if the `makeInput` was able to generate the input correctly. 
```suggestion
    const input = this.makeInput(this.flags.input)
    // We can assume that if one field is not provided, the rest aren't either
    if (!input.description) {
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
```diff
@@ -78,6 +78,33 @@ export default class ProposeOffchainConfig extends SolanaCommand {
   }
 
   static makeInputFromRDD = (rdd: any, stateAddress: string): OffchainConfig => {
+    if (!rdd) {
```
RodrigoAD: We should expect that we always receive `rdd` correctly here. We should check for it before calling `makeInputFromRDD`

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -41,6 +41,49 @@ export default class OCR2Inspect extends SolanaCommand {
     const billingAccessController = this.flags.billingAccessController || process.env.BILLING_ACCESS_CONTROLLER
     const requesterAccessController = this.flags.requesterAccessController || process.env.REQUESTER_ACCESS_CONTROLLER
 
+    // Return empty input if no rdd or user input provided
+    if (!rddPath) {
+      return {
+        description: '',
```
krebernisak: Hmm, this empty default should be moved outside this specific command to something like `/contracts/ocr2/types.ts` where we group all OCRw types/defaults together.

docs/InspectingContracts.md
```diff
@@ -0,0 +1,148 @@
...
+ℹ️   Transmission Payment: 1
+ℹ️   Observation Payment: 1
+ℹ️   Requester Access Controller: 5vkHdxPiTyfY5VdRpPu8tNTpeik6Cy93M6GzztMPWfAk
+ℹ️   Billing Access Controller: CBWXiwPGX6ykWtPGXp4cuFJP53SW81pe9q1YfUASWC46
+✨  Done in 12.61s.
+```
+
+## Via Solana Explorer
+
+[Solana Explorer](https://explorer.solana.com/) allows users to search for deployed programs and read their state. On the home page, search the address of the program account that you want to inspect. For example, if you want to inspect the OCR2 program for the BTC/USD feed on mainnet, search *2oyA8ZLwuWeAR5ANyDsiEGueUyDC8jFGFLSixSzT9KtV*.
```
krebernisak, line 80: Consider also mentioning another option: https://solscan.io

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -136,6 +177,37 @@ export default class OCR2Inspect extends SolanaCommand {
...
+          - From Round ID: ${oracle.fromRoundId}
+          - Payment Gjuels: ${oracle.paymentGjuels}
+      `,
+      )
+    })
+
+    const input = this.makeInput(this.flags.input)
+    // If input does not exist, just print config
+    if (!input.description) {
+      logger.info(`Min Answer: ${onChainState.config.minAnswer}`)
```
krebernisak: Just few lines before we group all logs together:

```javascript
logger.info(
        `Oracle Info:
          - Transmitter: ${oracle.transmitter}
          - Proposed Payee: ${oracle.proposedPayee}
          - From Round ID: ${oracle.fromRoundId}
          - Payment Gjuels: ${oracle.paymentGjuels}
      `,
      )
```

We should define a standard way to log large data objects like this, so we get a unified look & feel at the output.

krebernisak (commented):
Love the docs <3 

Those are going to become very important as we look to scale our integration efforts.

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+  maxDurationQueryNanoseconds: number
+  maxDurationObservationNanoseconds: number
+  maxDurationReportNanoseconds: number
+  maxDurationShouldAcceptFinalizedReportNanoseconds: number
+  maxDurationShouldTransmitAcceptedReportNanoseconds: number
+  configPublicKeys: string[]
+}
+
+export type InspectionInput = {
```
RodrigoAD: I imagine this change comes from the `NewTransmission` type, but this is not a common type, is only useful in its own context (the command) and I don't think should be in a common place. Also, if we bring every command input type here, the list can be unreadable. 
Let's have the Input types closer to where they are used for now. Same for `OffchainConfig`

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  offchainConfig: OffchainConfig
+  billingAccessController: string
+  requesterAccessController: string
+  billing: {
+    observationPaymentGjuels: string
+    transmissionPaymentGjuels: string
+  }
+}
+
+export const emptyInspectionInput: InspectionInput = {
```
RodrigoAD: I don't think this is necessary. The only use case seems here: https://github.com/smartcontractkit/chainlink-solana/pull/296/files#diff-f922fe4f599b5db4b4ac741811f33d24961406183e46286ce9e564a4144da114R30
But why don't we simply give `makeInput` the possibility to return `undefined`?
```ts
makeInput = (userInput): Input | undefined => {...}

...
   const input = this.makeInput(this.flags.input)
   if (!input) { // something
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/types.ts
kylesmartin: Going to remove the types file for now until we decide on how we want to restructure going forward. I'll stick with defining objects where they are used for now.

RodrigoAD (approved):
LGTM

PR #297, opened 2022-04-10
speed up smoke test, add soak test

skudasov:


gheorghestrimtu (dismissed):

tests/e2e/chaos/chainlink-relay-sol.yaml
```diff
@@ -9,14 +9,15 @@ charts:
   chainlink:
     index: 4
     values:
-      replicas: 19
+      replicas: 5
```
aalu1418, line 12: why reduce this down to 5? ideally we would like to test with what we use on mainnet (16 oracles + 1 bootstrap)

tests/e2e/chaos/chainlink-relay-sol.yaml
```diff
@@ -9,14 +9,15 @@ charts:
...
-          version: "develop.4bc99afb22dca23469d9a7bc2bdfa7ea49028ae6"
+          image: "public.ecr.aws/z0b1w9r9/chainlink"
+          version: "develop"
       env:
         eth_url: "ws://sol:8990"
         eth_disabled: "true"
+        SOLANA_ENABLED: "true"
         USE_LEGACY_ETH_ENV_VARS: "true"
         feature_external_initiators: "true"
         FEATURE_OFFCHAIN_REPORTING2: "true"
```
aalu1418, line 23: Might be good to double check some of these env vars? the pulumi env has the latest ones I've been using: https://github.com/smartcontractkit/chainlink-solana/blob/develop/ops/Pulumi.localnet.yaml#L19

`ETH_DISABLED` and `USE_LEGACY_ETH_ENV_VARS` were deprecated in favor of:
    - EVM_ENABLED = false
    - EVM_RPC_ENABLED = false
    
also, i don't think `external_initiators` need to be enabled still? (unless that is used for something i'm not aware about)

tests/e2e/chaos/chainlink-relay-sol.yaml
skudasov: It's a local run template, CI is still running 19 nodes, but we are planning to shrink it soon.

tests/e2e/solclient/solclient.go
```diff
@@ -157,20 +153,6 @@ func NewClient(cfg *NetworkConfig) (*Client, error) {
...
-		Store:         solana.NewWallet(),
-		Feed:          solana.NewWallet(),
-		Proposal:      solana.NewWallet(),
-		Owner:         solana.NewWallet(),
-		Mint:          solana.NewWallet(),
-		MintAuthority: solana.NewWallet(),
-		OCRVault:      solana.NewWallet(),
-		Authorities:   make(map[string]*Authority),
-	}
-}
```
gheorghestrimtu, line 172: Please let me know where I can find OCR, Owner, and Feed in the new changes

skudasov: In ContractDeployer, each instance has all the accounts and authorities.

aalu1418 (approved):

PR #298, opened 2022-04-12
new functions in solana client

aalu1418:
* test case for checking tx deduplication
* add `getSignatureStatuses` and `simulateTransaction` to client


aalu1418:
Note: there is a bug in decoding the response in solana-go for `simulateTransaction` so the current PR will not pass until solana-go is updated

Addressed here: https://github.com/gagliardetto/solana-go/pull/56

_RESOLVED_ 🎉

pkg/solana/client/client.go
```diff
@@ -151,6 +153,43 @@ func (c *Client) GetFeeForMessage(msg string) (uint64, error) {
...
+	}
+	return res.Value, nil
+}
+
+// https://docs.solana.com/developing/clients/jsonrpc-api#simulatetransaction
+func (c *Client) SimulateTx(ctx context.Context, tx *solana.Transaction) (*rpc.SimulateTransactionResult, error) {
+	ctx, cancel := context.WithTimeout(ctx, c.contextDuration)
+	defer cancel()
+
+	res, err := c.rpc.SimulateTransactionWithOpts(ctx, tx, &rpc.SimulateTransactionOpts{
```
krebernisak: Consider adding an optional input for `opts` so the user could use them, or if `nil` the fn uses these defaults.

pkg/solana/client/client_test.go
```diff
@@ -112,3 +115,150 @@ func TestClient_Reader_ChainID(t *testing.T) {
...
+	// check signature statuses
+	time.Sleep(2 * time.Second) // wait for processing
+	statuses, err := c.SignatureStatuses(ctx, []solana.Signature{sigSuccess, sigFail})
+	assert.NoError(t, err)
+
+	assert.Nil(t, statuses[0].Err)
+	assert.NotNil(t, statuses[1].Err)
+}
+
+func TestClient_SendTxDuplicates_Integration(t *testing.T) {
```
krebernisak, line 203: Great job confirming this!

krebernisak (dismissed):
Looks good to go, once `solana-go` is updated. Nice catch there!

pkg/solana/client/client.go
aalu1418: added!

krebernisak (approved):

PR #307, opened 2022-04-14
Add actionlint to lint gh action workflows

tateexon:


.github/workflows/chaos.yml
```diff
@@ -20,9 +20,9 @@ jobs:
         RUSTUP_HOME: "/root/.rustup"
         FORCE_COLOR: 1
     steps:
-      - uses: actions/checkout@v2
-      - run: echo $HOME
-      - run: echo $PATH
+      - uses: actions/checkout@v3
```
aalu1418: just wondering, why upgrading to v3?

seems like core is still on v2...are there any security implications about going from v2 => v3?

aalu1418 (approved):

PR #312, opened 2022-04-22
Logging improvements in Multisig

RodrigoAD:
Some improvements:
- Multisig state is always printed in a more clear way
- Proposal state is printed in a clear way if exists
- Stricter types for msig and proposal states
- Improved error handling in simulate tx

### DIFF
--- a/gauntlet/packages/gauntlet-serum-multisig/src/commands/multisig.ts
+++ b/gauntlet/packages/gauntlet-serum-multisig/src/commands/multisig.ts
@@ -242,8 +242,8 @@ export const wrapCommand = (command) => {
 
     createProposalAcount = async (): Promise<PublicKey> => {
       await prompt('A new Multisig Proposal account will be created. Continue?')
-      logger.log('Creating Multisig Proposal account...')
       const proposal = Keypair.generate()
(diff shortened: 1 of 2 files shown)


gauntlet/packages/gauntlet-serum-multisig/src/commands/multisig.ts
```diff
@@ -149,53 +173,77 @@ export const wrapCommand = (command) => {
...
       }
 
-      const isAlreadyExecuted = proposalState.didExecute
-      if (isAlreadyExecuted) throw new Error('Multisig Proposal is already executed')
-
-      this.require(
-        await this.isSameProposal(proposal, rawTx),
-        'The transaction generated is different from the Multisig Proposal provided',
-      )
+      if (proposalState.isExecuted)
```
akhilchainani: should this be `!proposalState.isExecuted`? Seems like before, we'd error if the proposal was already executed

RodrigoAD: Oh nice catch, you're totally right!

kylesmartin (approved):
With Akhil's change everything looks good. I tested all the serum_multisig commands locally and everything works as expected.

PR #323, opened 2022-04-29
RFC: chainlink api accepts Pubkey, not AccountInfo, and explicitly verify the program id matches id()

neery1218:
Hello! Nice to meet you all :)

1. assert that the chainlink pubkey matches crate::id() to avoid a potential vulnerability
2. change the chainlink interface  to accept a Pubkey instead of an AccountInfo. Passing an AccountInfo as an argument is clunky, since our instructions have to accept the chainlink program id as an account. IMO, The standard thing to do is to pass in the program Pubkey. See the spl token mint instruction as an example: https://github.com/solana-labs/solana-program-library/blob/master/token/program/src/instruction.rs#L787. Then, we can call the chainlink interface by doing:
```rust
        chainlink_solana::latest_round_data(chainlink_solana::id(), chainlink_feed_account)?;
```

### DIFF
--- a/contracts/crates/chainlink-solana/src/lib.rs
+++ b/contracts/crates/chainlink-solana/src/lib.rs
@@ -9,6 +9,7 @@ use solana_program::{
     account_info::AccountInfo,
     instruction::{AccountMeta, Instruction},
     program::invoke,
+    entrypoint::ProgramResult,
     program_error::ProgramError,
     pubkey::Pubkey,
 };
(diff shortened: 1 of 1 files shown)


neery1218:
didn't attempt to fix any tests or examples bc i wanted feedback first. But if this is good to go im happy to fix everything!

neery1218:
Actually on second thought, I'll still need to pass in the chainlink program account to my instructions bc we're making a CPI call. I'm not sure why chainlink::latest_round_data() has to be an instruction. This is just reading an account, correct? Is there a different interface that lets me read chainlink oracle data on solana _without_ calling an instruction?

neery1218:
regardless, i think (1) is useful. I can remove the other stuff, lmk!

aalu1418:
Hello @neery1218! Thanks for raising your concerns and a potential fix in your PR.

In the future, please do not disclose potential vulnerabilities via Github Issues or PRs as this is a public forum. Please use our Hacker1 or Immunefi bug bounty programs: https://blog.chain.link/expanding-the-chainlink-bug-bounty-program-to-immunefi/

Regarding your concerns and questions:
> assert that the chainlink pubkey matches crate::id() to avoid a potential vulnerability

The reason for the missing check for program_id is so we/others can use different program_ids across devnet/mainnet. There are the production program_ids for users to work with, but there are also other sets of programs that we and other external parties can use for development and testing.

> I'm not sure why chainlink::latest_round_data() has to be an instruction. This is just reading an account, correct?|

Mostly correct, it is reading an account based on the underlying parsing method. This ensures that the data is returned in a stable layout, allowing us to change how the storage account looks in the future as a sort of proxy feature.

Please feel free to ask more questions, but I will be closing out this PR now as this is not something that we are looking to change at the moment :slightly_smiling_face:

neery1218:
Thanks Aaron! This makes sense

PR #326, opened 2022-05-02
use refactored gauntlet-core Logger class

akhilchainani:


RodrigoAD:
Can we include the address book in this PR as well?

kylesmartin:
Updated imports look good

PR #329, opened 2022-05-03
Refactor EnvelopeSource.Fetch() and add tests

topliceanu:
From the logs:

```
[ERROR] failed Export component=feed-monitor error=runtime error: invalid memory address or nil pointer dereference feed_id=EaWU3CL55JVf3Rf92VLNjqL4PGLT57y7LN4diFD38qSv feed_name=SOL / USD index=0 logger=unset@unset network=solana-testnet-devnet
```

Potential candidates for this nil pointer crash:
    - [x] GetMultiply() - can't be this because all feeds have it!
    - [x] LinkBalance - can't be nil. 
    - [x] LatestAnswer - could be nil if the latest transmission doesn’t have a Data field
    - [x] LinkAvailableForPayment - computed in the client. Can be nil if linkBalance is nil. 
    - [x] JuelsPerFeeCoin - is always set to 0


pedrovsg (approved):
lgtm

PR #331, opened 2022-05-04
Feature/max retries configuration parameter

sdrug:


pkg/solana/config/config.go
```diff
@@ -21,6 +21,7 @@ var defaultConfigSet = configSet{
 	TxConfirmTimeout:    15 * time.Second,       // duration before discarding tx as unconfirmed
 	SkipPreflight:       true,                   // to enable or disable preflight checks
 	Commitment:          rpc.CommitmentConfirmed,
+	MaxRetries:          nil, // max number of retries, when nil - rpc node will do a reasonable number of retries
```
aalu1418: hmmm what do you think about having this default to `0`? since the sendTx function is primarily used in the context of the transaction manager and that implements custom retry logic

ideally it would default to 0, but it also introduces side-effects like we may always have to specify a max retries value if we want to use this outside the txm (setting it to nil wouldn't work anymore as it would just use the default)

aalu1418 (changes requested):
Looks great! just a thought on changing the default 😄

pkg/solana/config/config.go
sdrug: Yeah, having it `0` by default would mean we won't be able to set it to `nil` and thus no way to let the rpc node decide  on the number of retries (default rpc node functionality). I would prefer setting `0` explicitly in config since we change the default rpc node's behaviour.

sdrug: But in case it's somehow difficult to set up the config, let's have it `0`, it's one-liner anyways

aalu1418: we could introduce an function (like `SendTxWithOpts`) where `opts` can be passed in? if that is nil, then it uses the default set (where we can leave `maxRetries` as nil, sort of like how solana-go manages it)

and `SendTx` would just wrap `SendTxWithOpts` where `opts = nil`

(or instead of the whole `opts` struct, we can just pass the `maxRetries` param)

aalu1418: new suggestion from @krebernisak that i also like:
* set default `maxRetries = 0`
* if user configures parameter to be negative, use `maxRetries = nil` rather than default (if it's unset, it will continue to be 0)

(also document w/ comments!)

sdrug: Alright, I can do it this way! Is there place somewhere in the doc where I can describe it?

aalu1418: for now, i would say leave some comments on the function...i just looked at files in the docs folder and they are out of date (they are from before the nodes/chains refactor, I will update them once we merge this PR)

aalu1418 (approved):

PR #338, opened 2022-05-12
Median provider

connorwstein:


pkg/solana/contract_test.go
```diff
@@ -148,15 +147,10 @@ func TestGetLatestTransmission(t *testing.T) {
 }
 
 func TestStatePolling(t *testing.T) {
-	i := atomic.NewInt32(0)
-	wait := 5 * time.Second
```
connorwstein, line 154: I don't think this is worth introducing a time dependence test just to test the polling for loop. Just made it explicit calls to get the values. Loop is tested in the integration tests anyways

pkg/solana/state_cache.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+type StateCache struct {
+	utils.StartStopOnce
+	// on-chain program + 2x state accounts (state + transmissions)
+	ProgramID      solana.PublicKey
+	StateID        solana.PublicKey
+	StoreProgramID solana.PublicKey
+
+	stateLock sync.RWMutex
```
connorwstein, line 32: idiomatic to use non-pointers for the locks then just use pointers to the object holding the lock. Then you can make use of its default unlocked value

jmank88 (dismissed):

pkg/solana/transmissions_cache.go
```diff
@@ -32,17 +26,9 @@ type ContractTracker struct {
 	// private key for the transmission signing
 	Transmitter TransmissionSigner
```
aalu1418: I think `Transmitter` and `TxManager` can be removed from the `TransmissionsCache` and `StateCache` - leftovers from the `ContractTracker` split

pkg/solana/transmissions_cache.go
```diff
@@ -18,11 +16,7 @@ import (
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
 )
 
-var (
-	configVersion uint8 = 1
-)
-
-type ContractTracker struct {
+type TransmissionsCache struct {
```
aalu1418, line 19: Probably out of scope, but would be interesting if the cache could be generic between the Transmissions and State cache?

aalu1418: further extension would be generic caching utility moved to `chainlink-relay`?

go.mod
```diff
@@ -10,12 +10,10 @@ require (
 	github.com/gagliardetto/treeout v0.1.4
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
-	github.com/satori/go.uuid v1.2.0
-	github.com/smartcontractkit/chainlink v1.3.0
+	github.com/smartcontractkit/chainlink v1.4.1-rc1.0.20220512210926-5363811660bf
```
aalu1418: Is this a stable commit on develop or a branch? (just concerned that a squash may leave this commit to be garbage collected by github sometime in the future)

connorwstein: will need to do a follow up, since this is an interface change its a 3 PR process

pkg/solana/transmissions_cache.go
connorwstein: you could probably share the polling code, but thats pretty minimal and they may want to evolve independently anyways

jmank88 (dismissed):

aalu1418 (dismissed):

jmank88 (approved):

PR #342, opened 2022-05-16
pkg/solana: remove core dependency

jmank88:
https://app.shortcut.com/chainlinklabs/story/39857/remove-chainlink-terra-solana-chainlink-core-dependency-to-resolve-cycle
- [x] https://github.com/smartcontractkit/chainlink-relay/pull/50
- [x] work around `github.com/smartcontractkit/chainlink/core/services/keystore/keys/solkey`?


pkg/solana/relay.go
```diff
@@ -22,9 +22,7 @@ type TransmissionSigner interface {
 }
 
 // TODO: Goes away with solana txm
-type KeyStore interface {
-	Get(id string) (solkey.Key, error)
-}
+type GetKey func(id string) (TransmissionSigner, error)
```
jmank88: This works around the `package solkey` import. The messy part is contained in core, fortunately.
https://github.com/smartcontractkit/chainlink/pull/6629/files#r874192985

go.mod
```diff
@@ -10,8 +10,7 @@ require (
 	github.com/gagliardetto/treeout v0.1.4
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
-	github.com/smartcontractkit/chainlink v1.4.2-0.20220516185434-65f3be3d96da
```
connorwstein, line 13: 🎉

connorwstein (dismissed):

pkg/solana/config/config_test.go
```diff
@@ -15,18 +15,26 @@ import (
...
+	testTTL                    = mustDuration(4 * time.Minute)
+	testTxTimeout              = mustDuration(5 * time.Minute)
+	testTxRetryTimeout         = mustDuration(6 * time.Minute)
+	testTxConfirmTimeout       = mustDuration(7 * time.Minute)
 	testPreflight              = false
 	testCommitment             = "finalized"
 	testMaxRetries       int64 = 123
 )
 
+func mustDuration(d time.Duration) utils.Duration {
```
krebernisak, line 30: Should we also move this fn to `chainlink-relay`?

pkg/solana/relay.go
krebernisak: Consider renaming the `TransmissionSigner` interface simply as `Signer` and this type as:

```suggestion
type SignerProvider func(id string) (Signer, error)
```

pkg/solana/relay.go
```diff
@@ -35,18 +33,18 @@ var _ relaytypes.Relayer = &Relayer{}
 type Relayer struct {
 	lggr     logger.Logger
 	chainSet ChainSet
-	ks       KeyStore
+	getKey   func(id string) (TransmissionSigner, error)
```
krebernisak:
```suggestion
	getSigner   SignerProvider
```

pkg/solana/config/config_test.go
jmank88: I was actually happy to remove it from exported API :laughing: . I wouldn't be opposed to a `package testutils` with a `MustDuration(*testing.TB,time.Duration)` that fails the test instead of panicing though... :thinking: 
Bigger picture, I am hoping that this odd custom wrapper type won't be necessary with the new TOML config, and this will dissapear.

pkg/solana/relay.go
jmank88: Much better!

aalu1418 (approved):
🎉 🎉 🎉

PR #343, opened 2022-05-16
Point to develop

connorwstein:
Follow up https://github.com/smartcontractkit/chainlink-solana/pull/338#discussion_r873859181


connorwstein:
Not needed now that we have removed the dependency on core

PR #345, opened 2022-05-18
remove tx signing + fixes for median/config refactor

aalu1418:
breaking change: `relayer` no longer requires the `getSigner` method

corresponding txm PR in core: https://github.com/smartcontractkit/chainlink/pull/6445

tested locally! everything is working again 😄

### DIFF
--- a/pkg/solana/relay.go
+++ b/pkg/solana/relay.go
@@ -31,22 +31,20 @@ type TxManager interface {
 var _ relaytypes.Relayer = &Relayer{}
 
 type Relayer struct {
-	lggr      logger.Logger
-	chainSet  ChainSet
-	getSigner SignerProvider
-	ctx       context.Context
-	cancel    func()
+	lggr     logger.Logger
+	chainSet ChainSet
+	ctx      context.Context
+	cancel   func()
 }
 
 // Note: constructed in core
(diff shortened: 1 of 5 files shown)


jmank88 (dismissed):

jmank88 (approved):

pkg/solana/relay.go
```diff
@@ -200,9 +201,30 @@ var _ relaytypes.MedianProvider = &medianProvider{}
...
-	contract    median.MedianContract
-	transmitter types.ContractTransmitter
+	transmissionsCache *TransmissionsCache
+	reportCodec        median.ReportCodec
+	contract           median.MedianContract
+	transmitter        types.ContractTransmitter
+}
+
+// start both cache services
+func (p *medianProvider) Start(ctx context.Context) error {
```
connorwstein, line 211: Ah this was the missing piece?

connorwstein (approved):

pkg/solana/relay.go
aalu1418: that was probably the biggest piece, but there was some slight mixing of `transmissionSigner` vs `transmissionsID` which are probably too closely named 😆 

but I am curious why the e2e tests in core didn't fail since it didn't work locally when I tried 🤔

PR #352, opened 2022-06-10
pkg/solana/config: add Chain and Node types for toml config

jmank88:
https://app.shortcut.com/chainlinklabs/story/33621

Supports: https://github.com/smartcontractkit/chainlink/pull/6656

Requires: 
- [x] https://github.com/smartcontractkit/chainlink-relay/pull/51
- [x] #353

### DIFF
--- a/pkg/solana/config/config.go
+++ b/pkg/solana/config/config.go
@@ -1,10 +1,13 @@
 package config
 
 import (
+	"net/url"
 	"sync"
 	"time"
 
 	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/db"
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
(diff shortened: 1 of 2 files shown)


pkg/solana/config/config.go
```diff
@@ -194,3 +197,71 @@ func (c *config) MaxRetries() *uint {
...
+	ConfirmPollPeriod   *utils.Duration
+	OCR2CachePollPeriod *utils.Duration
+	OCR2CacheTTL        *utils.Duration
+	TxTimeout           *utils.Duration
+	TxRetryTimeout      *utils.Duration
+	TxConfirmTimeout    *utils.Duration
+	SkipPreflight       *bool
+	Commitment          *string
+	MaxRetries          *int64
+}
```
aalu1418, line 212: hmmm what's the difference in usage between this struct and this one? wondering if 
 maybe these can be combined?
 https://github.com/smartcontractkit/chainlink-solana/blob/466d7d1795ac665c02cb382ae2a42c3951c7b40c/pkg/solana/db/db.go#L37

jmank88: I tried to reuse the old ones originally, but the db types weren't compatible with TOML.
Fortunately, the db types only need to live through the transition, and will be dropped later on.

aalu1418 (commented):
might need to update the go.mod file in `tests` for CI?

(also not sure why the anchor tests for the contracts take >30 min to run @archseer any ideas?)

aalu1418 (dismissed):
Approved! but the contracts CI looks broken...

(it's been running for 2+ hrs now)

vyzaldysanchez (approved):

PR #361, opened 2022-06-17
Add versioning and release with changesets

kylesmartin:


kylesmartin:
This PR will require import updates to `chainlink` before merging.

Any instance of `github.com/smartcontractkit/chainlink-solana/pkg` would need to be changed to `github.com/smartcontractkit/chainlink-solana/relayer/pkg`

PR #362, opened 2022-06-18
Juels per fee coin (lamport)

topliceanu:


pkg/monitoring/event/types.go
```diff
@@ -0,0 +1,80 @@
...
+	SetConfigDiscriminator = sumSetConfig[:8]
+
+	sumSetBilling := sha256.Sum256([]byte("event:SetBilling"))
+	SetBillingDiscriminator = sumSetBilling[:8]
+
+	sumRoundRequested := sha256.Sum256([]byte("event:RoundRequested"))
+	RoundRequestedDiscriminator = sumRoundRequested[:8]
+
+	sumNewTransmission := sha256.Sum256([]byte("event:NewTransmission"))
+	NewTransmissionDiscriminator = sumNewTransmission[:8]
```
jmank88: Maybe a helper func could dedupe and allow these to be set inline intsead?

topliceanu: Well, I quite like that it's explicit how these prefixes are being computed. The overhead should be minimal anyway. Maybe I'm not understanding the objection here?

jmank88: I found this difficult to read, and was struggling to see why it needs to exist inside a `func init` at all, rather than be set inline where the variable is declared.

I'm not at all concerned about overhead during initialization. That happens either way.

If this compiles, then IMHO it is already much easier to understand:
```go
	SetConfigDiscriminator       = sha256.Sum256([]byte("event:SetConfig"))[:8]
```
If this can't be inlined trivially, or is still difficult to read, then calling a simple helper func that accepts just the string would make it simultaneously readable and testable:
```go
	SetConfigDiscriminator       = someHelperFunc("event:SetConfig")
```

topliceanu: Aha, that makes sense. I added a helper and refactored

jmank88 (dismissed):

pkg/monitoring/source_envelope.go
```diff
@@ -121,6 +133,51 @@ func (s *envelopeSource) Fetch(ctx context.Context) (interface{}, error) {
...
+			continue
+		}
+		txRes, err := s.client.GetTransaction(
+			ctx,
+			txSig.Signature,
+			&rpc.GetTransactionOpts{
+				Commitment: rpc.CommitmentConfirmed,
+				Encoding:   solana.EncodingBase64,
+			},
+		)
```
aalu1418, line 163: would it be good to have a `nil` result error?

aalu1418: (and a test case please)

pkg/monitoring/event/decode.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Decode extracts an event from the the encoded event given as a string.
+func Decode(base64Encoded string) (interface{}, error) {
+	buf, err := base64.StdEncoding.DecodeString(base64Encoded)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode event '%s' from base64: %w", base64Encoded, err)
+	}
+	discriminator, encoded := buf[:8], buf[8:]
+	_ = discriminator
```
aalu1418: is this line needed?

pkg/monitoring/event/decode.go
```diff
@@ -0,0 +1,91 @@
...
+	return output
+}
+
+// Decode extracts an event from the the encoded event given as a string.
+func Decode(base64Encoded string) (interface{}, error) {
+	buf, err := base64.StdEncoding.DecodeString(base64Encoded)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode event '%s' from base64: %w", base64Encoded, err)
+	}
+	discriminator, encoded := buf[:8], buf[8:]
```
aalu1418: might be good to check length here and move the magic `8` number to a constant?

pkg/monitoring/event/decode.go
```diff
@@ -0,0 +1,91 @@
...
+		return event, nil
+	}
+	if bytes.Equal(discriminator, SetBillingDiscriminator) {
+		var event SetBilling
+		if err = event.UnmarshalBinary(encoded); err != nil {
+			return nil, fmt.Errorf("failed to decode event '%v' of type '%T': %w", encoded, event, err)
+		}
+		return event, nil
+	}
+	if bytes.Equal(discriminator, RoundRequestedDiscriminator) {
```
aalu1418: would this be cleaner in a switch case?

aalu1418 (changes requested):
just some thoughts to handle any edge/error cases :)

pkg/monitoring/source_envelope.go
topliceanu: I'm afraid I don't understand what you mean. Can you reformulate, please?

pkg/monitoring/event/decode.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+		} else if matches := programLogEvent.FindStringSubmatch(log); matches != nil {
+			currentProgramID := invocationStack[len(invocationStack)-1]
+			if programIDBase58 == currentProgramID {
+				output = append(output, matches[1])
+			}
+		}
+	}
+	return output
+}
```
topliceanu: @aalu1418 thoughts on this part?! The idea is that that logs come in as a trace of the execution stack, ie. this contract call that contract and that contract calls this other contract, etc. (see this tests https://github.com/smartcontractkit/chainlink-solana/pull/362/files/3b6baf66f3aaddfefc74c36c0cee1745f5bfedad#diff-e30bc1b5acc41eed18060e1a802e77e8c3f4525e006aaee5ec55263e4e391438R14-R23) What I'm doing here is, identifying the log lines that have events in them from the contract we care about and not the other contracts in the call stack.

pkg/monitoring/source_envelope.go
aalu1418: FindStringSubmatch`txRes` returns a pointer which could possibly be nil...and then further down it is used as `txRes.Meta.LogMessages` so just a little bit of defensive programming to make sure you don't hit a nil pointer error :)

pkg/monitoring/event/decode.go
aalu1418: i think it makes sense, maybe something for extra validation would be to test some logs that are messed up (missing information, incorrect format, etc) just to verify it behaves properly when there's unexpected logs

topliceanu: Thanks. Yeah, currently only testing the happy path. Will add some more tests.

pkg/monitoring/source_envelope.go
topliceanu: Ah ok, now it makes sense. Thanks! My expectation was that txRes will only be nil if err != nil. But, it's a fair point, I'll add a check. I don't think this warrants a test case, or were you referring to some other functionality that is not tested and should be?!

topliceanu: 👍

pkg/monitoring/event/decode.go
topliceanu: Nope 🙃

topliceanu: nope 🙃

pkg/monitoring/event/decode.go
topliceanu: excellent idea!

pkg/monitoring/event/decode.go
topliceanu: yeah, and fewer lines of code!

aalu1418 (commented):
looking good! although I'm curious why the CI is failing to build the CL image.... @jmank88 or @skudasov if you have a sec, could you take a look?

it looks like some sort of dependency issue 🤔

jmank88:
I'm not sure...
Maybe a rebase would help? Even if there are no conflicts :shrug:

aalu1418:
> Maybe a rebase would help? Even if there are no conflicts shrug

I would second that, didn't realize this wasn't pointed to the latest head

aalu1418 (approved):

PR #376, opened 2022-07-14
Update workflows to use the updated github actions to run tests

tateexon:


.github/actions/build_contract_artifacts/action.yml
```diff
@@ -0,0 +1,47 @@
+name: Build Contract Artifacts
+description: Common builder for test contracts for automated tests
+
+runs:
+  using: composite
+  steps:
+    - name: Install latest Git version (>= 2.18.0) for actions/checkout
```
chainchad, line 7: Any way to avoid this step? Having to update apt with these packages will definitely add time to the builds.

Does this https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md have the git version you need?

chainchad (approved):
Nice 👍 

I can attest to the changes around changing inputs for AWS.

skudasov (approved):

.github/actions/build_contract_artifacts/action.yml
tateexon: I will look into updating this. I was just moving what the solana team had already put in place for the build to a common spot so I didn't copy paste the same logic into 4 places. Also this job finishes faster than the docker build job that runs in parallel at the moment so it is not yet the bottleneck.

PR #377, opened 2022-07-15
Fixes For Running Solana Smoke Tests As Part of Chainlink Repo

tateexon:
Due to us using the github "environment" stuff we can no longer use reusable workflows in the same way. This means we need to use actions and instead of using a reusable workflow. 

So the chainlink repo will now run both a build for the contracts and a test run using the run-tests action from chainlink-github-actions. Here is a test run using this strategy as well as the changes in this PR: https://github.com/smartcontractkit/chainlink/actions/runs/2679696938

Future improvement would be to have the tests in chainlink use a stable cached version of the solana artifacts so it would not build them every time.


.github/workflows/contract-release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Release Artifacts
     runs-on: ubuntu-latest
     container:
-      image: projectserum/build:v0.22.1
+      image: projectserum/build:v0.24.2
```
aalu1418, line 16: just wondering, why are we bumping the projectserum/build image?

skudasov (approved):

.github/workflows/contract-release.yml
tateexon: It is already bumped in the makefile so I was just bringing it to parity with what people are developing on. If we shouldn't bump it or I should move the makefile back down to 22.1 I can do that.

aalu1418: 👍🏼 ah no worries then

PR #380, opened 2022-07-27
Fixes

archseer:
Best reviewed commit-by-commit, I'm still cleaning up the branch.


tests/e2e/solclient/ocr2.go
```diff
@@ -353,10 +353,10 @@ func (m *OCRv2) proposeConfig(ocConfig contracts.OffChainAggregatorV2Config) err
 	}
 	instr = append(instr, ocr_2.NewProposePayeesInstruction(
 		m.Mint.PublicKey(),
-		payees,
 		m.Proposal.PublicKey(),
 		m.Owner.PublicKey()).Build(),
 	)
+	// TODO: how to set payees as remaining
```
archseer: @skudasov Any idea here? I need to set `remainingAccounts`

contracts/programs/ocr2/src/lib.rs
```diff
@@ -278,25 +278,23 @@ pub mod ocr2 {
     }
 
     #[access_control(proposal_owner(&ctx.accounts.proposal, &ctx.accounts.authority))]
-    pub fn propose_payees(
-        ctx: Context<ProposeConfig>,
-        token_mint: Pubkey,
-        payees: Vec<Pubkey>,
-    ) -> Result<()> {
+    pub fn propose_payees(ctx: Context<ProposeConfig>, token_mint: Pubkey) -> Result<()> {
```
krebernisak, line 295: How come we accept the `token_mint` as an input param when it's already available with `state.config.token_mint`, but then we don't accept token changes as required [here](https://github.com/smartcontractkit/chainlink-solana/blob/51691de3f05a5e67750a345292fe6ff7913a0321/contracts/programs/ocr2/src/lib.rs#L172)?

Looks like the LINK token is not settable/upgradeable, right?

contracts/programs/ocr2/src/lib.rs
```diff
@@ -278,25 +278,23 @@ pub mod ocr2 {
...
-        for account in ctx.remaining_accounts {
+        for account in payees {
             let account = Account::<'_, token::TokenAccount>::try_from(account)?;
             require!(account.mint == token_mint, InvalidTokenAccount);
         }
 
-        for (oracle, payee) in proposal.oracles.iter_mut().zip(payees.into_iter()) {
-            oracle.payee = payee;
+        for (oracle, payee) in proposal.oracles.iter_mut().zip(payees.iter()) {
+            oracle.payee = payee.key();
```
krebernisak, line 311: [How to audit Solana smart contracts Part 4: the Anchor framework](https://medium.com/coinmonks/how-to-audit-solana-smart-contracts-part-4-the-anchor-framework-ef42d944f086):
> Note: while accounts are deserialized and validated according to the macros, remaining_accounts are not, so its use must be very careful.

Is this something to worry about?

contracts/examples/hello-world/programs/hello-world/Cargo.toml
```diff
@@ -16,5 +16,5 @@ cpi = ["no-entrypoint"]
 default = []
 
 [dependencies]
-anchor-lang = "0.23.0"
-chainlink_solana = "0.1.2"
+anchor-lang = "0.25.0"
+chainlink_solana = "0.1.1"
```
krebernisak: Why the step back here?

contracts/tests/ocr2.spec.ts
```diff
@@ -582,7 +582,7 @@ describe("ocr2", async () => {
     // fetch payees
     account = await program.account.state.fetch(state.publicKey);
     let currentOracles = account.oracles.xs.slice(0, account.oracles.len);
-    let payees = currentOracles.map((oracle) => {
+    payees = currentOracles.map((oracle) => {
```
krebernisak, line 590: How did this compile before? 🤔

contracts/programs/ocr2/src/lib.rs
archseer: Yeah it's not because we'd have to specify up to 19 old accounts to pay out the oracles + 19 new token accounts belonging to the new token: over the limit

contracts/programs/ocr2/src/lib.rs
archseer: The previous loop verifies that they are valid token accounts that belong to the correct mint.

contracts/tests/ocr2.spec.ts
archseer: We now need to fetch payees on line 519 to pass them into proposePayees, before that it was passing them in directly by doing a `map` without using a variable

contracts/examples/hello-world/programs/hello-world/Cargo.toml
archseer: I think this should be 1.0.0 actually, `0.1.2` was pulled so it no longer exists: https://crates.io/crates/chainlink_solana/versions

tests/e2e/solclient/ocr2.go
```diff
@@ -343,20 +345,25 @@ func (m *OCRv2) proposeConfig(ocConfig contracts.OffChainAggregatorV2Config) err
...
 		m.Proposal.PublicKey(),
-		m.Owner.PublicKey()).Build(),
-	)
+		m.Owner.PublicKey())
+	// Add payees as remaining accounts
+	for i := 0; i < len(payees); i++ {
+		proposeInstr.Append(solana.Meta(payees[i]))
+	}
+	instr = append(instr, proposeInstr.Build())
+	// TODO: how to set payees as remaining
```
krebernisak: This is done, right?

```suggestion
```

tests/e2e/solclient/ocr2.go
```diff
@@ -343,20 +345,25 @@ func (m *OCRv2) proposeConfig(ocConfig contracts.OffChainAggregatorV2Config) err
 	}
 	// set one payee for all
 	instr := make([]solana.Instruction, 0)
+	// TODO: get associated addr
```
krebernisak, line 348: ?

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/token/transfer.ts
```diff
@@ -1,7 +1,12 @@
 import { Result } from '@chainlink/gauntlet-core'
 import { logger, BN, prompt } from '@chainlink/gauntlet-core/dist/utils'
 import { SolanaCommand, TransactionResponse } from '@chainlink/gauntlet-solana'
-import { ASSOCIATED_TOKEN_PROGRAM_ID, Token, TOKEN_PROGRAM_ID } from '@solana/spl-token'
+import {
+  ASSOCIATED_TOKEN_PROGRAM_ID,
```
krebernisak: `ASSOCIATED_TOKEN_PROGRAM_ID` and `TOKEN_PROGRAM_ID` don't seem used anymore.

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/initialize.ts
```diff
@@ -1,7 +1,7 @@
 import { Result } from '@chainlink/gauntlet-core'
 import { SolanaCommand, TransactionResponse } from '@chainlink/gauntlet-solana'
 import { Keypair, PublicKey, TransactionInstruction, SystemProgram, SYSVAR_RENT_PUBKEY } from '@solana/web3.js'
-import { ASSOCIATED_TOKEN_PROGRAM_ID, Token, TOKEN_PROGRAM_ID } from '@solana/spl-token'
+import { ASSOCIATED_TOKEN_PROGRAM_ID, getOrCreateAssociatedTokenAccount, TOKEN_PROGRAM_ID } from '@solana/spl-token'
```
krebernisak: `ASSOCIATED_TOKEN_PROGRAM_ID` and `TOKEN_PROGRAM_ID` don't seem used anymore.

contracts/programs/ocr2/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub mod ocr2 {
     ) -> Result<()> {
         let mut proposal = ctx.accounts.proposal.load_init()?;
 
-        proposal.version = 1;
+        proposal.version = STATE_VERSION;
```
krebernisak, line 112: Could these two versions diverge?
If yes, let's use a specific proposal version constant.

contracts/programs/ocr2/src/lib.rs
```diff
@@ -61,6 +61,25 @@ pub mod ocr2 {
             ctx.remaining_accounts,
         )?;
 
+        // Transfer out remaining balance from the token_vault
+        let balance_gjuels = token::accessor::amount(&ctx.accounts.token_vault.to_account_info())?;
```
krebernisak, line 66: We should probably check if this is zero, and only move forward with the transfer if it's not.

contracts/tests/ocr2.spec.ts
```diff
@@ -958,28 +1042,40 @@ describe("ocr2", async () => {
...
+    // TODO: it complains account is off curve
+    let recipientTokenAccount = await getOrCreateAssociatedTokenAccount(
+      provider.connection,
+      fromWallet,
+      token,
+      provider.wallet.publicKey,
+      true
+    );
+
+    let beforeBalance = (
```
krebernisak, line 1069: This variable doesn't seem used.

We should also assert that the balance is actually transferred on close.

contracts/programs/ocr2/src/lib.rs
```diff
@@ -563,11 +568,18 @@ fn pay_oracles_impl<'info>(
             oracle.payment_gjuels = 0;
             oracle.from_round_id = latest_round_id;
 
+            let to = if payee.owner == &token::ID {
+                payee.to_account_info()
+            } else {
+                // then the token account must be closed
+                token_receiver.to_account_info()
```
krebernisak, line 578: Shouldn't this happen only on close?

In other situations it could be more efficient to skip the oracle and keep the funds in the contract (on `pay_oracles`, `accept_proposal` & `set_billing`).

The way it's currently designed this will require setting this additional `token_receiver` account when calling these non-close methods, which might be confusing for contract operators.

contracts/examples/hello-world/tests/hello-world.ts
```diff
@@ -1,25 +1,25 @@
...
+import * as anchor from "@project-serum/anchor";
+import * as fs from "fs";
+import { Program, BN } from "@project-serum/anchor";
+import { HelloWorld } from "../target/types/hello_world";
 
 const CHAINLINK_PROGRAM_ID = "HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny";
 
-describe('hello-world', () => {
-  const provider = anchor.Provider.env();
+describe("hello-world", () => {
```
krebernisak, line 8: I've noticed many style changes in this PR but looks like we don't have prettier configured in this repository. I suggest we use the one from StarkNet integration and we reformat the codebase one more time: https://github.com/smartcontractkit/chainlink-starknet/blob/develop/.prettierrc.json

contracts/examples/hello-world/tests/hello-world.ts
```diff
@@ -32,54 +32,53 @@ describe('hello-world', () => {
...
-          owner: owner.publicKey,
           authority: owner.publicKey,
         },
-      });
-    
+        signers: [feed],
+        preInstructions: [
+          await storeProgram.account.transmissions.createInstruction(
+            feed,
+            8 + 192 + 6 * 24
```
krebernisak: Let's define these as descriptive constants in a single file and import them here

contracts/programs/ocr2/src/context.rs
```diff
@@ -15,14 +14,12 @@ pub struct Initialize<'info> {
     #[account(zero)]
     pub state: AccountLoader<'info, State>,
     pub feed: Account<'info, Transmissions>,
-    #[account(mut)]
-    pub payer: Signer<'info>,
+    // #[account(mut)]
+    // pub payer: Signer<'info>,
```
krebernisak: If these are not used anymore we should remove them

contracts/programs/ocr2/src/context.rs
```diff
@@ -33,21 +30,24 @@ pub struct Initialize<'info> {
...
-    pub rent: Sysvar<'info, Rent>,
-
-    pub system_program: Program<'info, System>,
-    pub token_program: Program<'info, Token>,
-    pub associated_token_program: Program<'info, AssociatedToken>,
+    // #[account(address = sysvar::rent::ID)]
+    // pub rent: Sysvar<'info, Rent>,
+    // pub system_program: Program<'info, System>,
+    // pub token_program: Program<'info, Token>,
+    // pub associated_token_program: Program<'info, AssociatedToken>,
```
krebernisak: If these are not used anymore we should remove them

contracts/programs/ocr2/src/lib.rs
```diff
@@ -159,11 +168,15 @@ pub mod ocr2 {
             ctx.accounts.token_vault.to_account_info(),
             ctx.accounts.vault_authority.clone(),
```
krebernisak, line 172: Why `clone()` vs `to_account_info()`?

Asking because I've noticed we use `ctx.accounts.vault_authority.to_account_info()` [here#L73](https://github.com/smartcontractkit/chainlink-solana/blob/7108b0b3954e0be023bac74bceb3a7b749f90b8a/contracts/programs/ocr2/src/lib.rs#L73)

contracts/tests/ocr2.spec.ts
```diff
@@ -592,12 +594,55 @@ describe("ocr2", async () => {
...
+        owner: owner.publicKey,
+        tokenMint: token,
+        tokenVault: tokenVault,
+        vaultAuthority: vaultAuthority,
+        requesterAccessController: requesterAccessController.publicKey,
+        billingAccessController: billingAccessController.publicKey,
+        // rent: SYSVAR_RENT_PUBKEY,
+        // systemProgram: SystemProgram.programId,
+        // tokenProgram: TOKEN_PROGRAM_ID,
+        // associatedTokenProgram: ASSOCIATED_TOKEN_PROGRAM_ID,
```
krebernisak: If these are not used anymore we should remove them

contracts/tests/ocr2.spec.ts
```diff
@@ -592,12 +594,55 @@ describe("ocr2", async () => {
...
     });
+  };
+
+  it("Initializes the OCR2 config", async () => {
+    await program.methods
+      .initialize(new BN(minAnswer), new BN(maxAnswer))
+      .accounts({
+        state: state.publicKey,
+        feed: feed.publicKey,
+        // payer: provider.wallet.publicKey,
```
krebernisak: If these are not used anymore we should remove them

contracts/tests/ocr2.spec.ts
```diff
@@ -592,12 +594,55 @@ describe("ocr2", async () => {
...
+        // rent: SYSVAR_RENT_PUBKEY,
+        // systemProgram: SystemProgram.programId,
+        // tokenProgram: TOKEN_PROGRAM_ID,
+        // associatedTokenProgram: ASSOCIATED_TOKEN_PROGRAM_ID,
+      })
+      .signers([state])
+      .preInstructions([
+        await program.account.state.createInstruction(state),
+        // await store.account.transmissions.createInstruction(transmissions, 8+192+8096*24),
+        // createFeed,
```
krebernisak, line 627: ?

contracts/examples/hello-world/programs/hello-world/Cargo.toml
```diff
@@ -16,5 +16,5 @@ cpi = ["no-entrypoint"]
 default = []
 
 [dependencies]
-anchor-lang = "0.23.0"
-chainlink_solana = "0.1.2"
+anchor-lang = "0.25.0"
+chainlink_solana = "1.0.0"
```
krebernisak, line 20: We didn't change versions for this update, right?

- So deployed contracts are 1.0.0, and this updated contracts will also be 1.0.0? 
- Will this mean we can fetch versions for old and new contracts and both will be the same?
- Why don't we bump this to 1.1.0?

contracts/programs/ocr2/src/lib.rs
archseer: You can't skip the oracle because then the oracles aren't fully paid out so `accept_proposal`/`set_billing` won't succeed. Are you proposing to just ignore the oracle and then have the operator manually withdraw funds to resolve? That could work, but it would be easy to miss if one of the nodes didn't get paid out. 

I think I set the `token_receiver` to whoever executes the call currently so there's no extra parameter.

contracts/programs/ocr2/src/lib.rs
archseer: They're basically the same:

```
    fn to_account_info(&self) -> AccountInfo<'info> {
        self.as_ref().clone()
    }

```

I'll switch to `to_account_info` for consistency

contracts/examples/hello-world/programs/hello-world/Cargo.toml
archseer: I'll update this to use `1.0`, we're not bumping the library to `1.1` because there wasn't any changes there though. The user facing queries are all stable and will work with both versions

contracts/examples/hello-world/tests/hello-world.ts
archseer: Let's do that in a follow-up PR since the diff is already quite big

contracts/programs/ocr2/src/lib.rs
archseer: Added, but I think it's unlikely to ever be exactly 0

contracts/programs/ocr2/src/lib.rs
archseer: I think they should be in sync, an older proposal shouldn't be applied to a newer state because the layout expectations might be different

contracts/tests/ocr2.spec.ts
archseer: I hoped to create these in a single transaction but had some issues getting it to work.

contracts/tests/ocr2.spec.ts
archseer: It is used a couple of lines down: `assert.ok(afterBalance > beforeBalance);`

contracts/programs/ocr2/src/lib.rs
krebernisak: I was suggesting that the owner probably doesn't expect to collect some dust every time they `set/accept_proposal` or `set_billing`, so it might go unnoticed. 

I think I'd prefer these operations to skip the invalid payee, keep the funds in the contract and then all the remaining funds are transferred out either on close (rare occasion) or potentially an owner can withdraw on behalf of an invalid oracle (new fn).

Feels cleaner to do it that way, but maybe it's just my preference. Wdyt?

contracts/examples/hello-world/tests/hello-world.ts
krebernisak: https://github.com/smartcontractkit/chainlink-solana/issues/394

contracts/programs/ocr2/src/lib.rs
krebernisak: https://github.com/smartcontractkit/chainlink-solana/issues/395

krebernisak (approved):
Huge effort and nice improvements. Great work, @archseer!

I've reviewed to the best of my ability and it looks solid.

Except the CI e2e test failing, there is this [discussion on oracle payments](https://github.com/smartcontractkit/chainlink-solana/pull/380/files#r970782232) we might still consider, otherwise looks good.

PR #389, opened 2022-08-25
Withdraw payment + docs

RodrigoAD:
Gauntlet Command to withdraw payments for Node Ops and Documentation

### DIFF
--- a/docs/gauntlet/withdraw.md
+++ b/docs/gauntlet/withdraw.md
@@ -1,0 +1,17 @@
+# Withdraw Payments using Gauntlet
+
+When Node Operators report data to a feed, they are compensated with LINK thats owned by the feed. Node operators can then withdraw the funds owed to them via gauntlet using this section below
+
+## Sample Command Usage
+
+Make sure to set up gauntlet using the steps described in [Gauntlet Setup](gauntlet-setup.md) before attempting to run the following command
+
+```bash
+yarn gauntlet ocr2:withdraw_payment --network=mainnet --recipient=<VALID_LINK_RECIPIENT_ADDRESS> <FEED_ADDRESS>
+```
+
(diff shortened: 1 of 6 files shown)


aalu1418 (commented):
Ideally I would like to see a test case with this not sure how easy that is?

Or atleast confirmation that this has been tested and verified that it works properly

RodrigoAD:
Ideally we would have Gauntlet tests, but for now it can only be manually tested. I did it myself and works as expected. The tx signer needs to be the owner of the payee address (token derived account of the signer)

docs/gauntlet/withdraw.md
```diff
@@ -0,0 +1,17 @@
+# Withdraw Payments using Gauntlet
+
+When Node Operators report data to a feed, they are compensated with LINK thats owned by the feed. Node operators can then withdraw the funds owed to them via gauntlet using this section below
+
+## Sample Command Usage
+
+Make sure to set up gauntlet using the steps described in [Gauntlet Setup](gauntlet-setup.md) before attempting to run the following command
+
+```bash
+yarn gauntlet ocr2:withdraw_payment --network=mainnet --recipient=<VALID_LINK_RECIPIENT_ADDRESS> <FEED_ADDRESS>
```
aalu1418: What does `VALID_LINK_RECIPIENT_ADDRESS` mean? I think it means that it is the associated token account...is that correct? 

Would be good to also add something like `VALID_LINK_RECIPIENT_ADDRESS` = the same account address used in setting the payee config (should be the associated account for the LINK token specific to the owner public key)

aalu1418 (changes requested):

docs/gauntlet/withdraw.md
RodrigoAD: Updated it @aalu1418, lmk if it's more clear now

aalu1418 (approved):
lgtm!

PR #405, opened 2022-10-06
Include SetComputeUnitPrice in reimbursement calculation

archseer:


contracts/programs/ocr2/src/lib.rs
```diff
@@ -800,18 +803,73 @@ impl Report {
...
 }
 
-fn calculate_reimbursement_gjuels(juels_per_lamport: u64, _signature_count: usize) -> Result<u64> {
+fn calculate_reimbursement_gjuels(
+    juels_per_lamport: u64,
+    signature_count: usize,
+    instructions_sysvar: Option<&AccountInfo<'_>>,
+) -> Result<u64> {
     const SIGNERS: u64 = 1;
+    const MICRO: u64 = 10u64.pow(3);
```
aalu1418:
```suggestion
    const MICRO: u64 = 10u64.pow(6);
```

contracts/tests/ocr2.spec.ts
```diff
@@ -223,6 +229,7 @@ describe("ocr2", async () => {
             isSigner: false,
           },
           { pubkey: storeAuthority, isWritable: false, isSigner: false },
+          { pubkey: SYSVAR_INSTRUCTIONS_PUBKEY, isWritable: false, isSigner: false },
```
aalu1418, line 232: nit: is there a test case where `SYSVAR_INSTRUCTIONS_PUBKEY` isn't passed to the contract?

thinking about the case where the contract is upgraded before the node is upgraded (or vice versa)

contracts/tests/ocr2.spec.ts
```diff
@@ -981,22 +990,81 @@ describe("ocr2", async () => {
...
+    console.log(t.meta.logMessages); // 195_785/195_997
+    account = await program.account.state.fetch(state.publicKey, 'confirmed');
+    currentOracles = account.oracles.xs.slice(0, account.oracles.len);
+    // get the transmitter
+    oracle = currentOracles[i];
+    let after = oracle.paymentGjuels;
+    
+    let payment = after.sub(before).toNumber();
+    let execUnits = (25_000 * (f+1) + 21_000);
+    let computeUnitCostLamports = execUnits * computePrice / Math.pow(10, 3);
```
aalu1418:
```suggestion
    let computeUnitCostLamports = execUnits * computePrice / Math.pow(10, 6);
```

aalu1418 (changes requested):
looking good! just one fix that has to be addressed, the other comment would be nice but I would at least like confirmation that this doesn't break things if the program is upgraded before the nodes

archseer:
> but I would at least like confirmation that this doesn't break things if the program is upgraded before the nodes

No tests since I'd like to remove this codepath soon but I reverted https://github.com/smartcontractkit/chainlink-solana/pull/405/commits/c81b45835739bb157e401c1d5f4f5902c2f215ba and ran the tests to ensure they still pass

aalu1418 (approved):
LGTM!

this can be merged as long as we are sure that the program change is compatible in both ways:
old CL node <=> new programs
new CL node <=> old programs

archseer:
CL nodes shouldn't upgrade until the program is first upgraded.

PR #406, opened 2022-10-07
Multisig batching

archseer:
(DO NOT SQUASH)

- `serum_multisig:approve` allows approving multiple transactions at once
- `proposeConfig` is now a single command, replacing `createProposal`, `proposeConfig`, `proposeOffchainConfig`, `proposePayees` and `finalizeProposal`
  - to support this, `sendRawTx` now uses more reliable send & confirm code that can retry
  - if a proposal fails halfway through:
    - it can be resumed by passing a `--proposal <PROPOSAL_ID>` 
    - if aborting, it should be manually closed to reclaim rent
- `localnet.sh` sets up a local node with programs pre-deployed and a pre-funded account
- `token:create_account` was changed to allow creating multiple token accounts at once (single tx)
- `setup.dev.flow` now actually works again (pending https://github.com/smartcontractkit/gauntlet-core/pull/100)
- `typescript`, `nodejs` and `golang` were upgraded


gauntlet/packages/gauntlet-serum-multisig/src/commands/approve.ts
```diff
@@ -0,0 +1,44 @@
+import { SolanaCommand, TransactionResponse } from '@chainlink/gauntlet-solana'
+import { PublicKey } from '@solana/web3.js'
+import { Result } from '@chainlink/gauntlet-core'
+import { logger } from '@chainlink/gauntlet-core/dist/utils'
+import { CONTRACT_LIST, getContract } from '../lib/contracts'
+
+export default class Approve extends SolanaCommand {
+  static id = 'serum_multisig:approve'
```
RodrigoAD, line 8: When is this going to be used?

gauntlet/packages/gauntlet-serum-multisig/src/commands/approve.ts
```diff
@@ -0,0 +1,44 @@
...
+          transaction: new PublicKey(tx),
+          owner: signer,
+        })
+        .instruction()
+    ))
+
+    return ixs
+  }
+
+  //execute not needed, this command cannot be ran outside of multisig
```
RodrigoAD, line 41: This should be run by the operator, not the multisig

archseer:
Not sure why the CI is breaking. Maybe the golang bump?

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/proposeConfig.ts
```diff
@@ -82,15 +258,81 @@ export default class ProposeConfig extends SolanaCommand {
...
       .accounts({
-        proposal,
+        proposal: proposal.publicKey,
+        authority: signer,
+      })
+      .instruction()
+
+    // proposePayees
+
+    // TODO: validate that all payees exist as token accounts on chain
```
aalu1418, line 293: open TODO?

aalu1418 (commented):
Just thinking: combining the create, proposeConfig, proposeOffchainConfig, proposePayees, accept, finalize is nice when transactions are working smoothly on chain. but when individual TXs start to fail that would interrupt the proposal flow and make it more difficult to restart.

I wonder if a flow command would be better..which allows users to restart the process in the middle in case something fails. Wdyt?

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -86,9 +161,15 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
       price?: number
     } = {},
   ): Promise<TransactionSignature> => {
-    const { blockhash, lastValidBlockHeight } = await this.provider.connection.getLatestBlockhash()
     if (overrides.units) logger.info(`Sending transaction with custom unit limit: ${overrides.units}`)
     if (overrides.price) logger.info(`Sending transaction with custom unit price: ${overrides.price}`)
+
+    // let confirmLevel: TransactionConfirmationStatus = 'confirmed'
+    let confirmLevel: TransactionConfirmationStatus = 'finalized'
```
aalu1418: i would recommend using `confirmed` rather than `finalized` here

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -99,11 +180,188 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+        })
+      }
+      throw new SolanaError({ message: 'Transaction failed', txid })
+    } finally {
+      done = true
+    }
+    return txid
+
+    // TODO: skipPreflight: true since we previously simulated it?
+    // return await sendAndConfirmRawTransaction(this.provider.connection, rawTransaction) // TODO: { maxRetries: 5 }
```
aalu1418: open TODOs?

and yeah if it's already simulated, no need to do preflight checks again...but it doesn't also hurt 🤷🏼 this code doesn't have to be the most optimized

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -130,14 +388,17 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
         blockhash,
         lastValidBlockHeight,
       })
+      // TODO: makeTx is missing extra signatures
```
aalu1418: another TODO

aalu1418 (commented):
also wondering...are there tests for these changes?

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/proposeConfig.ts
archseer: I usually leave `TODO`s in the code as hints of future improvements to whoever looks at this code next (e.g. https://github.com/torvalds/linux/search?l=C&p=99&q=TODO), the goal is not to get to TODO count 0

archseer:
> also wondering...are there tests for these changes?

There are no gauntlet tests at all in this repo, but I'm working on this in https://github.com/smartcontractkit/chainlink-solana/pull/413

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
archseer: I'm still debugging this, but I think `simulateTransaction` available in web3.js defaults to using `finalized`, so if sending a sequence of transactions with `confirmed`, the following simulate might run too soon and not see previous changes. 

I have both a workaround as well as work towards an update: `simulateTransaction` has a new version with a confirmation level, but it requires the new transaction version (instead of legacy). I'll try to get those changes in

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
archseer: I think with the extra signatures this would then be equivalent to all the preflight checks

archseer:
> but when individual TXs start to fail that would interrupt the proposal flow and make it more difficult to restart.

I hope that the improved transmission logic will make failures less common. 

> I wonder if a flow command would be better..which allows users to restart the process in the middle in case something fails. Wdyt?

I'll work on a `--proposal` flag that can automatically lookup the proposal state and smartly resume from the correct step without user input. I like the merged command because it does all validation ahead of time before attempting to send transactions (I also hoped to do all the simulations ahead of time but most of these calls depend on previous state)

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -86,24 +139,204 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+      }
+    })()
+
+    try {
+      await this.awaitTransactionSignatureConfirmation(txid, timeout, CONFIRM_LEVEL, currentBlockhash)
+    } catch (err: any) {
+      if (err.timeout) {
+        throw new TimeoutError({ txid })
+      }
+      let simulateResult: SimulatedTransactionResponse | null = null
```
RodrigoAD: - Don't understand why we simulate after we send the tx, shouldn't we do it before?
- We already have a simulateTx method, we should update it with this more detailed implementation and use it here (if necessary)

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -86,24 +139,204 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+        this.provider.connection.sendRawTransaction(rawTransaction, {
+          skipPreflight: true,
+        })
+        if (retryAttempts <= maxRetries) {
+          retryAttempts = retryAttempts++
+        } else {
+          break
+        }
+      }
+    })()
```
RodrigoAD: this could be perfectly abstracted into a `sendWithRetries` method:
```suggestion
    let done = false
    const isDone = () => done
    const timeout = 60_000
    const txid = await this.sendWithRetries(isDone, rawTransaction, timeout)
```

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -86,24 +139,204 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+    } finally {
+      done = true
+    }
+    return txid
+
+    // TODO: skipPreflight: true since we previously simulated it?
+    // return await sendAndConfirmRawTransaction(this.provider.connection, rawTransaction) // TODO: { maxRetries: 5 }
+  }
+
+  async awaitTransactionSignatureConfirmation(
```
RodrigoAD: This whole fn is almost impossible to understand. I know it comes from Mango, but some refactor or at least comments I feel is necessary. The next one who could come here to update something will appreciate it

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -86,24 +139,204 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
       price?: number
     } = {},
   ): Promise<TransactionSignature> => {
-    const { blockhash, lastValidBlockHeight } = await this.provider.connection.getLatestBlockhash()
     if (overrides.units) logger.info(`Sending transaction with custom unit limit: ${overrides.units}`)
     if (overrides.price) logger.info(`Sending transaction with custom unit price: ${overrides.price}`)
+
+    this.simulateTx(this.wallet.publicKey, rawTxs)
```
RodrigoAD:
```suggestion
    await this.simulateTx(this.wallet.publicKey, rawTxs)
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/token/createAccount.ts
```diff
@@ -9,39 +9,42 @@ import { isValidTokenAccount } from './utils'
...
-    await prompt(`Continue to create new Token associated account to ${newAccountBase.toString()}`)
-    logger.loading('Creating account...')
-    const tx = await this.signAndSendRawTx([ix])
+    await prompt(`Continue to create new Token associated accounts?`)
+    logger.loading('Creating accounts...')
+    const tx = await this.signAndSendRawTx(ixs)
 
-    logger.success(`New account created at ${associatedAcc.toString()} on tx ${tx}`)
+    logger.success(`New accounts created on tx ${tx}`)
+    // logger.success(`New account created at ${associatedAcc.toString()} on tx ${tx}`)
```
RodrigoAD, line 47:
```suggestion
```

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -124,20 +357,24 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+      const { value: simulationResponse } = await this.provider.connection.simulateTransaction(tx, {
+        commitment: CONFIRM_LEVEL,
+      })
       if (simulationResponse.err) {
         throw new Error(JSON.stringify({ error: simulationResponse.err, logs: simulationResponse.logs }))
       }
       logger.success(`Tx simulation succeeded: ${simulationResponse.unitsConsumed} units consumed.`)
       return simulationResponse.unitsConsumed
     } catch (e) {
+      console.log(e.message)
```
RodrigoAD, line 241: This should be handled upstream, after we throw

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
archseer: This is just a direct port, I'll delete most of this code once https://github.com/solana-labs/solana/pull/28290 lands upstream

archseer: The reason they simulate afterwards on error is: if you skip the preflight checks and didn't simulate the tx beforehand you can see how exactly the tx failed (full error output)

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
archseer: I had to log here since some of the errors returned weren't parsable JSON and the error handling code would crash

archseer:
Note: still doing some debugging, at least in the `setup.dev.flow` the digest from the input vs calculated from the state in `approveProposal` don't match

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
RodrigoAD: > This is just a direct port, I'll delete most of this code once [solana-labs/solana#28290](https://github.com/solana-labs/solana/pull/28290) lands upstream

Please, then add a big comment explaining this

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setup.dev.flow.ts
```diff
@@ -213,78 +201,45 @@ export default class SetupFlow extends FlowCommand<TransactionResponse> {
...
-        },
-        args: [FlowCommand.ID.contract(this.stepIds.OCR_2)],
-      },
-      {
-        name: 'Propose Offchain Config',
-        command: ProposeOffchainConfig,
-        flags: {
-          input: {
+            oracles: this.configInput.oracles,
+            f: this.configInput.f,
```
RodrigoAD, line 220:
```suggestion
            f: this.configInput.f,
            userSecret: randomSecret
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setup.dev.flow.ts
```diff
@@ -213,78 +201,45 @@ export default class SetupFlow extends FlowCommand<TransactionResponse> {
...
             offchainConfig: offchainConfigInput,
-            proposalId: this.getReportStepDataById(FlowCommand.ID.data(this.stepIds.PROPOSAL, 'proposal')),
           },
-          proposalId: FlowCommand.ID.data(this.stepIds.PROPOSAL, 'proposal'),
           secret: randomSecret,
```
RodrigoAD:
```suggestion
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setup.dev.flow.ts
```diff
@@ -213,78 +201,45 @@ export default class SetupFlow extends FlowCommand<TransactionResponse> {
...
-      },
       {
         name: 'Accept Proposal',
         command: AcceptProposal,
         flags: {
+          link: FlowCommand.ID.contract(this.stepIds.TOKEN),
           input: {
-            secret: randomSecret,
+            proposalId: FlowCommand.ID.data(this.stepIds.PROPOSAL, 'proposal'),
+            secret: FlowCommand.ID.data(this.stepIds.PROPOSAL, 'secret'),
```
RodrigoAD:
```suggestion
            secret: randomSecret,
```

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ocr2/setup.dev.flow.ts
```diff
@@ -213,78 +201,45 @@ export default class SetupFlow extends FlowCommand<TransactionResponse> {
...
-            tokenMint: this.getReportStepDataById(FlowCommand.ID.contract(this.stepIds.TOKEN)),
-            oracles: acceptPropOracles,
+            f: this.configInput.f,
+            tokenMint: FlowCommand.ID.contract(this.stepIds.TOKEN),
+            oracles: this.configInput.oracles,
             offchainConfig: offchainConfigInput,
           },
           proposalId: FlowCommand.ID.data(this.stepIds.PROPOSAL, 'proposal'),
-          secret: randomSecret,
+          secret: FlowCommand.ID.data(this.stepIds.PROPOSAL, 'secret'),
```
RodrigoAD:
```suggestion
```

RodrigoAD (commented):
That should fix the proposal match issue I imagine

gauntlet/packages/gauntlet-serum-multisig/src/commands/approve.ts
RodrigoAD: Reopening this, the whole purpose of the multisig command is validation. Here the users will approve proposals they don't know what they do.

If approval batching is required, it should be built within the multisig wrapper, so Gauntlet can make proper validations and inform the users about the actions they'll do.

RodrigoAD (changes requested):

gauntlet/packages/gauntlet-serum-multisig/src/commands/approve.ts
archseer: This is a quick workaround from an immediate need: without batching we need a ledger approval per feed per owner. This at least reduces it to a command (per 6-7 feeds) per owner.

It would be great to fix this in the multisig wrapper, but it's a larger job since we'd need to provide inputs for multiple invocations, verify them then pack that into a single tx. This is slightly tricky UX-wise too, since each command can take additional flags, for example `--secret` per `ocr2:accept_proposal:multisig`.

I think the multisig wrapper needs a general re-work to stop being so stateful too. It's very easy to accidentally call the code without a proposal (or worse yet, the code is faulty and fetches no proposal), which ends up creating new proposals rather than approving.

We've already had that happen on starknet: `proposalId` was missing in the implementation but the command was running successfully (and tests passing), creating new proposals over and over again: https://github.com/smartcontractkit/chainlink-starknet/commit/dacaae3d1dc4a3a21396fd7aad965288ff2f111e#diff-8257597e02534a1f29c2fcb789a25491779bd7ac33359412616550f5c422d83cR208

PR #408, opened 2022-10-12
QA-79 Use projectserum version from makefile in github workflows

tateexon:
Reduces the number of places the projectserum image version is needed to be updated.
The action can be used in the chainlink repo to keep its version in sync with this repo.


aalu1418 (approved):
lgtm

PR #412, opened 2022-10-25
fix: Specify next owner when accepting ownership

archseer:
If a multisig is accepting ownership then we don't want to use the current wallet/signer but the multisig address.

### DIFF
--- a/gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ownership/acceptOwnership.ts
+++ b/gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ownership/acceptOwnership.ts
@@ -15,11 +15,12 @@ export const makeAcceptOwnershipCommand = (
     static id = `${contractId}:accept_ownership`
     static category = contractId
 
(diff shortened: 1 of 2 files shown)


gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/ownership/acceptOwnership.ts
```diff
@@ -31,18 +32,19 @@ export const makeAcceptOwnershipCommand = (
...
+      const authority = new PublicKey(this.flags.authority)
 
       const state = new PublicKey(this.args[0])
 
       const tx = await program.methods
         .acceptOwnership()
         .accounts({
           state: state,
-          authority: signer,
+          authority: authority,
```
RodrigoAD, line 47: This should be sent as the signer from every environment, including multisig. Let's wait to see how the issue resolves

RodrigoAD, line 47: This should be sent as the signer from every environment, including multisig. Let's wait to see how the issue resolves

RodrigoAD (changes requested):

PR #423, opened 2022-11-07
solana payouts more explicit log

aalu1418:


RodrigoAD (approved):
This is definitely a better message, thanks @aalu1418

PR #426, opened 2022-11-10
improve logging for ownership acceptance

aalu1418:


chainchad:
Tested this and it works as expected now, thanks!

PR #431, opened 2022-11-23
Refactoring tests and adding soak fixes

smickovskid:
- Refactored structure and configs
- Moved spec to TOML
- Added Soak runner
- Renamed tests to `integration-tests`


Makefile
```diff
@@ -81,10 +81,9 @@ test_relay_unit:
 	go test -v ./pkg/...
 
 test_smoke:
-	SELECTED_NETWORKS=solana \
-	cd ./tests && go test -timeout 24h -count=1 -json $(args) ./e2e/smoke 2>&1 | tee /tmp/gotest.log | gotestfmt
+	cd ./integration-tests &&\
+	SELECTED_NETWORKS=SIMULATED go test -timeout 24h -count=1 -json $(args) ./smoke 2>&1 | tee /tmp/gotest.log | gotestfmt
```
gheorghestrimtu, line 85: I think it should be `SELECTED_NETWORKS=solana`

tateexon (dismissed):

Makefile
smickovskid: Not added currently in core, we can use simulated

tateexon (approved):

PR #450, opened 2023-01-07
Make fetching the juels/lamport values from on-chain more resiliant

topliceanu:


pkg/monitoring/source_envelope.go
```diff
@@ -133,6 +126,24 @@ func (s *envelopeSource) Fetch(ctx context.Context) (interface{}, error) {
...
+		return nil, fmt.Errorf("failed to read the feed's link balance: %w", err)
+	}
+	if linkBalanceRes == nil || linkBalanceRes.Value == nil {
+		return nil, fmt.Errorf("link balance not found for token vault")
+	}
+	linkBalance, success := big.NewInt(0).SetString(linkBalanceRes.Value.Amount, 10)
+	if !success {
+		return nil, fmt.Errorf("failed to parse link balance value: %s", linkBalanceRes.Value.Amount)
+	}
+	if linkBalance.Cmp(big.NewInt(0)) == 0 {
```
cll-gg: Would it work to move `big.NewInt(0)` into a constant/global variable to prevent allocating a new big int every time?

pkg/monitoring/source_envelope.go
```diff
@@ -162,23 +173,32 @@ func (s *envelopeSource) getJuelsPerLamport(ctx context.Context) (*big.Int, erro
 			},
 		)
 		if err != nil {
-			return nil, fmt.Errorf("failed to fetch tx with signature %s: %w", txSig.Signature, err)
+			s.log.Infow("failed to fetch tx with signature %s: %w", txSig.Signature, err)
```
cll-gg: Not fully sure, but shouldn't this be `s.log.Infof` instead of `s.log.Infow`? See https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Infow

topliceanu: `s.log` is of type `github.com/smartcontractkit/chainlink-relay/pkg/monitoring.Logger` ([file](https://github.com/smartcontractkit/chainlink-solana/blob/9a494f74f9a3883c572d8e4ebee51fa453121c8d/pkg/monitoring/source_envelope.go#L21)) which is a `github.com/smartcontractkit/chainlink-relay/pkg/logger.Logger` ([file](https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/monitoring/logger.go)) which is this https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/logger/logger.go and does support `Infow`.

cll-gg: Indeed `Infow` is supported by the logger but according to [the implementation](https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Infow) `Infow` is meant to use with keyValues as args, while `Infof` is meant to be used with `fmt`-formatted messages. The latter is being used here with `%s` and `%w`, so that's why I expected `Infof`.
However, if it works fine then there's no reason to spend effort on changing this.

topliceanu: Ah yes. You're right! It works as is but it's not ideal, the values are not keyed correctly. I'll fix

topliceanu: Done. Btw, do you know what this failure is https://github.com/smartcontractkit/chainlink-solana/actions/runs/3872239678/jobs/6602413818 I'm not sure I'm causing it!

cll-gg: No clue! When I run [`./scripts/install-solana-ci.sh`](https://github.com/smartcontractkit/chainlink-solana/blob/develop/scripts/install-solana-ci.sh) locally it works properly:

```
<snip>
} # this ensures the entire script is downloaded #'
downloading v1.10.33 installer
  ✨ 1.10.33 initialized
<snip>
```

pedrovsg (approved):

pkg/monitoring/source_envelope.go
topliceanu: it passes now 🤷

PR #475, opened 2023-03-03
Extracted txm + fees from core

sdrug:
[Ticket](https://smartcontract-it.atlassian.net/jira/software/c/projects/BCI/boards/23?modal=detail&selectedIssue=BCI-889&assignee=6164d22dc5388b0069da976b)
Summary of transition:
- Fixed tests
- Got rid of some core dependencies (e.g. testutils)
- Switched to local package analogues (e.g. core logger -> relay logger)
- No changes to the txm logic
- Keystore dependency preserved for now


.github/workflows/relay.yml
```diff
@@ -27,4 +27,4 @@ jobs:
       - name: Test
         run: go test ./pkg/... -v
       - name: Test with the race detector enabled
-        run: go test ./pkg/... -v -race -count=10
+        run: go test ./pkg/... -v -race -count=1 -timeout=15m
```
aalu1418: wondering: why are we changing the race tests?

setting `-count=1` makes the check less effective 🤔

go.mod
```diff
@@ -8,68 +8,268 @@ require (
 	github.com/gagliardetto/gofuzz v1.2.2
 	github.com/gagliardetto/solana-go v1.4.1-0.20220428092759-5250b4abbb27
 	github.com/gagliardetto/treeout v0.1.4
+	github.com/google/uuid v1.3.0
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.12.1
+	github.com/prometheus/client_golang v1.14.0
+	github.com/smartcontractkit/chainlink v1.12.0
```
aalu1418: is it possible to avoid importing core? importing core introduces a circular dependency and also brings a lot of extra unneeded modules

you might have to move/copy some things around to `chainlink-relay` or into `chainlink-solana`

pkg/solana/txm/txm.go
```diff
@@ -0,0 +1,536 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana"
+	solanaClient "github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/config"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/fees"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logger"
+
+	"github.com/smartcontractkit/chainlink/core/services"
+	"github.com/smartcontractkit/chainlink/core/services/keystore"
+	"github.com/smartcontractkit/chainlink/core/utils"
```
aalu1418: i think `services` is just used for the serviceCtx interface which exists inside `chainlink-relay`

`keystore`: this might be tricker...you might need to copy the keystore interface and move `solkey` from core 🤔 

`utils` there might be an equivalent in `chainlink-relay` or it can be copied over to the relay modules

aalu1418 (changes requested):
seems like there's potentially atleast a key type (`solkey`) extraction involved in a clean move of the solana txm out of core

@jmank88 any thoughts on if there's another way?

pkg/solana/txm/txm.go
aalu1418: note these imports also show up in a few test files as well

jmank88: If the keystore is a pain you can use my LOOP branch which minimized the interface

sdrug: Services/utils - 👌 
Regarding keystore - I thought we agreed to not touch it for now. So i left it in the core (also the reason for the core dependency for the comment above). Do you want to copy it over too?

.github/workflows/relay.yml
sdrug: The were primarily two reasons:
- 10 iterations take forever to complete now with the txm tests added
- There is an error in the tests: it fails starting from the second iteration mismatching the prom counters. My first idea was that tests do not cleanup properly after completion but I wasn't able to catch it immediately. Sure I can dig into it more, it's just not the first priority.

jmank88: Race detection is not meant to pass tests, and timeout parameter is how we bound the runtime

jmank88: I would recommend using shuffle like we do in core

pkg/solana/txm/txm.go
jmank88: We cannot create an import cycle, so something different needs to be done.

.github/workflows/relay.yml
sdrug: Those tests are failing with count > 1 without race flag too. To replicate on core develop branch:
`go test ./core/chains/solana/soltxm -v -count=3`

I'll have another look tomorrow, it could be I'm missing something.

pkg/solana/txm/txm.go
sdrug: There is no import cycle currently - the core branch [builds and passes tests](https://github.com/smartcontractkit/chainlink/pull/8627) successfully. Or you mean it happens after the keystore is extracted?

jmank88: Importing chainlink from chainlink-solana creates a cycle. This is not allowed.

sdrug: Well, circular dependency generates compile error. So technically since it compiles we don't have it - makes sense to me cause we aren't importing the whole core dependency tree. But I see what you mean, I'll move over the keystore too.

jmank88: You absolutely are importing the whole tree, that is the problem. You are creating a _module_ cycle, and infecting both repos with all dependencies.

augustbleeds:
please link the ticket in the description so JIRA can pretty link to it :)

.github/workflows/relay.yml
aalu1418: > There is an error in the tests: it fails starting from the second iteration mismatching the prom counters. My first idea was that tests do not cleanup properly after completion but I wasn't able to catch it immediately. Sure I can dig into it more, it's just not the first priority.

ah right that set of tests doesn't work well with race tests cuz of prom

another work around could be to tag that test file with `//go:build integration`

aalu1418: on another note: we need have the CI also run the integration tests so the non-race tests should have `-tags integration` added to it

pkg/solana/txm/txm.go
aalu1418: just to add to what is Jordan is saying: 

we try to avoid importing core into relays because it introduces many unneeded dependencies (in `go.mod`) that the relay does not need - you can see that in these changes where cosmos + ethereum SDKs are now in the dependency tree. this makes it incredibly difficult to update dependencies in core and relays without many PRs that are just bumping core and relay versions (especially if dependencies have breaking changes)

sdrug:
@aalu1418 @jmank88
Please give it another try. Related PRs:
- [Core](https://github.com/smartcontractkit/chainlink/pull/8627)
- [Relay](https://github.com/smartcontractkit/chainlink-relay/pull/88)

aalu1418 (commented):
looking good! I will wait until the PR in `chainlink-relay` gets merged and the reference gets bumped for final approval

pkg/solana/keys/key.go
```diff
@@ -0,0 +1,110 @@
...
+
+// GoString wraps String()
+func (raw Raw) GoString() string {
+	return raw.String()
+}
+
+var _ fmt.GoStringer = &Key{}
+
+// Key represents Solana key
+type Key struct {
```
jmank88: I was not expecting the keystore _implementation_ to be moved. What is the reason for this?

aalu1418: StarkNet is also set up this way - so that might be why?

if I understand correctly, the keystore itself hasn't moved just the base implementation of the solana key

jmank88: But this means that the chainlink repo has to import this module, with all of the chain-specific dependencies, right? We'll just have to move it back eventually in that case.

sdrug: Correct, I used Starknet for the reference.
Correct, only solkey implementation was moved over.

aalu1418: ahhhh right - this gets weird with plugins because we don't want core importing the relay or the relay importing core

jmank88: FWIW this is another reason to only use the minimal interface required by the txm, so that all of this other stuff doesn't even come up.

aalu1418: yeah that makes sense now - I didn't think this far down the line

so that minimal interface needs to live in `chainlink-relay` where `chainlink-solana` doesn't need direct knowledge of the the solana keystore that lives in core

aalu1418: i'm debating if this should be done in a follow up PR 🤔

jmank88: you should be able to keep it even simpler for now, w/o a common interface in relay just yet. but I'm hacking on one now for reference:
https://github.com/smartcontractkit/chainlink-relay/pull/80/files#diff-84405d57992ad70fad77392b91f683b092a624afe5037a279f362b21474f4dfbR36-R39

jmank88: (don't worry about context)

aalu1418: makes sense, so on the `chainlink-solana` side we can define a simple keystore interface that provides a sign method with inputs (context, string address, byte data) and outputs (bytes signature, error)

then on the core side, add an additional keystore method that matches

how should the keystore respond if the key doesn't exist? I'm thinking just throw an error

jmank88: You can wrap the existing pretty trivially:
https://github.com/smartcontractkit/chainlink/pull/8517/files#diff-29f963c1e6175aceb3883554c8a75d78e3fc565dbd374ec1daf54d5036c0bc58R25-R35

jmank88: I think preserving error behavior for now is simplest, but we may have to adjust it later (and I'll be sure to document what the plugin API expects).

aalu1418: alrighty, I will handle this in a PR built on this one and make updates :)

aalu1418:
corresponding core PR with keystore changes: https://github.com/smartcontractkit/chainlink/pull/8627

pkg/solana/keys/key.go
aalu1418: resolving this conversation for now since a simplified keystore interface is now used based on previous discussions

jmank88 (approved):

aalu1418 (approved):

PR #484, opened 2023-03-17
bump chainlink-relay; update Relayer.Healthy()

jmank88:
`chainSet.Healthy()` is no longer available on latest chainlink-relay

Seems like passing the required e2e tests depends on:
- https://github.com/smartcontractkit/chainlink-starknet/pull/217

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.14.0
-	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230309154839-2b6a5b078888
+	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230316183824-9f1e5e11e2b7
 	github.com/smartcontractkit/libocr v0.0.0-20221209172631-568a30f68407
 	github.com/stretchr/testify v1.8.2
 	go.uber.org/multierr v1.8.0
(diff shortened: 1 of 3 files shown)


aalu1418 (approved):
lgtm! just seems like it needs a corresponding bump on the starknet side to merged into core

jmank88:
Hmmm I already updated starknet and thought this would pick that up. Do you mean that the core repo has to include that starknet version too?

aalu1418:
> Hmmm I already updated starknet and thought this would pick that up. Do you mean that the core repo has to include that starknet version too?

yeah - i believe it builds off of core develop. we could also do the thing where we run CI against a branch in core if you want the solana/starknet bumps to happen together?

jmank88:
> > Hmmm I already updated starknet and thought this would pick that up. Do you mean that the core repo has to include that starknet version too?
> 
> yeah - i believe it builds off of core develop. we could also do the thing where we run CI against a branch in core if you want the solana/starknet bumps to happen together?

How do we do that?
This doesn't seem like something that should block merge :thinking:  IIUC it's just forcing more unnecessary incremental PRs in this case.

aalu1418:
> > > Hmmm I already updated starknet and thought this would pick that up. Do you mean that the core repo has to include that starknet version too?
> > 
> > 
> > yeah - i believe it builds off of core develop. we could also do the thing where we run CI against a branch in core if you want the solana/starknet bumps to happen together?
> 
> How do we do that? This doesn't seem like something that should block merge 🤔 IIUC it's just forcing more unnecessary incremental PRs in this case.

it would be like any other breaking interface change...but this one is weird since it's a completely separate dependency that is causing the issue rather than changes here.

I can also override and merge this PR - and then all you need to do is bump the starknet (and maybe relay) dependency in core?

jmank88:
> > > > Hmmm I already updated starknet and thought this would pick that up. Do you mean that the core repo has to include that starknet version too?
> > > 
> > > 
> > > yeah - i believe it builds off of core develop. we could also do the thing where we run CI against a branch in core if you want the solana/starknet bumps to happen together?
> > 
> > 
> > How do we do that? This doesn't seem like something that should block merge thinking IIUC it's just forcing more unnecessary incremental PRs in this case.
> 
> it would be like any other breaking interface change...but this one is weird since it's a completely separate dependency that is causing the issue rather than changes here.
> 
> I can also override and merge this PR - and then all you need to do is bump the starknet (and maybe relay) dependency in core?

Yeah, I thought it was simplest to bump everything at once in core afterwards.

aalu1418:
👍🏼 

Leaving a note: overriding CI check to merge
- breaking changes have been fixed in chainlink-starknet (no logic changes needed for solana)
- all versions will be bumped in core together

PR #485, opened 2023-03-17
fix go mod tidy

tateexon:


jmank88 (approved):

aalu1418 (approved):
@tateexon ping me if you need an override to merge - a starknet bump needs to get merged into core first before the e2e test will build

PR #486, opened 2023-03-18
pkg/solana: remove unnecessary config update method

jmank88:
This is no longer used with static TOML config.


aalu1418 (approved):
:killitwithfire:

jmank88:
@aalu1418 Could you merge when you have a chance? Core update coming next, then the test should start passing again https://github.com/smartcontractkit/chainlink/pull/8751

PR #499, opened 2023-04-18
txm fix race signature tracking race condition when slow RPC

aalu1418:
tickets: 
* https://smartcontract-it.atlassian.net/browse/BCI-1149
* https://smartcontract-it.atlassian.net/browse/BCI-1150


jmank88 (dismissed):

pkg/solana/txm/utils.go
```diff
@@ -69,3 +71,35 @@ func convertStatus(res *rpc.SignatureStatusesResult) uint {
...
+}
+
+func (s *signatureList) List() []solana.Signature {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return s.sigs
+}
+
+func (s *signatureList) Length() int {
+	return len(s.List())
```
jmank88: I think this is OK, but I had to read it carefully and know how slices work to understand that. What about this?
```suggestion
	s.lock.RLock()
	defer s.lock.RUnlock()
	return len(s.sigs)
```
Should be a little more efficient too.

jmank88 (approved):

PR #508, opened 2023-05-11
make logs in ci show as they occur and output a pretty format after

tateexon:


jmank88 (approved):

archseer (approved):
👍🏻 Let's do the same on starknet

PR #512, opened 2023-05-16
fix: race condition in tracking retry signatures

aalu1418:
adding test case for checking race condition in broadcasting and tracking signatures

there are two types of txs:
* (type 1) first bumped tx which saves the tx signature to the tracking list
* (type 2) follow up txs that rebroadcast the bumped tx and compare the rebroadcast sig to the one in the list

there are a few race conditions that can occur depending on client response times (broadcasting pattern is type 1 -> X of type 2 -> type 1 again -> repeat until timeout
* type 1 returns quickly, type 2 returns slowly, new type 1 is broadcast: this mixes up which index is used to check the signature against. Solution: use bump count as the index
* type 1 returns slowly, type 2 returns quickly: the rebroadcast txs return before the original bumped tx. the signature is not present when the type 2 txs need to check against it. Solution: use a waitgroup to which will cause the comparison to wait until the type 1 tx returns and saves the signature
* first type 1 returns slowly, type 2 returns normally, new type 1 returns quickly: this can mix up the order of signatures in the list. Solution: instead of using `append()` allocate an index and set.

### DIFF
--- a/pkg/solana/txm/txm.go
+++ b/pkg/solana/txm/txm.go
@@ -204,16 +204,18 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
 		return solanaGo.Transaction{}, uuid.Nil, solanaGo.Signature{}, errors.Wrap(initSendErr, "tx failed initial transmit")
 	}
 
-	// used for tracking rebroadcasting only in SendWithRetry
-	var sigs signatureList
-	sigs.Allocate()
-	sigs.Set(0, sig)
-
 	// store tx signature + cancel function
 	id, initStoreErr := txm.txs.New(sig, cancel)
 	if initStoreErr != nil {
 		cancel() // cancel context when exiting early
 		return solanaGo.Transaction{}, uuid.Nil, solanaGo.Signature{}, errors.Wrapf(initStoreErr, "failed to save tx signature (%s) to inflight txs", sig)
+	}
+
+	// used for tracking rebroadcasting only in SendWithRetry
+	var sigs signatureList
+	sigs.Allocate()
+	if initSetErr := sigs.Set(0, sig); initSetErr != nil {
+		return solanaGo.Transaction{}, uuid.Nil, solanaGo.Signature{}, fmt.Errorf("failed to save initial signature in signature list: %w", initSetErr)
 	}
 
 	txm.lggr.Debugw("tx initial broadcast", "id", id, "signature", sig)
@@ -226,10 +228,13 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
 		tick := time.After(0)
 		bumpCount := uint(0)
 		bumpTime := time.Now()
+		var wg sync.WaitGroup
+
 		for {
 			select {
 			case <-ctx.Done():
 				// stop sending tx after retry tx ctx times out (does not stop confirmation polling for tx)
+				wg.Wait()
 				txm.lggr.Debugw("stopped tx retry", "id", id, "signatures", sigs.List())
 				return
 			case <-tick:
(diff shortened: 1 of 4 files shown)


pkg/solana/txm/txm.go
```diff
@@ -269,13 +274,17 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
...
+						if setErr := sigs.Set(int(count), retrySig); setErr != nil {
+							// this should never happen
+							txm.lggr.Errorw("INVARIANT VIOLATION", "error", setErr)
+						}
+						txm.lggr.Debugw("tx rebroadcast with bumped fee", "id", id, "fee", getFee(count), "signatures", sigs.List())
 					}
 
-					// this should never happen (should match the last signature saved to sigs)
-					if fetchedSig, fetchErr := sigs.Get(index); fetchErr != nil || retrySig != fetchedSig {
+					sigs.Wait(int(count)) // wait until firts bump tx has set the tx signature
```
jmank88:
```suggestion
					sigs.Wait(int(count)) // wait until first bump tx has set the tx signature
```

pkg/solana/txm/txm.go
```diff
@@ -204,8 +204,10 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
 		return solanaGo.Transaction{}, uuid.Nil, solanaGo.Signature{}, errors.Wrap(initSendErr, "tx failed initial transmit")
 	}
 
+	// used for tracking rebroadcasting only in SendWithRetry
 	var sigs signatureList
-	sigs.Append(sig) // add initial signature
+	sigs.Allocate()
+	sigs.Set(0, sig)
```
jmank88: Unchecked error

jmank88 (dismissed):

pkg/solana/txm/utils.go
```diff
@@ -98,10 +99,42 @@ func (s *signatureList) Length() int {
...
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.sigs[index] = sig
+	s.wg[index].Done()
+	return nil
+}
+
+func (s *signatureList) Wait(index int) *sync.WaitGroup {
```
jmank88: Do all of the callers just call `.Wait()`? If so, could simplify by making this actually block.

yongkangc:
> (type 1) first bumped tx which saves the tx signature to the tracking list
> (type 2) follow up txs that rebroadcast the bumped tx and compare the rebroadcast sig to the one in the list

what is the main difference between type 1 and type 2 in terms of when they are being used in the business logic sense? i.e when would type 1 be called and when would type 2 be called?

pkg/solana/txm/utils.go
```diff
@@ -98,10 +99,42 @@ func (s *signatureList) Length() int {
...
+	return len(s.sigs) - 1
+}
+
+func (s *signatureList) Set(index int, sig solana.Signature) error {
+	v, err := s.Get(index)
+	if err != nil {
+		return err
+	}
+
+	if !v.IsZero() {
```
jmank88, line 129: Does this condition need to be checked again after locking? Otherwise two routines could both pass this and just overwrite.

pkg/solana/txm/utils.go
aalu1418: ah right, i was trying to fix a race test - but yeah there's an even simpler way to do this

aalu1418:
> > (type 1) first bumped tx which saves the tx signature to the tracking list
> > (type 2) follow up txs that rebroadcast the bumped tx and compare the rebroadcast sig to the one in the list
> 
> what is the main difference between type 1 and type 2 in terms of when they are being used in the business logic sense? i.e when would type 1 be called and when would type 2 be called?

@ExtremelySunnyYK right now the strategy for broadcasting on solana is to basically rebroadcast transactions with an exponential backoff and combine it with fee bumping. so the flow looks something like the diagram below - any txs broadcast within the `rebroadcast` label are type 2, and any txs like `initialTx` or `BumpedTx` are type 1

```mermaid
stateDiagram-v2
    [*] --> InitialTx: new tx
    InitialTx --> InitialTx: rebroadcast
    InitialTx --> BumpedTx: fee bump
    BumpedTx --> BumpedTx: rebroadcast
    BumpedTx --> [repeat]: continue bumping and rebroadcasting
    [repeat] --> [*]: confirmation/timeout
```

jmank88 (approved):

PR #520, opened 2023-06-06
Added metrics to smoke

smickovskid:
- Added gauntlet support
- Added Testnet deployment support
- Added gauntlet wrapper
- Added scheduled workflow that will run on testnet
- Added method to transfer funds
- Added additional details to the gauntlet `ocr2:inspect:responses` command to return the last 10 rounds with details

New programs have been deployed on testnet and the PRIVATE_KEY will be added to the repo secrets which owns the programs so we don't deploy each time.

There are some repetitions in the job, chain, node creation, marked those as TODO since a bigger refactor will be required since we can't fully port out of using solana CLI with gauntlet due to K8 forwarding having issues with UDP packets. The TODO will decouple the job spec, bridges and all related actions into one method but for now we need testnet daily tests so that will be done in the future.


contracts/programs/access-controller/src/lib.rs
```diff
@@ -4,7 +4,7 @@ use std::mem;
 
 use arrayvec::arrayvec;
 
-declare_id!("9xi644bRR8birboDGdTiwBq3C7VEeR7VuamRYYXCubUW");
+declare_id!("5Gm8JySC3RzGonihF1hj1kZkjHyyW7qxJNvyzq1ieAY5");
```
aalu1418: these shouldn't change - they correspond to production program addresses on mainnet & devnet

gauntlet/packages/gauntlet-solana-contracts/networks/.env.devnet
```diff
@@ -1,8 +1,8 @@
 NODE_URL=https://api.devnet.solana.com
 
-PROGRAM_ID_OCR2=cjg3oHmg9uuPsP8D6g29NWvhySJkdYdAo9D25PRbKXJ
-PROGRAM_ID_ACCESS_CONTROLLER=9xi644bRR8birboDGdTiwBq3C7VEeR7VuamRYYXCubUW
-PROGRAM_ID_STORE=HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny
+PROGRAM_ID_OCR2=9v5sGbNhyTM23yVvczzPbk5NtCqrARQpZjuzebBur43h
+PROGRAM_ID_ACCESS_CONTROLLER=5Gm8JySC3RzGonihF1hj1kZkjHyyW7qxJNvyzq1ieAY5
+PROGRAM_ID_STORE=ANooJg7j9UxgJ2DNuLqUJs5h956qzc5LdPLFFtQdDb8M
```
aalu1418: these shouldn't change - they correspond to production program addresses on mainnet & devnet which would break the engops workflow

it might be better to create something a new `.env.smoke` file instead to have the specific programs

contracts/programs/access-controller/src/lib.rs
aalu1418: note this comment is also relevant for the other 2 programs too

integration-tests/common/test_common.go
```diff
@@ -52,12 +54,77 @@ type Contracts struct {
...
+	Signer      string `json:"signer"`
+	Transmitter string `json:"transmitter"`
+	Payee       string `json:"payee"`
+}
+
+type OCR2OnChainConfig struct {
+	Oracles    []Oracle `json:"oracles"`
+	F          int      `json:"f"`
+	ProposalId string   `json:"proposalId"`
+}
```
aalu1418: just curious - are these types already defined somewhere? they feel very familiar - could they be imported from libocr?

aalu1418: if not - a good follow up would be to move these to somewhere so other e2e test could use it

integration-tests/gauntlet/gauntlet_solana.go
```diff
@@ -0,0 +1,514 @@
...
+	_, err = sg.FinalizeProposal(sg.ProposalAddress)
+	if err != nil {
+		return err
+	}
+	_, err = sg.AcceptProposal(sg.ProposalAddress, "this is an testing only secret", proposalAccept, sg.OcrAddress)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
aalu1418, line 514: this would be a stretch goal to include:
* testing the payout mechanism works (harder)
* closing the feed works (this would be easier if the billing was just set to 0)

integration-tests/solclient/solclient.go
```diff
@@ -485,3 +488,64 @@ func (c *Client) GasStats() *blockchain.GasStats {
 func (c *Client) AddHeaderEventSubscription(key string, subscriber blockchain.HeaderEventSubscription) {
 	panic("implement me")
 }
+
+func SendFunds(senderPrivateKey string, receiverPublicKey string, lamports uint64, rpcClient *rpc.Client, wsClient *ws.Client) error {
+
```
aalu1418, line 493: is there a method to recover funds from the CL nodes once the test wraps up?

integration-tests/gauntlet/gauntlet_solana.go
aalu1418: if you close out the feed, you would also recover some SOL back to the original deployer

aalu1418 (changes requested):

contracts/programs/access-controller/src/lib.rs
smickovskid: Reverted

gauntlet/packages/gauntlet-solana-contracts/networks/.env.devnet
smickovskid: Reverted, was not needed after all since I dynamically create the gauntlet network file

integration-tests/common/test_common.go
smickovskid: Yes, we reuse this in Starknet and we would anywhere we use OCRV2, this should be handled in the new bix testing framework. 

I checked libocr, but there is a discrepancy in the json naming between gauntlet and libocr

https://github.com/smartcontractkit/libocr/blob/2793dc05923e226c8783deca106ff0f7fd36d686/offchainreporting2/internal/config/offchainreporting2_offchain_config.pb.go#L23-L44

integration-tests/solclient/solclient.go
smickovskid: I tried implementing it but it will require a bit more time, having issues decoding the private key. Create a Jira for it so I tackle it separately

integration-tests/gauntlet/gauntlet_solana.go
smickovskid: Created task to followup on this as well

integration-tests/gauntlet/gauntlet_solana.go
```diff
@@ -0,0 +1,514 @@
...
+	_, err = sg.StoreSetValidatorConfig(sg.FeedAddress, 8000)
+	if err != nil {
+		return "", err
+	}
+
+	ocr2Config := &OCR2Config{
+		MinAnswer:     "0",
+		MaxAnswer:     "10000000000",
+		Transmissions: sg.FeedAddress,
+	}
```
aalu1418, line 462: something also worth testing in the future: positive and negative values

aalu1418 (dismissed):
just something for the future, but otherwise looks good to me!

aalu1418 (approved):

PR #538, opened 2023-08-07
Fixing testnet run

smickovskid:
- Fixing testnet run


krebernisak (approved):

integration-tests/gauntlet/gauntlet_solana.go
```diff
@@ -422,6 +422,10 @@ func (sg *SolanaGauntlet) FetchTransmissions(ocrState string) ([]Transmission, e
 
 func (sg *SolanaGauntlet) DeployOCR2() (string, error) {
 	var err error
+	err = sg.InstallDependencies()
```
krebernisak, line 425: We'll try to simplify all of these dependencies with Nix and Gauntlet++.

Gauntlet CLI (singular, core CLI) should be available in every env, and then we only need to set it up with local plugins. I see this as a specific project task (Gauntlet test), where before execution we'll enter a specific Nix env for that task (or general project shell env as default), and then execute the Go test code to run the test.

cc @archseer

PR #545, opened 2023-09-26
[TT-414] Solana Dockerization

tateexon:
Converts TestSolanaOCRV2Smoke to use the dockerized environment to run the test in CI.
Does not change TestSolanaGauntletOCRV2Smoke or the soak test.


.github/actions/build_contract_artifacts/action.yml
```diff
@@ -22,7 +22,7 @@ runs:
         repository: smartcontractkit/chainlink-solana
         ref: ${{ inputs.ref }}
     - name: Setup go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
         go-version-file: "go.mod"
```
jmank88, line 27: nit: isn't this the default value?

integration-tests/docker/test_env/sol.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+	c, err := tc.GenericContainer(context.Background(), tc.GenericContainerRequest{
+		ContainerRequest: *cReq,
+		Reuse:            true,
+		Started:          true,
+		Logger:           l,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "cannot start Solana container")
```
jmank88: Would std lib errors be sufficient here?
```suggestion
		return fmt.Errorf("cannot start Solana container: %w", err)
```

jmank88 (dismissed):

.github/actions/build_contract_artifacts/action.yml
tateexon: Doesn't look like it: https://github.com/actions/setup-go/blob/main/action.yml#L7-L8

jmank88 (dismissed):

.github/actions/build_contract_artifacts/action.yml
jmank88: hmm, but doesn't this suggest it will find it?
https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

jmank88 (approved):

PR #547, opened 2023-10-04
pkg/solana: copy chains/solana packages from core

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2684

This PR copies core/chains/solana and core/chains/solana/monitor from the chainlink repo.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/10860

Requies:
- https://github.com/smartcontractkit/chainlink-relay/pull/187

PR #550, opened 2023-10-10
BCF-2685: move LOOPP command from core

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2685

Blocked by:
- #547

Supports:
- https://github.com/smartcontractkit/chainlink/pull/10973

Requires:
- https://github.com/smartcontractkit/chainlink-relay/pull/193


krehermann (dismissed):

patrickhuie19 (dismissed):

patrickhuie19 (dismissed):

pkg/solana/cmd/chainlink-solana/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana"
+)
+
+const (
+	loggerName = "PluginSolana"
+)
+
+func main() {
+	s := loop.MustNewStartedServer(loggerName)
```
krehermann, line 20: What happens with this logger when there are multple loopps running? Does it  some unique-ification?

pkg/solana/cmd/chainlink-solana/main.go
```diff
@@ -0,0 +1,75 @@
...
+	var cfg struct {
+		Solana solana.TOMLConfig
+	}
+
+	if err := d.Decode(&cfg); err != nil {
+		return nil, fmt.Errorf("failed to decode config toml: %w:\n\t%s", err, config)
+	}
+
+	opts := solana.ChainOpts{
+		Logger:   c.Logger,
```
krehermann, line 63: Here, too, i'm concerned about the logger name being unique

pkg/solana/cmd/chainlink-solana/main.go
jmank88: Ah right we might need to audit these since switching to one chain per relayer, and move chain id higher up. Same issue on all chains.

krehermann: Yeah, at some point I took care to name the logger uniquely when creating a relayer in core, not sure if that is still there and gives and ideas

jmank88: So in theory we would just have to change `<family>.<id>` to `<family>-<id>`, but it's not that simple because we don't have chain id yet when we boot up - maybe we should inject it via env var?

On the other hand, it would be nice to not have to rely on the plugin to do things correctly, and we are already piping the logs back through the host anyways. Instead of just passing them through named as-is, we could set the relay ID prefix on the host side.
https://github.com/smartcontractkit/chainlink-relay/blob/4c281ae24783040330c1ef13a609d2090ac300b1/pkg/loop/logger.go#L63

jmank88:
@aalu1418 This is stable and e2e is passing on core side:
- https://github.com/smartcontractkit/chainlink/pull/10973
- https://github.com/smartcontractkit/chainlink/actions/runs/6600154881/job/17930112425?pr=10973

pkg/solana/cmd/chainlink-solana/main.go
jmank88: Actually, I think we are already prefixing on the host side.
I wrote this up before realizing that: https://smartcontract-it.atlassian.net/browse/BCF-2709
Well we can double check and lock it in with a regression test anyways :shrug:

krehermann (approved):

patrickhuie19 (approved):

aalu1418 (approved):

PR #555, opened 2023-11-16
switch from chainlink-relay to chainlink-common

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2784

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/243


jmank88:
There was a bit of scope creep on this one, but only to swap out deprecated, alias types for their identical underlying type.

aalu1418 (approved):

PR #610, opened 2024-02-27
Skeleton Implementation of Solana Chain Reader Service

EasterTheBunny:
This commit adds a new package that contains a limited chain reader for solana that implements the basic interfaces of ServiceCtx and ChainReader. All functions required for ServiceCtx are complete and included in a test. All functions required for ChainReader return an 'unimplemented' error.

### DIFF
--- a/pkg/solana/chain_reader.go
+++ b/pkg/solana/chain_reader.go
@@ -1,0 +1,76 @@
+package solana
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	commonservices "github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/solana/chain_reader.go
```diff
@@ -0,0 +1,76 @@
+package solana
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	commonservices "github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+	commontypes "github.com/smartcontractkit/chainlink-common/pkg/types"
```
reductionista: You have `"github.com/smartcontractkit/chainlink-common/pkg/types"` imported twice here, either one is fine but we don't need both.

EasterTheBunny: Oh shoot. This was the result of some copy pasta. I'll fix that.

reductionista (approved):
LGTM

PR #612, opened 2024-03-04
Implement Basic GetLatestValue for Solana

EasterTheBunny:
This commit fills in the functionality for `GetLatestValue` such that one or more accounts with an associated Anchor IDl can be read and mapped to native Go structs. Methods and prodecures can be defined in a config where an IDL and codec modifiers can be provided.


pkg/solana/chain_reader_test.go
```diff
@@ -1,21 +1,37 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/config"
+)
+
+const (
+	Namespace   = "NameSpace"
+	NamedMethod = "NamedMethod1"
 )
 
 func TestSolanaChainReaderService_ServiceCtx(t *testing.T) {
 	t.Parallel()
```
nolag: Please make sure to run the chain reader tests from the common package?

pkg/solana/codec/solana_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/test-go/testify/require"
+
+	codeccommon "github.com/smartcontractkit/chainlink-common/pkg/codec"
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codec"
+)
+
+func TestNewIDLCodec(t *testing.T) {
+	t.Parallel()
```
nolag, line 22: Please make sure to run the interface tests from common

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
+/*
+Package codec provides functions to create a codec from an Anchor IDL using NewIDLCodec. All primitives map 1:1, however
+some common types are either not supported or may not have expected behavior.
```
nolag: I don't like the second part.  Not supported is one thing, unexpected behaviour scares me.

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
...
+
+Using time.Time as a native struct for encoding or decoding to and IDL unixTimestamp primitive will not work without
+the codec modifier `ModifierEpochToTime`. The native value to represent timestamps is int64.
```
nolag: I think it's only a few lines change in common to fix this.  Probably worth it.  I think I pointed you to the code

pkg/solana/chain_reader_test.go
```diff
@@ -35,3 +51,223 @@ func TestSolanaChainReaderService_ServiceCtx(t *testing.T) {
...
+	Option           *string
+	DefinedArray     []codec.ObjectRef2
+	BasicVector      []string
+	TimeVal          int64
+	DurationVal      time.Duration
+	PublicKey        ag_solana.PublicKey
+	EnumVal          uint8
+}
+
+type mockedRPCClient struct {
```
nolag: I'm scared of using a mock here.  Can the test have the Solana test validator running?  It's ok to get a docker for that or similar.

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
...
+
+func asStruct(
+	def IdlTypeDef,
+	refs *codecRefs,
+	name string,
+	caser cases.Caser,
+) (string, encodings.TypeCodec, error) {
+	if def.Type.Fields == nil {
+		return name, nil, fmt.Errorf("%w: provided def type fields should not be nil", types.ErrInvalidConfig)
+	}
```
nolag: may need to be careful here.  It's valid in EVM to have nothing, you can just use encodings.Empty

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
...
+
+func asOption(opt *IdlTypeOption, refs *codecRefs) (encodings.TypeCodec, error) {
+	// Go doesn't have an `Option` type; use pointer to type instead
+	// this should be automatic in the codec
+	codec, err := processFieldType(opt.Option, refs)
+	if err != nil {
+		return nil, err
+	}
+
+	return codec, nil
```
nolag: just return processFieldType(...)

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// already exists as a type in the typed codecs
+	if savedCodec, ok := refs.codecs[definedName.Defined]; ok {
+		return savedCodec, nil
+	}
+
+	// codec by defined type doesn't exist
+	// process it using the provided typeDefs
+	nextDef := refs.typeDefs.GetByName(definedName.Defined)
```
nolag, line 185: Do you need to verify circular dependencies?  Like if A has a pointer to B and B has a pointer to A?

pkg/solana/codec/solana.go
```diff
@@ -0,0 +1,297 @@
+/*
+Package codec provides functions to create a codec from an Anchor IDL using NewIDLCodec. All primitives map 1:1, however
+some common types are either not supported or may not have expected behavior.
+
+Using time.Time as a native struct for encoding or decoding to and IDL unixTimestamp primitive will not work without
+the codec modifier `ModifierEpochToTime`. The native value to represent timestamps is int64.
+
+Enums are only supported in their basic form of uint8 values. Enums with variants are not supported.
+*/
+package codec
```
nolag, line 20: Will you need new modifiers to extract fields?  Since you get back the whole account, the part you want to read may be a subfield.

pkg/solana/codec/solana.go
EasterTheBunny: In the case of `nil`, would returning an empty struct suffice?

pkg/solana/bindings.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+
+	if len(bindings) == 1 {
+		// get the item type from the binding codec
+		return bindings[0].CreateType(forEncoding)
+	}
+
+	// default to map when multiple bindings exist
+	return &map[string]any{}, nil
```
nolag: I don't think this will work.  Big ints being decoded to any can cause overflow errors.  If this is so you can read from multiple accounts, I think you need to make a merged struct with the two results.

pkg/solana/codec/testutils.go
```diff
@@ -0,0 +1,267 @@
+package codec
```
nolag: I would put this in a different package like testutils under codec.  I don't like when test functions leak to the implementation.

pkg/solana/codec/testutils.go
```diff
@@ -0,0 +1,267 @@
...
+type ObjectRef2 struct {
+	Prop1 uint32
+	Prop2 *big.Int
+	Prop3 int16
+	Prop4 int32
+	Prop5 int64
+	Prop6 []byte
+}
+
+const jsonIDLWithAllTypes = `{
```
nolag: Please put this in a .json file and use // go:embed

pkg/solana/codec/time.go
```diff
@@ -0,0 +1,54 @@
+package codec
```
nolag: nit: File name should be duration.  I don't recall, were you going to add time.Time later?  I thought you did the work to get it working in common.

Also, do you need set or map primitives that rust provides?

pkg/solana/config/chain_reader.go
```diff
@@ -0,0 +1,37 @@
...
+	ProcedureTypeInternal ProcedureType = iota
+	ProcedureTypeAnchor
+)
+
+type ChainReaderProcedure chainDataProcedureFields
+
+type chainDataProcedureFields struct {
+	// IDLAccount refers to the account defined in the IDL.
+	IDLAccount string `json:"idlAccount"`
+	// Type describes the procedure type to use such as internal for static values,
```
nolag, line 30: What's an example of internal for static values?  Why wouldn't that just have modifiers to make something hard-coded?

pkg/solana/config/chain_reader.go
```diff
@@ -0,0 +1,37 @@
...
+	// anchor-read for using an anchor generated IDL to read values from an account,
+	// or custom structure for reading from a native account.
```
nolag: I don't think you support the custom structure yet.  It might be worth saying that the currently supported value is anchor-read, but that others will be added in future for X

pkg/solana/codec/testutils.go
EasterTheBunny: If I put it under `codec`, it creates a circular dependency. I had thought about putting it in `pkg` or `internal`, but that felt really messy. Maybe `internal/codec/testutils`?

pkg/solana/codec/time.go
EasterTheBunny: I added a hook for `time.Time`. The default for `unixTimestamp` is int64 and I added a hook to allow mappings between the two.

The IDL does not include set or map primitives as far as I could tell, which was strange. The issue is that when parsing the IDL, if it doesn't have those structures available, I don't have a starting point in building those types with the codec.

pkg/solana/config/chain_reader.go
EasterTheBunny: So far this is a placeholder. Hard-coding might make sense, but the CLIP introduced the concept of local procedures that don't exactly use RPC calls. I'm not sure what those are yet so I left the option to add them.

pkg/solana/config/chain_reader.go
EasterTheBunny: I can do that. More explanation would help in usability.

pkg/solana/bindings.go
EasterTheBunny: Such as looping through the bindings and creating a struct from all the types returned from each?

pkg/solana/codec/testutils.go
EasterTheBunny: I can do that when I move it to another package.

pkg/solana/codec/testutils.go
EasterTheBunny: I removed the function that would create a circular dependency and moved the contents to `testutils` in `codec`. I also moved json to files and embedded them.

pkg/solana/bindings.go
EasterTheBunny: I used `reflect.StructOf` to create a merged struct.

pkg/solana/chain_reader_test.go
EasterTheBunny: TODO: https://smartcontract-it.atlassian.net/browse/BCF-3060

pkg/solana/chain_reader_test.go
```diff
@@ -35,3 +66,596 @@ func TestSolanaChainReaderService_ServiceCtx(t *testing.T) {
...
+		}
+	}
+
+	r.reader.service = svc
+	r.reader.client = client
+
+	return r.reader
+}
+
+type wrappedTestChainReader struct {
```
nolag: Please add the TODO comment with a ticket number for replacing this with an actual Solana test validator :).

pkg/solana/chain_reader_test.go
```diff
@@ -35,3 +66,596 @@ func TestSolanaChainReaderService_ServiceCtx(t *testing.T) {
...
+		{Name: strings.Join([]string{AnyContractName, DifferentMethodReturningUint64, "0"}, "."), Address: r.address[2], Pending: true},
+		{Name: strings.Join([]string{AnyContractName, MethodReturningUint64Slice, "0"}, "."), Address: r.address[3], Pending: true},
+		{Name: strings.Join([]string{AnyContractName, MethodReturningSeenStruct, "0"}, "."), Address: r.address[4], Pending: true},
+		{Name: strings.Join([]string{AnySecondContractName, MethodReturningUint64, "0"}, "."), Address: r.address[5], Pending: true},
+	}
+}
+
+func (r *chainReaderInterfaceTester) MaxWaitTimeForEvents() time.Duration {
+	// From trial and error, when running on CI, sometimes the boxes get slow
+	maxWaitTime := time.Second * 20
```
nolag: not sure that's true for this, you're using a mock so it should return right away.  1s could be good until you have a real implementation.

nolag (dismissed):
Approved with a couple small comments.

pkg/solana/chain_reader_test.go
EasterTheBunny: Ahh, in the code. Will do

pkg/solana/chain_reader_test.go
EasterTheBunny: This will have no effect until events are implemented.

pkg/solana/config/chain_reader.go
reductionista: I think we can remove the comment.  static hard coding was the only concrete reason we introduced it in the CLIP, which the codec now takes care of instead.  I was thinking the idea of "internal" methods that don't have to hit the actual smart contract might be useful for other things, but since we have no known need for it yet... we can always re-introduce it later if it turns out there is one

reductionista (dismissed):

reductionista:
LGTM!

pkg/solana/codec/duration.go
aalu1418, line 1: is this solana specific or would it be better in `chainlink-common`?

pkg/solana/example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/codec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExample(t *testing.T) {
```
aalu1418: nit: can we name this something more specific?

aalu1418 (changes requested):
could we move the chainreader components into `pkg/solana/chainreader`?

i'd like to try and organize things in the repo a little better :)

pkg/solana/example_test.go
EasterTheBunny: This file should be removed. It doesn't do anything useful for test coverage.

pkg/solana/codec/duration.go
EasterTheBunny: It was to satisfy the IDL duration type, so at the moment it is solana specific.

aalu1418 (dismissed):
concerns addressed!
I will leave the logic review to @reductionista and @nolag - who approved previously

aalu1418 (approved):

PR #614, opened 2024-03-11
Keystone: forwarder + write target

archseer:


pkg/solana/capabilities/targets/write_target.go
```diff
@@ -0,0 +1,214 @@
...
+	// No signature validation in the MVP demo
+	signatures := [][]byte{} // TODO: validate each sig is 64 bytes
+
+	data = append(data, uint8(len(signatures))) // length prefix
+	for _, sig := range signatures {
+		data = append(data, sig...)
+	}
+
+	// TODO: encode inputs into data
+	data = append(data, inputs["report"].([]byte)...)
```
patrick-dowell, line 156: looks like we are missing the evaluateParams step from the EVM side. Any reason this isn't included the same way it was done on EVM? Seems weird to pull the data from "report" vs the more complex eval we are doing on EVM. If that's what the TODO is for, no worries :)

pkg/solana/capabilities/targets/write_target.go
```diff
@@ -0,0 +1,214 @@
...
+		accounts = append(accounts, &sdk.AccountMeta{
+			PublicKey:  publicKey,
+			IsWritable: account.IsWritable,
+			IsSigner:   account.IsSigner,
+		})
+	}
+
+	tx, err := sdk.NewTransaction(
+		[]sdk.Instruction{
+			sdk.NewInstruction(config.ForwarderProgramID, accounts, data),
```
patrick-dowell, line 184: why does solana use multiple accounts here?

pkg/solana/capabilities/targets/write_target.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error on Transmit.NewTransaction")
+	}
+
+	if err = cap.chain.TxManager().Enqueue(authority.String(), tx); err != nil {
+		return err
+	}
+
+	go func() {
+		// TODO: cast tx.Error to Err (or Value to Value?)
```
patrick-dowell, line 198: In this case, we are just returning any errors and no values right? I'm assuming the solana manager doesn't have any sort of transaction receipt or a blockhash to confirm that the transaction has finalized.

patrick-dowell (commented):
Left a few comments

bolekk:
What are the next steps here? How can we move this forward?

archseer:
I was hoping to discuss as a parking lot item today, we can merge the current design and only handle success route for now. Failure handling will require one of the suggested solutions which will need changes outside the write target

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+        // );
+
+        let hash = hash::hash(&raw_report).to_bytes();
+
+        let raw_signatures = raw_signatures.chunks(SIGNATURE_LEN);
+        for signature in raw_signatures {
+            // TODO:
+        }
+
+        // check if PDA exists, if so terminate the call
```
augustbleeds, line 86: when would the PDA exist? there's no functionality for actually initializing the forwarder PDA in this program

PR #617, opened 2024-03-14
BCF-3016 Convert Multiple Binding Calls to Parallel

EasterTheBunny:
The binding calls are expected to be run in sequence as they map over
the returnVal, but the associated RPC calls can be run in parallel.
This commit preloads all data over RPC in parallel before mapping over
the returnVal in sequence.

### DIFF
--- a/pkg/solana/chainreader/account_read_binding.go
+++ b/pkg/solana/chainreader/account_read_binding.go
@@ -3,7 +3,6 @@ package chainreader
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/gagliardetto/solana-go"
 
--- a/pkg/solana/chainreader/account_read_binding.go
+++ b/pkg/solana/chainreader/account_read_binding.go
@@ -3,6 +3,7 @@ package chainreader
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/gagliardetto/solana-go"
 
@@ -15,6 +16,8 @@ type BinaryDataReader interface {
 	ReadAll(context.Context, solana.PublicKey) ([]byte, error)
 }
 
+const maxResultAge = time.Minute
+
 // accountReadBinding provides decoding and reading Solana Account data using a defined codec. The
 // `idlAccount` refers to the account name in the IDL for which the codec has a type mapping.
 type accountReadBinding struct {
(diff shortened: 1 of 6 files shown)


pkg/solana/chainreader/chain_reader.go
```diff
@@ -91,8 +93,35 @@ func (s *SolanaChainReaderService) GetLatestValue(ctx context.Context, contractN
...
+
+		switch tReturnVal.Kind() {
+		case reflect.Struct, reflect.Map:
+		default:
+			return fmt.Errorf("%w: multiple bindings is only supported for struct and map", types.ErrInvalidType)
+		}
+
+		// for multiple bindings, preload the remote data in parallel
+		for _, binding := range bindings {
+			go func(ctx context.Context, rb readBinding) {
```
jmank88: It looks like the lifecyles of these goroutines needs to be managed carefully, since we can short circuit down below and return error before waiting for all the bindings. The simplest thing to do would be to cancel a context and block on a local wait group before returning. If that introduces weird latency, then they could still be "abandoned" from this method as long as `Close()`ing this `SolanaChainReaderService` waits for them. That would be more complicated to wire up though.

pkg/solana/chainreader/account_read_binding.go
```diff
@@ -22,14 +22,66 @@ type accountReadBinding struct {
...
+		chLoadedBts: make(chan []byte, 1),
+		chLoadErr:   make(chan error, 1),
+	}
 }
 
 var _ readBinding = &accountReadBinding{}
 
-func (b *accountReadBinding) GetLatestValue(ctx context.Context, _ any, outVal any) error {
+func (b *accountReadBinding) PreLoad(ctx context.Context) {
+	b.clearLoaded()
```
jmank88: Why do these need to be cleared each time? Are they re-used?

If this is necessary, then I think there is a race since this is called from a goroutine which could be delayed an arbitrary amount of time. So if there was something to clear, it is possible that the second for loop receives from the channel on the main goroutine before this clear call happens, delivering the data that was meant to be cleared.

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: I decided to wrap the incoming context with a cancel context and add a wait group. Then I cancel the context on exit and wait for the wait group to complete. The only caveat is that if the remote call doesn't respect the context and exit, the call to `GetLatestValue` on the service could hang.

Also I wired up another wait group on the service `Close` function. The only issue there is that it doesn't call cancel to any existing processes.

pkg/solana/chainreader/account_read_binding.go
EasterTheBunny: Instead of maintaining a single result/error channel in the binding, a result struct is provided to both `PreLoad` and `GetLatestValue` on the binding. This allows both calls to be linked by a single result struct that is isolated from other calls either in the same loop or in a different loop.

jmank88 (dismissed):

pkg/solana/chainreader/account_read_binding.go
```diff
@@ -15,6 +16,8 @@ type BinaryDataReader interface {
 	ReadAll(context.Context, solana.PublicKey) ([]byte, error)
 }
 
+const maxResultAge = time.Minute
+
```
aalu1418: what's this for and can we either remove/document? seems like it's more of a config value?

EasterTheBunny: Shoot, that needs to be removed.

aalu1418 (dismissed):

aalu1418 (approved):

PR #654, opened 2024-04-03
SOM: balance monitoring

aalu1418:
https://smartcontract-it.atlassian.net/browse/BCI-2900

- [x] reorganize
- [x] renaming
- [x] improve test case coverage
- [x] add node balance monitoring
- [x] add node balance monitoring tests
- [x] ~https://github.com/smartcontractkit/chainlink-common/pull/449~ (superseded)
- [x] https://github.com/smartcontractkit/chainlink-common/pull/467

Primary additions are:
- source_node_balances.go
- metrics/nodebalances.go
- exporters/nodebalances.go
everything else is reorganization


pkg/monitoring/source_feed_balances.go
```diff
@@ -0,0 +1,135 @@
...
+		return types.Balances{}, fmt.Errorf("balancesSource.addresses is nil")
+	}
+	if len(s.addresses) == 0 {
+		return types.Balances{}, nil
+	}
+
+	for key, address := range s.addresses {
+		go func(key string, address solana.PublicKey) {
+			defer wg.Done()
+			res, err := s.client.GetBalance(ctx, address, rpc.CommitmentProcessed)
```
topliceanu, line 112: I initially thought that the reason why you are instantiating this source with all the nodes is that you can get the balances for multiple accounts in one PRC call! 

Is this not possible?

If it's not, I think it would be clearer if you build a source for each nodeconfig: it's the same number of goroutines but the concurrency is handled for you by common/monitoring; it also simplifies the code, particularly the exporter

topliceanu (approved):
Looks alright

pkg/monitoring/source_feed_balances.go
aalu1418: that makes sense! unfortunately there's not a single RPC call that could be used

I agree that it would be cleaner to have the node iteration to be at a higher level, but allowing the source/exporter to determine the loop implementation makes the network metrics more flexible!

for example: a node balance source could implement a loop, and a block height source would not need to. both would be able to fit into the network source type

PR #672, opened 2024-04-19
Add discriminator to the account and allow encoding/decoding of non-account types

nolag:
When an account is stored from an Anchor contract on Solana, a discriminator is added as the first 8 bytes to ensure that the right data is being loaded.  This was missed when decoding the account.

Additionally, users can define types and should be able to encode them to either sign or write to chain. This PR adds that ability as well.

### DIFF
--- a/pkg/solana/codec/solana.go
+++ b/pkg/solana/codec/solana.go
@@ -60,9 +60,18 @@ func NewNamedModifierCodec(original types.RemoteCodec, itemType string, modifier
 	return modCodec, err
 }
 
-// NewIDLCodec is for Anchor custom types
-func NewIDLCodec(idl IDL, builder encodings.Builder) (types.RemoteCodec, error) {
-	accounts := make(encodings.LenientCodecFromTypeCodec)
+// NewIDLAccountCodec is for Anchor custom types
+func NewIDLAccountCodec(idl IDL, builder encodings.Builder) (types.RemoteCodec, error) {
+	return newIDLCoded(idl, builder, idl.Accounts, true)
+}
+
+func NewIDLDefinedTypesCodec(idl IDL, builder encodings.Builder) (types.RemoteCodec, error) {
+	return newIDLCoded(idl, builder, idl.Types, false)
+}
+
+func newIDLCoded(
+	idl IDL, builder encodings.Builder, from IdlTypeDefSlice, includeDiscriminator bool) (types.RemoteCodec, error) {
+	typeCodecs := make(encodings.LenientCodecFromTypeCodec)
 
 	refs := &codecRefs{
 		builder:      builder,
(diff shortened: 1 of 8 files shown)


pkg/solana/codec/solana_test.go
```diff
@@ -35,11 +39,32 @@ func TestNewIDLCodec(t *testing.T) {
 	require.Equal(t, expected, decoded)
 }
 
+func TestNewIDLDefinedTypesCodecCodec(t *testing.T) {
+	/// TODO this should run the codec interface tests
```
jmank88: TODO for this PR?

nolag: I wasn't planning on it because I just adjusted from the existing test.  If I have time, I can come back to it in another PR.

pkg/solana/codec/solana.go
```diff
@@ -123,8 +133,17 @@ func asStruct(
...
 	name string, // name is the struct name and can be used in dependency checks
 	caser cases.Caser,
+	includeDiscriminator bool,
 ) (string, encodings.TypeCodec, error) {
-	named := make([]encodings.NamedTypeCodec, len(*def.Type.Fields))
+	desLen := 0
+	if includeDiscriminator {
+		desLen = 1
+	}
+	named := make([]encodings.NamedTypeCodec, len(*def.Type.Fields)+desLen)
```
aalu1418, line 142: is there a nil pointer concern for dereferencing?

aalu1418 (commented):
overall looks good to me, just one small concern

pkg/solana/codec/solana.go
nolag: I didn't add it in this PR, but I don't think so.  The call only gets made if the type is a struct so I expect the fields would be present but empty if there are no fields.

There are a few derefs later on as well.

aalu1418 (dismissed):
lgtm for the solana specifics!

ideally someone from the chainreader side can also approve

pkg/solana/codec/solana.go
```diff
@@ -60,9 +60,18 @@ func NewNamedModifierCodec(original types.RemoteCodec, itemType string, modifier
...
+// NewIDLAccountCodec is for Anchor custom types
+func NewIDLAccountCodec(idl IDL, builder encodings.Builder) (types.RemoteCodec, error) {
+	return newIdlCoded(idl, builder, idl.Accounts, true)
+}
+
+func NewIDLDefinedTypesCodec(idl IDL, builder encodings.Builder) (types.RemoteCodec, error) {
+	return newIdlCoded(idl, builder, idl.Types, false)
+}
+
+func newIdlCoded(
```
jmank88:
```suggestion
func newIDLCoded(
```

jmank88 (approved):

aalu1418 (approved):

PR #674, opened 2024-04-22
e2e: custom core ref from pr body

aalu1418:
core ref: 9efe7cab4c4d258073c889d49100529e2d8ce1dc


jmank88 (dismissed):

.github/workflows/e2e_custom_cl.yml
```diff
@@ -73,23 +73,34 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - name: Get core ref from PR body
```
tateexon, line 79: Should we move this out to its own job with an output and then use that output in both places instead of duplicated the shell script?

archseer (approved):

PR #688, opened 2024-04-30
feat: block history estimator

aalu1418:
https://smartcontract-it.atlassian.net/browse/BCI-2811

- [x] estimator
- [x] test cases
	- [x] unit
	- [x] integration (creating multiple runs to test both estimators - big changes are indentation changes)
- [x] configs (polling + bumping) 

upstream:
- https://github.com/smartcontractkit/chainlink-common/pull/517
- https://github.com/smartcontractkit/chainlink-solana/pull/709

downstream:
- https://github.com/smartcontractkit/chainlink/pull/13348


pkg/solana/fees/block_history.go
```diff
@@ -0,0 +1,137 @@
...
+
+	client *utils.LazyLoad[client.ReaderWriter]
+	cfg    config.Config
+	lgr    logger.Logger
+
+	price uint64
+	lock  sync.RWMutex
+}
+
+// NewBlockHistoryEstimator creates a new fee estimator that parses historical fees from a fetched block
```
augustbleeds, line 33:
```suggestion
// NewBlockHistoryEstimator creates a new fee estimator that parses historical fees from the latest block
```

pkg/solana/txm/txm.go
```diff
@@ -148,9 +148,12 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
 	// https://github.com/gagliardetto/solana-go/blob/main/transaction.go#L252
 	key := baseTx.Message.AccountKeys[0].String()
 
+	// only calculate base price once
```
augustbleeds, line 151: if bumping occurs after a new block is produced, wouldn't re-calculating the base be helpful?

augustbleeds (dismissed):
looks good!

pkg/solana/txm/txm.go
aalu1418: that's true - but there is also the possibility that the base value decreases and then the bumped value could be lower than the previous value.

only calculating the base once guarantees that the fee will always increase

augustbleeds (approved):

PR #697, opened 2024-05-07
chore: github action version bumps

erikburt:
## What

Updating Github Action references in all workflows.

## Why

Github Actions node16 deprecation. See [blog post (github.blog)](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/)
> Node16 has been out of support since [September 2023](https://github.com/nodejs/Release/#end-of-life-releases). As a result we have started the deprecation process of Node16 for GitHub Actions. We plan to migrate all actions to run on Node20 by Spring 2024.
> Following on from our [warning in workflows using Node16](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) we will start enforcing the use of Node20 rather than Node16 on the 13th of May.

## Notes

RE-2531

## Complications

There are complications with updating due to the usage of `projectserum/build` image which is based of an older version of ubuntu which doesn't support node20. This causes an error when you try and use `node20` actions within the container. I've adjusted this update to account for that, but it's likely going to cause issues on May 13 due to the above.

Outdated Dependency Paths:

```
codeql-analysis.yml -> analyze -> actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c -> node16
codeql-analysis.yml -> analyze -> actions/setup-go@v3 -> node16
codeql-analysis.yml -> analyze -> github/codeql-action/analyze@39d8d7e78f59cf6b40ac3b9fbebef0c753d7c9e5 -> node16
codeql-analysis.yml -> analyze -> github/codeql-action/autobuild@39d8d7e78f59cf6b40ac3b9fbebef0c753d7c9e5 -> node16
codeql-analysis.yml -> analyze -> github/codeql-action/init@39d8d7e78f59cf6b40ac3b9fbebef0c753d7c9e5 -> node16
contract-release.yml -> get_projectserum_version -> actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c -> node16
contract-release.yml -> release-artifacts -> actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c -> node16
contract-release.yml -> release-artifacts -> softprops/action-gh-release@v1 -> node16
e2e_custom_cl.yml -> check_test_compilation -> actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c -> node16
e2e_custom_cl.yml -> check_test_compilation -> actions/setup-go@v3 -> node16
[...]

PR #701, opened 2024-05-09
Use Docker Run Instead of GHA container image for node 16 shutdown

tateexon:
This temporarily gets around node 16 shutdown causing this to break starting 5/13/2024
If the update to the backpackapp/build is done before Monday then this can be closed and or reverted.


erikburt (approved):
Looks good to me - I'm going to try and pull this into #697 and test everything w/ node20 actions.

erikburt:
Closing in favour of #702

PR #702, opened 2024-05-10
Chore/update GitHub actions

erikburt:
This is a combination of #697 and #701. 

#697 - a general bump of github actions dependencies for the upcoming GHA node16 deprecation. See [blog post (github.blog)](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/).
    - This PR purposefully left out some updates due to a node20 incompatibility with projectserum v0.25.0, and how the workflows were setup.

#701 - @tateexon refactored the workflows to not run the jobs within the projectserum container, but instead ran the container with dispatched commands. This allowed us to use node20 actions in the job.


tateexon (dismissed):

.github/workflows/e2e_testnet_daily.yml
```diff
@@ -141,16 +142,17 @@ jobs:
         get_solana_sha,
       ]
     container:
```
momentmaker: should this be commented out too?

.github/workflows/rust.yml
```diff
@@ -27,73 +27,108 @@ jobs:
     runs-on: ubuntu-latest
     needs: [get_projectserum_version]
     container:
```
momentmaker: should this be commented out too?

.github/workflows/rust.yml
```diff
@@ -27,73 +27,108 @@ jobs:
...
+    #   run: |
+    #    cd examples/hello-world
+    #    yarn install --frozen-lockfile
+    #    anchor test
 
   rust_lint:
     name: Rust Lint
     runs-on: ubuntu-latest
     needs: [get_projectserum_version]
     container:
```
momentmaker: should this be commented out too?

erikburt: Good catch - commented them all out. 9c551f9

.github/workflows/rust.yml
erikburt: Good catch - commented them all out. 9c551f9

.github/workflows/e2e_testnet_daily.yml
erikburt: Good catch - commented them all out. 9c551f9

momentmaker (approved):

alejoberardino (approved):

PR #706, opened 2024-05-14
golangci-lint: update ruleset

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3203


Makefile
jmank88, line 1: We should drop in @HenryNguyen5 's script or similar here

aalu1418 (approved):

PR #709, opened 2024-05-17
cleanup: remove db config

aalu1418:
core ref: c00f33248fe41faafaf67315d3c2665ba0656488

downstream:
- https://github.com/smartcontractkit/chainlink/pull/13253


pkg/solana/config/toml.go
```diff
@@ -31,7 +26,7 @@ func (cs TOMLConfigs) validateKeys() (err error) {
 	chainIDs := config.UniqueStrings{}
 	for i, c := range cs {
 		if chainIDs.IsDupe(c.ChainID) {
-			err = errors.Join(err, config.NewErrDuplicate(fmt.Sprintf("%d.ChainID", i), *c.ChainID))
+			err = multierr.Append(err, config.NewErrDuplicate(fmt.Sprintf("%d.ChainID", i), *c.ChainID))
```
aalu1418: enforces usage of the `;` delimiter - `errors.Join` was using a `\n` delimiter

pkg/solana/config/toml.go
```diff
@@ -178,6 +162,21 @@ func setFromChain(c, f *solcfg.Chain) {
...
+	}
+	if f.ComputeUnitPriceMin != nil {
+		c.ComputeUnitPriceMin = f.ComputeUnitPriceMin
+	}
+	if f.ComputeUnitPriceDefault != nil {
+		c.ComputeUnitPriceDefault = f.ComputeUnitPriceDefault
+	}
+	if f.FeeBumpPeriod != nil {
+		c.FeeBumpPeriod = f.FeeBumpPeriod
+	}
```
aalu1418, line 180: adding missing configs

pkg/solana/config/toml.go
```diff
@@ -243,6 +242,9 @@ func (c *TOMLConfig) MaxRetries() *uint {
 	if c.Chain.MaxRetries == nil {
 		return nil
 	}
+	if *c.Chain.MaxRetries < 0 {
+		return nil // interpret negative numbers as nil (prevents unlikely case of overflow)
+	}
```
aalu1418, line 250: overflow handling

aalu1418 (commented):
overall goal was to remove the old DB configs and reduce duplicate configuration code

go.mod
```diff
@@ -17,11 +17,11 @@ require (
 	github.com/smartcontractkit/chainlink-common v0.1.7-0.20240516150131-e1be553a9d10
 	github.com/smartcontractkit/libocr v0.0.0-20240326191951-2bbe9382d052
 	github.com/stretchr/testify v1.9.0
+	go.uber.org/multierr v1.11.0
```
jmank88: The linter should be flagging this and telling you to use `errors.Join` from std lib instead.

Is the linter broken again?!

jmank88: It is running but giving a false positive :facepalm:

jmank88 (changes requested):
Fix linter

go.mod
aalu1418: i was trying to avoid using it but it was causing issues in the core CI where a config test started to fail:

```
--- FAIL: TestConfig_Validate (0.00s)
    --- FAIL: TestConfig_Validate/invalid (0.00s)
        config_test.go:1327: 
            	Error Trace:	/home/runner/work/chainlink/chainlink/core/services/chainlink/config_test.go:1540
            	            				/home/runner/work/chainlink/chainlink/core/services/chainlink/config_test.go:1327
            	            				
  <truncated>
  
  
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -73,5 +73,5 @@
            	            	 			- Nodes: missing: must have at least one node
            	            	-	- Solana: 5 errors:
            	            	+	- Solana: 4 errors:
            	            	 		- 1.ChainID: invalid value (mainnet): duplicate - must be unique
            	            	-		- 1.Nodes.1.Name: invalid value (bar): duplicate - must be unique
            	            	+	1.Nodes.1.Name: invalid value (bar): duplicate - must be unique
            	            	 		- 0.Nodes: missing: must have at least one node
   ```
   
   basically the error delimiter changed and it wasn't splitting properly 🤔: https://github.com/smartcontractkit/chainlink-solana/pull/709#discussion_r1605589134

Comment (jmank88) [reply go.mod]: Hmmm but this output is not based on the formatted errors, it is based on trying to apply a custom format. I think we are probably hard coded for the old multierr types instead of using the interface implemented by the std lib. Can we fix that in common/core?

Comment (aalu1418) [reply go.mod]: core just uses multierr.Errors to unwrap everything: https://github.com/smartcontractkit/chainlink/blob/0d58a8d5db24f42720226e73328e501637ba59c5/core/utils/errors.go#L17

I'm mostly confused why this is happening since errors.Join should provide `Unwrap() []error` which should break down things properly within multierr.Errors

File: pkg/solana/config/toml.go [hunk 5]
```diff
@@ -173,19 +163,36 @@ func setFromChain(c, f *solcfg.Chain) {
...
-		err = errors.Join(err, config.ErrEmpty{Name: "ChainID", Msg: "required for all chains"})
+		errA = append(errA, config.ErrEmpty{Name: "ChainID", Msg: "required for all chains"})
 	}
 
 	if len(c.Nodes) == 0 {
-		err = errors.Join(err, config.ErrMissing{Name: "Nodes", Msg: "must have at least one node"})
+		errA = append(errA, config.ErrMissing{Name: "Nodes", Msg: "must have at least one node"})
 	}
-	return
+	return errors.Join(errA...)
```
jmank88, line 195: Is this deferred join still necessary?

jmank88: It looks like the internal types will superficially be different, but if we are just using `Unwrap() []error`, then we still see the same slice: https://go.dev/play/p/65QFKJrpODU

aalu1418: yeah we are really just using `Unwrap() []error` but I think everything is passed around via the `error` interface? 

this seems like the simplest way to meet that interface without implementing a new type

jmank88: I mean that there should be no difference. Are you certain this is necessary?

jmank88: If you mean that we depend on _always_ having something that implements `Unwrap() []error`, then couldn't we can relax that assumption and do a conversion for the base cases of 0/1 on the other side?

aalu1418: update: this is the only way that works right now - doing a recursive unwrap within core does not solve the issue

jmank88 (approved):

pkg/solana/config/toml.go
aalu1418: revisiting in a future ticket: https://smartcontract-it.atlassian.net/browse/BCI-3330

PR #588, opened 2024-02-13
Dust off dependencies, update to anchor 0.29, solana 1.17

archseer:
https://smartcontract-it.atlassian.net/browse/BCI-2905

PR #711, opened 2024-05-20
fix: remove default pass on linters

aalu1418:
- [x] relayer
- [x] integration-tests

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -3,7 +3,7 @@
 pkgs.mkShell {
   nativeBuildInputs = with pkgs; [
     (rust-bin.stable.latest.default.override { extensions = ["rust-src"]; })
-    lld_10
+    # lld_10
     llvm_11
     stdenv.cc.cc.lib
     pkg-config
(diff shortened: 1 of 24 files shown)


shell.nix
```diff
@@ -3,7 +3,7 @@
 pkgs.mkShell {
   nativeBuildInputs = with pkgs; [
     (rust-bin.stable.latest.default.override { extensions = ["rust-src"]; })
-    lld_10
+    # lld_10
```
jmank88, line 6: What does this do?

jmank88 (approved):

shell.nix
aalu1418: my understanding it's a toolchain but @archseer also has it disabled in a PR that's updating even more dependencies https://github.com/smartcontractkit/chainlink-solana/pull/588/files#diff-e53dfbfffe62ae3c0b411b3938ccffa9fb6a2ecc565f55785ef8daa756631a6b

PR #803, opened 2024-07-23
cleanup: deduplicate shared cache logic

aalu1418:
### Description

deduplicate shared logic between the transmission + cache logic, potentially reusable for future work


pkg/solana/state_cache.go
```diff
@@ -2,139 +2,38 @@ package solana
...
-
-	timestamp := time.Now()
-	monitor.SetCacheTimestamp(timestamp, "ocr2_median_state", c.chainID, c.StateID.String())
-	// acquire lock and write to state
-	c.stateLock.Lock()
-	defer c.stateLock.Unlock()
-	c.state = state
-	c.stateTime = timestamp
-	return nil
+	return c.Read()
```
toblich: Stupid/nitpicky quick question 😅: is there a reason to keep this "alias" method around and not just replace it with explicit calls to `Read`? A quick lookup in the repo makes it seem like there are few places where they are called, so I guess it would be simple enough to remove it, or are we exposing this code elsewhere (e.g. as a lib) that could be making calls to this?

pkg/solana/client/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func (c *Cache[R]) Name() string {
+	return c.lggr.Name()
+}
+
+// Start polling
+func (c *Cache[R]) Start(ctx context.Context) error {
+	return c.StartOnce("cache", func() error {
```
toblich: Qq: Looking at the StateMachine code, particularly https://github.com/smartcontractkit/chainlink-common/blob/4ed45812784df8f781beab8748a5a1b504527990/pkg/services/state.go#L85, I see that the first argument is only used in logs to identify what happened. Would it thus make sense to append the `metricName` to the name here (e.g. `cache_ocr2_median_state` for the `StateCache` and `cache_ocr2_median_transmission` for the `TransmissionsCache`) so that logs better point to what component failed?

pkg/solana/client/cache.go
```diff
@@ -0,0 +1,143 @@
...
+			c.lggr.Warnf("error in initial fetch %s", err)
+		}
+		go c.Poll()
+		return nil
+	})
+}
+
+// Close stops the polling
+func (c *Cache[R]) Close() error {
+	return c.StopOnce("cache", func() error {
```
toblich: Same question regarding the first arg

pkg/solana/transmissions_cache.go
```diff
@@ -4,133 +4,31 @@ import (
...
-
-	timestamp := time.Now()
-	monitor.SetCacheTimestamp(timestamp, "ocr2_median_transmissions", c.chainID, c.TransmissionsID.String())
-	// acquire lock and write to state
-	c.ansLock.Lock()
-	defer c.ansLock.Unlock()
-	c.answer = answer
-	c.ansTime = timestamp
-	return nil
+	return c.Read()
```
toblich: Same question here 😄

pkg/solana/client/cache.go
```diff
@@ -0,0 +1,143 @@
...
+			if err != nil {
+				c.lggr.Errorf("error in Poll.fetch %s", err)
+			}
+			// Note negative duration will be immediately ready
+			tick = time.After(utils.WithJitter(c.cfg.OCR2CachePollPeriod()) - time.Since(start))
+		}
+	}
+}
+
+// ReadAnswer reads the latest state from memory with mutex and errors if timeout is exceeded
```
toblich:
```suggestion
// Read reads the latest state from memory with mutex and errors if timeout is exceeded
```

toblich (commented):
I lack a lot of context about the codebase, but this LGTM overall :+1:
Just a couple of small questions regarding tiny details

pkg/solana/client/cache.go
```diff
@@ -0,0 +1,143 @@
...
+type CacheGetter[R any] func(ctx context.Context) (res R, slot uint64, err error)
+
+// Cache is a generic implementation for caching data from the chain
+type Cache[R any] struct {
+	services.StateMachine
+
+	// identifier
+	metricName string
+	Account    solana.PublicKey
+	ChainID    string
```
jadepark-dev, line 28: It's not actually about these changes, but I’m curious about the choice of using a string for ChainID instead of an enum. I assume it’s for deserialization or improved visibility across different components. I would love to hear your thoughts on this!

pkg/solana/state_cache.go
aalu1418: good point! afaik this is not exposed further so standardizing the func name is probably fine. I'll give it a go

pkg/solana/client/cache.go
aalu1418: yes, great suggestion!

pkg/solana/client/cache.go
aalu1418: good question! it could be an enum but having something hard coded would require changes in the case a new solana-based network pops up

string is more useful than an int too because solana does not have chain ids like EVM networks so we use the genesis hash to determine which network we are running on

jadepark-dev: It makes total sense, thanks!

jadepark-dev (approved):
Looks good to me!

toblich (approved):
LGTM :+1:

PR #824, opened 2024-08-20
MultiNode Integration: Initial Setup

DylanTinianov:
### Description

Initial setup for MultiNode integration.

- Added common MultiNode files [`/multinode` folder will be extracted to a common repo later] 
- Modified local MutliNode metric naming to avoid duplicates
- Created outline for required configs
- Created StringID type for MultiNode integration
- Extended `chain` and `client` for MultiNode integration with feature flag

Closes https://smartcontract-it.atlassian.net/browse/BCI-4060

### Requires Dependencies
Chain Agnostic Multinode PR: https://github.com/smartcontractkit/chainlink/pull/13386
MultiNode updates will be copied into Solana until MultiNode code is extracted.


pkg/solana/chain.go
```diff
@@ -68,7 +70,7 @@ func NewChain(cfg *config.TOMLConfig, opts ChainOpts) (Chain, error) {
 	if !cfg.IsEnabled() {
 		return nil, fmt.Errorf("cannot create new chain with ID %s: chain is disabled", *cfg.ChainID)
 	}
-	c, err := newChain(*cfg.ChainID, cfg, opts.KeyStore, opts.Logger)
+	c, err := newChain(*cfg.ChainID, cfg, cfg.MultiNodeEnabled(), opts.KeyStore, opts.Logger)
```
aalu1418: why pass the `cfg` and then a value you can fetch from the config?

pkg/solana/chain.go
```diff
@@ -85,6 +87,11 @@ type chain struct {
 	balanceMonitor services.Service
 	lggr           logger.Logger
 
+	// if multiNode is enabled, the clientCache will not be used
+	multiNodeEnabled bool
```
aalu1418: why track the config here if it already exists in `cfg`?

pkg/solana/chain.go
```diff
@@ -186,13 +194,13 @@ func (v *verifiedCachedClient) LatestBlockhash() (*rpc.GetLatestBlockhashResult,
 	return v.ReaderWriter.LatestBlockhash()
 }
 
-func (v *verifiedCachedClient) ChainID() (string, error) {
+func (v *verifiedCachedClient) ChainID(ctx context.Context) (mn.StringID, error) {
```
aalu1418, line 197: is this to conform to a common interface?

if so, is it defined somewhere and can you add a constraint like `_ someInterface = (*someStruct)(nil)`

pkg/solana/chain.go
```diff
@@ -378,6 +446,13 @@ func (c *chain) Start(ctx context.Context) error {
 		c.lggr.Debug("Starting txm")
 		c.lggr.Debug("Starting balance monitor")
 		var ms services.MultiStart
+		if c.multiNodeEnabled {
+			c.lggr.Debug("Starting multinode")
+			err := ms.Start(ctx, c.multiNode, c.txSender)
```
aalu1418: can you call `MultiStart.Start()` twice?

if not, you might need to append it to an array of services instead

it might also be easier to read to use an array too - so the starting only happens in one place?

pkg/solana/chain.go
```diff
@@ -387,6 +462,13 @@ func (c *chain) Close() error {
 		c.lggr.Debug("Stopping")
 		c.lggr.Debug("Stopping txm")
 		c.lggr.Debug("Stopping balance monitor")
+		if c.multiNodeEnabled {
+			c.lggr.Debug("Stopping multinode")
+			err := services.CloseAll(c.multiNode, c.txSender)
```
aalu1418: same thing here?

pkg/solana/config/multinode.go
```diff
@@ -0,0 +1,87 @@
+package config
+
+import "time"
+
+type MultiNode struct {
+	// TODO: Determine current config overlap https://smartcontract-it.atlassian.net/browse/BCI-4065
```
aalu1418, line 6: while the config defaults are probably chain specific, the parameters themselves do not seem chain specific - ideally this could be extracted

pkg/solana/client/rpc_client.go
```diff
@@ -0,0 +1,312 @@
...
+	//TODO implement me
+	panic("implement me")
+}
+
+func (c *RPCClient) GetInterceptedChainInfo() (latest, highestUserObservations mn.ChainInfo) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func NewRPCClient(endpoint string, cfg config.Config, requestTimeout time.Duration, log logger.Logger) (*RPCClient, error) {
```
aalu1418: isn't this just a copy paste of the `client.go` code with additional methods for multinode?

why not do something like:

```go
type MultiNodeClient struct {
	Client
}

<add multinode specific methods>
```

pkg/solana/client/multinode/ctx.go
aalu1418, line 1: note: i did not review multinode specific logic - has there been a review of the extracted logic?

aalu1418 (commented):
looks like a good start!

pkg/solana/chain.go
DylanTinianov: Good point!

pkg/solana/client/rpc_client.go
DylanTinianov: We can actually just extend the existing client instead of needing this. Updated :)

pkg/solana/client/multinode/ctx.go
DylanTinianov: Yes the MultiNode logic has gone through multiple rounds of reviews, and will be merged soon. https://github.com/smartcontractkit/chainlink/pull/13386

pkg/solana/chain.go
DylanTinianov: Had to modify the `Reader` interface's `ChainID` method to accept context and return the stringID type in order to be used with MultiNode. The `verifiedCachedClient` is already defined as containing `ReaderWriter`, so the constraint is already explicit :)

pkg/solana/config/multinode.go
DylanTinianov: Sounds good, I'll create another PR to implement the default TOML configs for MultiNode. We can get that working first and then decide how we can extract things after.

pkg/solana/chain.go
DylanTinianov: Fixed.

pkg/solana/chain.go
```diff
@@ -221,6 +229,61 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+				lggr.Warnw("failed to create client", "name", *nodeInfo.Name, "solana-url", nodeInfo.URL.String(), "err", err.Error())
+				continue
+			}
+
+			newNode := mn.NewNode[mn.StringID, *client.Head, *client.Client](
+				mnCfg, mnCfg, lggr, *nodeInfo.URL.URL(), nil, *nodeInfo.Name,
+				int32(i), mn.StringID(id), 0, rpcClient, chainFamily)
+
+			nodes = append(nodes, newNode)
+		}
```
aalu1418: there's quite a bit of dereferencing happening in these lines - would any of these ever be `nil` and cause a panic?

pkg/solana/chain.go
```diff
@@ -221,6 +229,61 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+			[]mn.SendOnlyNode[mn.StringID, *client.Client]{},
+			mn.StringID(id),
+			chainFamily,
+			mnCfg.DeathDeclarationDelay(),
+		)
+
+		// TODO: implement error classification
+		classifySendError := func(tx *solanago.Transaction, err error) mn.SendTxReturnCode {
+			return 0 // TODO ClassifySendError(err, clientErrors, logger.Sugared(logger.Nop()), tx, common.Address{}, false)
+		}
```
aalu1418: future note: ideally we have this in a separate file since it probably could be fairly large

pkg/solana/client/multinode/ctx.go
aalu1418: do you think it's possible to have the core logic merged first to make sure there's no more changes or does it not matter?

i'm mostly wondering how stable the core changes are since they're not linked and just copied over right now

pkg/solana/chain.go
aalu1418: this might be useful reference: https://github.com/anza-xyz/agave/blob/master/sdk/src/transaction/error.rs

pkg/solana/chain.go
```diff
@@ -221,6 +229,61 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+		mnCfg := cfg.MultiNodeConfig()
+
+		var nodes []mn.Node[mn.StringID, *client.Client]
+
+		for i, nodeInfo := range cfg.ListNodes() {
+			// create client and check
+			rpcClient, err := client.NewClient(nodeInfo.URL.String(), cfg, DefaultRequestTimeout, logger.Named(lggr, "Client."+*nodeInfo.Name))
+			if err != nil {
+				lggr.Warnw("failed to create client", "name", *nodeInfo.Name, "solana-url", nodeInfo.URL.String(), "err", err.Error())
+				continue
```
dhaidashenko: Nit: returning an error instead of just logging might be a good idea. CL Node is not functional if we fail to create all clients, so it should not start

pkg/solana/chain.go
```diff
@@ -221,6 +229,61 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+				mnCfg, mnCfg, lggr, *nodeInfo.URL.URL(), nil, *nodeInfo.Name,
+				int32(i), mn.StringID(id), 0, rpcClient, chainFamily)
+
+			nodes = append(nodes, newNode)
+		}
+
+		multiNode := mn.NewMultiNode[mn.StringID, *client.Client](
+			lggr,
+			mn.NodeSelectionModeRoundRobin,
+			time.Minute, // TODO: set lease duration
```
dhaidashenko: Lease does not support RoundRobin. It might make sense to set lease duration to 0 to avoid confusion

pkg/solana/client/client.go
```diff
@@ -65,6 +68,25 @@ type Client struct {
...
+	}
+	return int64(*h.BlockHeight)
+}
+
+func (h *Head) BlockDifficulty() *big.Int {
+	return nil
+}
+
+func (h *Head) IsValid() bool {
+	return true
```
dhaidashenko: Should be false, if BlockHeight is nil

pkg/solana/chain.go
DylanTinianov: Added nil checks :)

pkg/solana/chain.go
DylanTinianov: Added this to the todo comments

pkg/solana/client/multinode/ctx.go
DylanTinianov: It shouldn't matter since we're expecting the possibility of changes being required for the Solana integration. I'll make sure any other changes are applied in the follow-up PRs as well. We will be extracting all the MultiNode code as soon as we get our repository set up! Ideally we make as much progress on Solana in parallel while EVM finishes soak testing.

pkg/solana/chain.go
```diff
@@ -238,11 +238,14 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
 		var nodes []mn.Node[mn.StringID, *client.Client]
 
 		for i, nodeInfo := range cfg.ListNodes() {
+			if nodeInfo == nil || nodeInfo.Name == nil || nodeInfo.URL == nil {
+				return nil, fmt.Errorf("node config contains nil: %+v", nodeInfo)
+			}
```
aalu1418: this might be useful here?
https://github.com/smartcontractkit/chainlink-solana/blob/a7050b5193cde253b9f649c2bd59b083a34636a3/pkg/solana/config/config.go#L153

aalu1418 (changes requested):

pkg/solana/chain.go
DylanTinianov: Nice, will add this to config validation instead!

pkg/solana/config/toml.go
```diff
@@ -191,6 +192,20 @@ func (c *TOMLConfig) ValidateConfig() (err error) {
...
+		} else if *node.Name == "" {
+			err = errors.Join(err, config.ErrEmpty{Name: "Name", Msg: "required for all nodes"})
+		}
+		if node.URL == nil {
+			err = errors.Join(err, config.ErrMissing{Name: "URL", Msg: "required for all nodes"})
+		} else if (*url.URL)(node.URL) == nil {
+			err = errors.Join(err, config.ErrEmpty{Name: "URL", Msg: "required for all nodes"})
+		}
+	}
+
```
aalu1418: can we deduplicate this code please? most if already exists here: https://github.com/smartcontractkit/chainlink-solana/blob/50ca9e4936f46f9085b6222ee6f6af602b292a85/pkg/solana/config/config.go#L153

pkg/solana/config/toml.go
```diff
@@ -112,6 +112,7 @@ type TOMLConfig struct {
 	// Do not access directly, use [IsEnabled]
 	Enabled *bool
 	Chain
+	MultiNode
 	Nodes Nodes
```
aalu1418, line 117: i think you're also missing handling multinode config in the `(c *TOMLConfig) SetFrom(f *TOMLConfig)` func?

pkg/solana/client/client.go
```diff
@@ -65,6 +68,26 @@ type Client struct {
...
+type Head struct {
+	rpc.GetBlockResult
+}
+
+func (h *Head) BlockNumber() int64 {
+	if h.BlockHeight == nil {
+		return 0
+	}
+	//nolint:gosec
+	return int64(*h.BlockHeight)
```
aalu1418: what's the actual lint error behind this? would be helpful to document why

pkg/solana/client/client.go
```diff
@@ -79,6 +102,56 @@ func NewClient(endpoint string, cfg config.Config, requestTimeout time.Duration,
...
+
+func (c *Client) SubscribeToHeads(ctx context.Context) (<-chan *Head, mn.Subscription, error) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (c *Client) SubscribeToFinalizedHeads(ctx context.Context) (<-chan *Head, mn.Subscription, error) {
+	//TODO implement me
+	panic("implement me")
+}
```
aalu1418, line 124: does this require WS support?

i thought we were slowly moving away from it because it's not reliable? we never implemented WS support because it had issues with memory leaks and instead went with a polling mechanism

pkg/solana/chain.go
```diff
@@ -221,6 +229,61 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+
+			nodes = append(nodes, newNode)
+		}
+
+		multiNode := mn.NewMultiNode[mn.StringID, *client.Client](
+			lggr,
+			mn.NodeSelectionModeRoundRobin,
+			0,
+			nodes,
+			[]mn.SendOnlyNode[mn.StringID, *client.Client]{},
```
aalu1418: is this just a placeholder or is the plan to always assume there are no SendOnlyNodes?

pkg/solana/client/client.go
```diff
@@ -65,6 +68,26 @@ type Client struct {
 	requestGroup *singleflight.Group
 }
 
+type Head struct {
+	rpc.GetBlockResult
+}
+
+func (h *Head) BlockNumber() int64 {
+	if h.BlockHeight == nil {
```
aalu1418: could just use `h.IsValid()` here

pkg/solana/client/client.go
aalu1418, line 81: random thought: why are we using a signed int as a block number and not a uint? can block numbers be negative?

aalu1418 (changes requested):

pkg/solana/config/toml.go
DylanTinianov: Yeah leaving the rest of the TOML configuration for another PR including the default values

pkg/solana/client/client.go
DylanTinianov: Yeah I was thinking about this too. Will have to ask Dmytro if there's a reason when he's back, although I've never heard of negative block numbers. I was looking into using unsigned int in the Chainlink repo, but I didn't want to increase the scope of my MultiNode PR.

pkg/solana/chain.go
DylanTinianov: Are send only RPCs relevant for Solana? I don't see them used anywhere currently, so I assumed there would be probably be none. This can change if we want to include them however.

aalu1418: i think it's relevant if we want to configure reading from a single node and just sending txs to many nodes?

pkg/solana/client/client.go
DylanTinianov: This does not require WS support and polling will be used for the implementation.

pkg/solana/chain.go
DylanTinianov: Added Send Only to the node config

pkg/solana/client/client.go
DylanTinianov: Added a comment to document this for now

pkg/solana/chain_test.go
```diff
@@ -174,7 +175,7 @@ func TestSolanaChain_VerifiedClient(t *testing.T) {
 	testChain.id = "incorrect"
 	c, err = testChain.verifiedClient(node)
 	assert.NoError(t, err)
-	_, err = c.ChainID()
+	_, err = c.ChainID(context.Background())
```
dhaidashenko: nit: use `tests.Context(t)`

pkg/solana/client/client.go
dhaidashenko: It seems to be a legacy from the initial iteration to create generalized components. I'm not sure why it was defined as int64. 
I agree with Dylan that we should keep this change out of scope for the MultiNode PR

dhaidashenko (dismissed):

dhaidashenko (dismissed):

aalu1418 (approved):

PR #829, opened 2024-08-22
[BCI-3989][solana] - CR methods err when service unstarted

Farber98:
### Task Description:
CR methods should throw an error if service is unstarted

### This PR:
- Adds `Close` method enforced in the `chainReaderInterfaceTester`. 

### Ticket:
- [BCI-3989](https://smartcontract-it.atlassian.net/browse/BCI-3989)

### Depends on:
- https://github.com/smartcontractkit/chainlink-common/pull/705

[BCI-3989]: https://smartcontract-it.atlassian.net/browse/BCI-3989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ilija42 (dismissed):

jmank88 (dismissed):

aalu1418 (changes requested):
code looks good to me! - but please merge the `chainlink-common` PR first and update the corresponding ref in `chainlink-solana`

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -520,7 +520,9 @@ func (r *chainReaderInterfaceTester) GetChainReader(t *testing.T) types.Contract
 
 	require.NoError(t, svc.Start(context.Background()))
 	t.Cleanup(func() {
-		require.NoError(t, svc.Close())
+		if svc.Ready() == nil {
+			require.NoError(t, svc.Close())
+		}
```
jmank88: These need to remain unchanged. It is never correct to skip `Close()` after calling `Start`, and this should be the only place that it happens.
```suggestion
			require.NoError(t, svc.Close())
```

Farber98: You are right! I'll `Start` again the service for the tests that `Close` it instead of modifying the `cleanup` func

[`e666621` (#829)](https://github.com/smartcontractkit/chainlink-solana/pull/829/commits/e666621734d165e7be89de66ec5d4986f67f7e5d)

Farber98: Right! Doing it [same way as core](https://github.com/smartcontractkit/chainlink/blob/v2.14.7-goreleaser-test23/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go#L69) so we don't skip `Close` but can execute it a second time without failing hard [`a0b44bd` (#829)](https://github.com/smartcontractkit/chainlink-solana/pull/829/commits/a0b44bd2b4d82d45f770d5ee728808d1cb513599)

jmank88: >  but can execute it a second time without failing hard

What do you mean by this? It is incorrect by definition to call `Close` a second time. This must be the only location that calls `Close`, and it must check the returned `error`.

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -621,6 +625,11 @@ func (r *wrappedTestChainReader) GetLatestValue(ctx context.Context, contractNam
 
 		fallthrough
 	default:
+		// If you called a method and the service is not started
+		if r.service.Ready() != nil {
+			return errors.New("service not ready")
```
jmank88: Let's include the error from `Ready()` instead of discarding it.

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -520,7 +520,7 @@ func (r *chainReaderInterfaceTester) GetChainReader(t *testing.T) types.Contract
 
 	require.NoError(t, svc.Start(context.Background()))
 	t.Cleanup(func() {
-		require.NoError(t, svc.Close())
+		_ = svc.Close()
```
jmank88: This check is important. Why are you removing it?
```suggestion
		require.NoError(t, svc.Close())
```

jmank88 (changes requested):

pkg/solana/chainreader/chain_reader_test.go
Farber98: I'll take `GetLatestValue` as an example for code references.

I'm trying to test that [calling](https://github.com/smartcontractkit/chainlink-common/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/types/interfacetests/chain_reader_interface_tests.go#L65) the `CR` methods returns an err when the service is not `Started`. To achieve this, I'm calling the [Close()](https://github.com/smartcontractkit/chainlink-common/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/types/interfacetests/chain_reader_interface_tests.go#L70) method that was added to the testing interface "early". The method called in Solana is [this one](https://github.com/smartcontractkit/chainlink-solana/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/solana/chainreader/chain_reader_test.go#L537).

In Solana [GetChainReader()](https://github.com/smartcontractkit/chainlink-solana/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/solana/chainreader/chain_reader_test.go#L512-L513) is calling `Start()` and the `cleanup()` is calling `Close()`. When we run this test, we [Close()](https://github.com/smartcontractkit/chainlink-common/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/types/interfacetests/chain_reader_interface_tests.go#L70) the service for the first time and then when executing the `cleanup()`, we will get an error because we are trying to `Close()` again. That's why I removed the `require.NoError()` to do it [the same way](https://github.com/smartcontractkit/chainlink/blob/v2.14.7-goreleaser-test23/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go#L69) as it is in Core. By doing this, we are sure that the service was closed but we don't fail with the second `Close()` attempt.

The other thing I could do is remove the `Start()` and `Close()` calls inside the `GetChainReader()` in [Solana](https://github.com/smartcontractkit/chainlink-solana/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/solana/chainreader/chain_reader_test.go#L512-L513) and [Core](https://github.com/smartcontractkit/chainlink/blob/BCI-3989-cr-methods-error-when-unstarted/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go#L201) and execute them as needed in every test. It will introduce a lot of redundant calls to `Start()` and `Close()` but we will be able to control when we want to call these funcs avoiding double `Start/Close` problems.

Let me know your thoughts and if you have a better approach to achieve this

pkg/solana/chainreader/chain_reader_test.go
Farber98: Answered here: https://github.com/smartcontractkit/chainlink-solana/pull/829#discussion_r1742553823

pkg/solana/chainreader/chain_reader_test.go
jmank88: > That's why I removed the require.NoError() to do it [the same way](https://github.com/smartcontractkit/chainlink/blob/v2.14.7-goreleaser-test23/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go?rgh-link-date=2024-09-03T19%3A16%3A20Z#L69) as it is in Core.

This existing code in incorrect. This is a bug.

> By doing this, we are sure that the service was closed but we don't fail with the second Close() attempt.

Suppressing the error without checking it does not guarantee this. Other errors are possible than a double close error - but that still should not be ignored regardless. It is incorrect by definition to call `Close` a second time, and it is never necessary to do so.

> The other thing I could do is remove the Start() and Close() calls inside the GetChainReader() in [Solana](https://github.com/smartcontractkit/chainlink-solana/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/solana/chainreader/chain_reader_test.go?rgh-link-date=2024-09-03T19%3A16%3A20Z#L512-L513) and [Core](https://github.com/smartcontractkit/chainlink/blob/BCI-3989-cr-methods-error-when-unstarted/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go?rgh-link-date=2024-09-03T19%3A16%3A20Z#L201) and execute them as needed in every test. It will introduce a lot of redundant calls to Start() and Close() but we will be able to control when we want to call these funcs avoiding double Start/Close problems.

I don't understand what you mean by this. Can you elaborate?

If the goal is to evaluate errors returned when not started, then all that is necessary is to make method calls either before starting or after closing. Is something preventing that from being done in a straightforward way?

ilija42: The core code should be fixed, there is an error when` Close()` is called because the db instance gets closed before CR. This is happening because`GetChainReader()` calls into `NewSqlxDB()` which calls into `t.Cleanup()` before `Setup()` is called and calls `t.Cleanup()` which means that sqlx db gets cleaned up first. 
We shouldn't ignore this error, but rather fix the cause

Farber98: Thanks for the clarification! 

> The other thing I could do is remove the Start() and Close() calls inside the GetChainReader() in [Solana](https://github.com/smartcontractkit/chainlink-solana/blob/BCI-3989-cr-methods-error-when-unstarted/pkg/solana/chainreader/chain_reader_test.go?rgh-link-date=2024-09-03T20%3A33%3A48Z#L512-L513) and [Core](https://github.com/smartcontractkit/chainlink/blob/BCI-3989-cr-methods-error-when-unstarted/core/services/relay/evm/evmtesting/chain_reader_interface_tester.go?rgh-link-date=2024-09-03T20%3A33%3A48Z#L201) and execute them as needed in every test. It will introduce a lot of redundant calls to Start() and Close() but we will be able to control when we want to call these funcs avoiding double Start/Close problems.

I was thinking on exposing the `Start()` and `Close()` methods in the `ChainReaderInterfaceTester` and be able to call them from the tests in the moments we want rather than receiving it already started by the `Setup()` and having a mandatory `Close()` call at the end executed by the `cleanup()`. In this way we could call `Start()` and `Close()` when we need to and defer the `Close()` call if we want it to act as a `cleanup()`

Farber98: Pushed these changes in the 3 repos. CI will fail until https://github.com/smartcontractkit/chainlink/pull/14252 is merged and I rebase `core`. Tests were passing locally.

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -621,6 +624,11 @@ func (r *wrappedTestChainReader) GetLatestValue(ctx context.Context, contractNam
 
 		fallthrough
 	default:
+		// If you called a method and the service is not started
+		if err := r.service.Ready(); err != nil {
+			return fmt.Errorf("service not ready. err: %w", err)
+		}
+
```
ilija42: Shouldn't this check be before the switch statement?

Farber98: Good catch! Fixed

PR #844, opened 2024-09-06
MultiNode Integration: TOML Configurations

DylanTinianov:
### Description
Added required TOML configurations for MultiNode integration.

Closes: https://smartcontract-it.atlassian.net/browse/BCFR-302

Core Bump Solana + Fix Tests PR: https://github.com/smartcontractkit/chainlink/pull/14575

### Requires Dependencies
This PR is a follow up to the MultiNode Initial Setup PR: https://github.com/smartcontractkit/chainlink-solana/pull/824


pkg/solana/config/multinode.go
```diff
@@ -1,87 +1,163 @@
...
+
+	c.nodeIsSyncingEnabled = ptr(false)
+	c.finalizedBlockPollInterval = config.MustNewDuration(5 * time.Second)
+	c.enforceRepeatableRead = ptr(true)
+	c.deathDeclarationDelay = config.MustNewDuration(10 * time.Second)
+
+	// Chain Configs
+	c.nodeNoNewHeadsThreshold = config.MustNewDuration(10 * time.Second)
+	c.noNewFinalizedHeadsThreshold = config.MustNewDuration(10 * time.Second)
+	c.finalityDepth = ptr(uint32(0))
```
patrick-dowell: I'm not sure the helper function is adding much here. Wouldn't this be cleaner?

```
c.finalityDepth = &uint32(0)
c.finalityTagEnabled = &true
c.finalizedBlockOffset = &uint32(0)
```

pkg/solana/config/multinode.go
```diff
@@ -1,87 +1,163 @@
...
-}
+func (c *MultiNode) FinalizedBlockOffset() uint32 { return *c.finalizedBlockOffset }
 
 func (c *MultiNode) SetDefaults() {
-	// TODO: Set defaults for MultiNode config https://smartcontract-it.atlassian.net/browse/BCI-4065
-	c.multiNodeEnabled = false
+	c.multiNodeEnabled = ptr(false)
+
+	// Node Configs
+	c.pollFailureThreshold = ptr(uint32(5))
```
patrick-dowell: The default values here - where are they coming from? For example, the 5 for syncThreshold or the 10 for pollInterval. Could these come from another source or if they are to be hardcoded, could we provide some documentation as to why these are the best default values for these fields?

pkg/solana/config/multinode.go
```diff
@@ -1,87 +1,163 @@
...
+	c.finalizedBlockOffset = ptr(uint32(0))
+}
+
+func (c *MultiNode) SetFrom(fs *MultiNode) {
+	if fs.multiNodeEnabled != nil {
+		c.multiNodeEnabled = fs.multiNodeEnabled
+	}
+
+	// Node Configs
+	if fs.pollFailureThreshold != nil {
```
patrick-dowell: Here you could use a helper function or reflection to clean this up. I like the reflection way of doing stuff like this unless you want to manually exclude some of the fields.

patrick-dowell (commented):
Some minor comments around cleaning up this code.

aalu1418 (changes requested):
usually changing TOML configs requires updating tests in core with default values

please link a draft PR in core that bumps `chainlink-solana` in core and validates that all tests are passing

pkg/solana/config/multinode.go
DylanTinianov: That syntax actually doesn't work in Go! We would have to do the following instead:

```
c.finalityDepth = new(uint32)
c.finalityDepth = 0

c.finalityTagEnabled = new(bool)
*c.finalityTagEnabled = true
...
```

This ends up adding a lot of lines. In the case where our default value is not 0 or false we would have to add a second line to set the value after the `new` pointer initialization. If you think this approach is still better then I don't mind changing it.

pkg/solana/config/multinode.go
DylanTinianov: I had a meeting with Dmytro and Aaron where we discussed what default values make sense for each config. I can for sure add some comments though!

pkg/solana/config/multinode.go
patrick-dowell: oh! If it doesn't work in Go, then I'd just keep the helper function. Weird that the syntax doesn't work though!

dhaidashenko (changes requested):
Current implementation will override with default any value set in the file.
We should you `if ... != nil` approach in  `SetDefaults` method

pkg/solana/chain.go
```diff
@@ -230,7 +230,7 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
 		clientCache: map[string]*verifiedCachedClient{},
 	}
 
-	if cfg.MultiNodeEnabled() {
+	if cfg.MultiNodeConfig().Enabled() {
```
aalu1418: can we standardize how this is being used?

there's `cfg.MultiNodeConfig().Enabled()` and `cfg.MultiNode.Enabled()` both being used in here

pkg/solana/chain.go
```diff
@@ -258,8 +258,8 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
 
 		multiNode := mn.NewMultiNode[mn.StringID, *client.Client](
 			lggr,
-			mn.NodeSelectionModeRoundRobin,
-			0,
+			mnCfg.SelectionMode(),
+			mnCfg.LeaseDuration(),
```
aalu1418, line 262: nit: if more and more configs need to be passed in to `NewMultiNode`, it would be cleaner to just pass `mnCfg` and use a minimal interface underneath as the input type

that way, you don't need to change the function signature every time you need to add a new config

pkg/solana/config/multinode.go
```diff
@@ -1,87 +1,206 @@
...
+	mn "github.com/smartcontractkit/chainlink-solana/pkg/solana/client/multinode"
+)
+
+// MultiNodeConfig is a wrapper to provide required functions while keeping configs Public
+type MultiNodeConfig struct {
+	MultiNode
+}
 
 type MultiNode struct {
-	// TODO: Determine current config overlap https://smartcontract-it.atlassian.net/browse/BCI-4065
```
aalu1418, line 6: what was the result of checking for shareable config?

pkg/solana/config/toml.go
```diff
@@ -285,7 +286,7 @@ func (c *TOMLConfig) ListNodes() Nodes {
 	return c.Nodes
 }
 
-func (c *TOMLConfig) MultiNodeConfig() *MultiNode {
+func (c *TOMLConfig) MultiNodeConfig() *MultiNodeConfig {
 	return &c.MultiNode
 }
```
aalu1418: is this still needed if we just use `c.MultiNode`?

aalu1418:
can you also link the core PR in PR description? it was buried in the test/deployment events

pkg/solana/config/multinode.go
DylanTinianov: Did not end up using any shared configs.

pkg/solana/config/toml.go
DylanTinianov: This wrapper is needed in order to allow the config names to be public, while also having the required functions names which are the same.

pkg/solana/chain.go
DylanTinianov: Yeah this would require changing the MultiNode, which is out of scope for now. Will certainly make that change if the constructor needs to change again.

pkg/solana/chain.go
DylanTinianov: Good catch, we actually don't even need `MultiNodeConfig()`. Updated to remove this function and use cfg.MultiNode.Enabled()

pkg/solana/config/multinode.go
dhaidashenko: It would be nice to reuse the config across different chains. The EVM case might be a bit complex, but for new integrations like Aptos/Tron, it does not make sense to duplicate this code.
Could you create a follow up ticket to address that?

dhaidashenko (approved):

aalu1418 (approved):

pkg/solana/config/multinode.go
DylanTinianov: Yeah sounds good, we can extract this to the framework repo when doing the extraction.

PR #845, opened 2024-09-06
Implement RPC Client Methods

DylanTinianov:
### Description
Implemented the RPCClient interface as part of the Solana Client to enable the use of MultiNode.
Ticket: https://smartcontract-it.atlassian.net/browse/BCFR-298


pkg/solana/chain.go
```diff
@@ -292,11 +292,11 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
 	tc := func() (client.ReaderWriter, error) {
 		return ch.getClient()
 	}
-	ch.txm = txm.NewTxm(ch.id, tc, cfg, ks, lggr)
+	ch.txm = txm.NewTxm(ch.id, tc, cfg, ch.multiNode, ks, lggr)
```
dhaidashenko: Why do you need to pass multiNode into txm, if you are already injected it with `ch.getClient()`

pkg/solana/chain.go
```diff
@@ -292,11 +292,11 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
 	bc := func() (monitor.BalanceClient, error) {
 		return ch.getClient()
 	}
-	ch.balanceMonitor = monitor.NewBalanceMonitor(ch.id, cfg, lggr, ks, bc)
+	ch.balanceMonitor = monitor.NewBalanceMonitor(ch.id, cfg, cfg.MultiNode.Enabled(), lggr, ks, bc)
```
dhaidashenko: Same here

pkg/solana/chain_test.go
```diff
@@ -152,9 +210,12 @@ func TestSolanaChain_VerifiedClient(t *testing.T) {
 
 	ch := solcfg.Chain{}
 	ch.SetDefaults()
+	mn := solcfg.MultiNodeConfig{}
```
dhaidashenko: Do we need multiNode in this test? 
It seems to be focused `verifiedClient()`  that does not interact with MultiNode

pkg/solana/chain_test.go
```diff
@@ -199,10 +260,13 @@ func TestSolanaChain_VerifiedClient_ParallelClients(t *testing.T) {
...
```
dhaidashenko: same here

pkg/solana/client/client.go
```diff
@@ -66,10 +67,45 @@ type Client struct {
 
 	// provides a duplicate function call suppression mechanism
 	requestGroup *singleflight.Group
+
+	// MultiNode
+	pollInterval               time.Duration
```
dhaidashenko: We should not modify existing clients. MultiNode feature flag should guarantee that if it's disabled we'll run pre-MultiNode code

pkg/solana/client/client.go
```diff
@@ -66,10 +67,45 @@ type Client struct {
...
+	finalizedBlockPollInterval time.Duration
+	stateMu                    sync.RWMutex // protects state* fields
+	subs                       map[mn.Subscription]struct{}
```
dhaidashenko: need to have a separate lock for subs. This will help us avoid deadlocks

pkg/solana/client/client.go
```diff
@@ -82,75 +118,207 @@ func (h *Head) BlockNumber() int64 {
...
 }
 
 func (c *Client) SubscribeToHeads(ctx context.Context) (<-chan *Head, mn.Subscription, error) {
-	//TODO implement me
-	panic("implement me")
+	if c.pollInterval == 0 {
+		return nil, nil, errors.New("PollInterval is 0")
+	}
+	timeout := c.pollInterval
+	poller, channel := mn.NewPoller[*Head](c.pollInterval, c.LatestBlock, timeout, c.log)
```
dhaidashenko: poller should be registered as subscription. Otherwise Node's state transition won't cause it and won't force application layer to reconnect

pkg/solana/client/client.go
```diff
@@ -82,75 +118,207 @@ func (h *Head) BlockNumber() int64 {
...
 }
 
 func (c *Client) SubscribeToFinalizedHeads(ctx context.Context) (<-chan *Head, mn.Subscription, error) {
-	//TODO implement me
-	panic("implement me")
+	if c.finalizedBlockPollInterval == 0 {
+		return nil, nil, errors.New("FinalizedBlockPollInterval is 0")
+	}
+	timeout := c.finalizedBlockPollInterval
+	poller, channel := mn.NewPoller[*Head](c.finalizedBlockPollInterval, c.LatestFinalizedBlock, timeout, c.log)
```
dhaidashenko: same here

pkg/solana/client/client.go
```diff
@@ -82,75 +118,207 @@ func (h *Head) BlockNumber() int64 {
...
+func (c *Client) onNewHead(ctx context.Context, requestCh <-chan struct{}, head *Head) {
+	if head == nil {
+		return
+	}
+
+	c.chainInfoLock.Lock()
+	defer c.chainInfoLock.Unlock()
+	if !mn.CtxIsHeathCheckRequest(ctx) {
+		c.highestUserObservations.BlockNumber = max(c.highestUserObservations.BlockNumber, head.BlockNumber())
+		c.highestUserObservations.TotalDifficulty = mn.MaxTotalDifficulty(c.highestUserObservations.TotalDifficulty, head.BlockDifficulty())
```
dhaidashenko: total difficulty is not applicable. We can remove it from here

dhaidashenko (changes requested):

pkg/solana/chain.go
DylanTinianov: Fair point, we don't need to pass in the multiNode but we still need to know if it's enabled. We can pass in a TOMLConfig which contains the multinode config and use that to check instead. 

If MultiNode is enabled, the TXM will use the passed in `tc` function directly, instead of only using it with the lazy loader (which would negate the proper use of MultiNode).

pkg/solana/chain.go
DylanTinianov: Similarly here I'll change `cfg` to be the full TOMLConfig which includes whether MultiNode is enabled.

pkg/solana/chain_test.go
DylanTinianov: We don't activate MultiNode here, just need it to be defined.  I will use `cfg.SetDefaults()` instead!

pkg/solana/client/client.go
DylanTinianov: Created `MultiNodeClient` which wraps `Client` with the additional requirements for MultiNode. It will only be used only when the feature flag is enabled.

pkg/solana/client/multinode_client.go
```diff
@@ -0,0 +1,306 @@
...
+		sub.Unsubscribe()
+		return fmt.Errorf("failed to register subscription - all in-flight requests were canceled")
+	default:
+	}
+	// TODO: BCI-3358 - delete sub when caller unsubscribes.
+	m.subs[sub] = struct{}{}
+	return nil
+}
+
+func (m *MultiNodeClient) Dial(ctx context.Context) error {
```
patrick-dowell, line 96: You should put a comment here as it's weird to have a function that just returns nil. Is this a TODO to implement later? Is this functionality not needed for Solana?

DylanTinianov: It's not needed for Solana, I added a comment :)

pkg/solana/client/multinode_client.go
```diff
@@ -0,0 +1,307 @@
...
+	return int64(*h.BlockHeight)
+}
+
+func (h *Head) BlockDifficulty() *big.Int {
+	// Not relevant for Solana
+	return nil
+}
+
+func (h *Head) IsValid() bool {
+	return h.BlockHeight != nil && h.BlockHash != nil
```
dhaidashenko:
```suggestion
	return h != nil && h.BlockHeight != nil && h.BlockHash != nil
```

dhaidashenko (dismissed):

dhaidashenko (approved):

pkg/solana/client/multinode_client.go
```diff
@@ -0,0 +1,307 @@
...
+	poller, channel := mn.NewPoller[*Head](finalizedBlockPollInterval, m.LatestFinalizedBlock, timeout, m.log)
+	if err := poller.Start(ctx); err != nil {
+		return nil, nil, err
+	}
+
+	err := m.registerSub(&poller, chStopInFlight)
+	if err != nil {
+		poller.Unsubscribe()
+		return nil, nil, err
+	}
```
aalu1418, line 142: does this poller capture every new block on solana? or does it just poll intermittently and fetch the latest every X seconds?

we will likely run into performance issues if we are capturing _every_ new block

aalu1418 (commented):
code looks good to me! just a question on how the underlying block ingestion works

pkg/solana/client/multinode_client.go
DylanTinianov: It doesn't capture every block, just every X seconds which we can specify in the TOML configs.

aalu1418 (approved):

aalu1418 (approved):

PR #849, opened 2024-09-10
[BCF-3143] Multiple Address Bindings

EasterTheBunny:
### Description

Support for multiple address bindings in `ContractReader`.


pkg/solana/chainreader/chain_reader.go
```diff
@@ -187,17 +205,36 @@ func (s *SolanaChainReaderService) QueryKey(_ context.Context, _ types.BoundCont
...
-	return s.bindings.Bind(bindings)
+	for _, binding := range bindings {
+		if err := s.bindings.Bind(binding); err != nil {
+			return err
+		}
+
+		s.lookup.bindAddressForContract(binding.Name, binding.Address)
+	}
+
+	return nil
```
ilija42, line 216: should Solana also have something like Binding registry to handle this in the same place>

pkg/solana/chainreader/chain_reader.go
```diff
@@ -145,11 +163,11 @@ func (s *SolanaChainReaderService) GetLatestValue(ctx context.Context, readIdent
 			}
 
 			wg.Add(1)
-			go func(ctx context.Context, rb readBinding, res *loadedResult) {
+			go func(ctx context.Context, rb readBinding, res *loadedResult, address string) {
 				defer wg.Done()
 
-				rb.PreLoad(ctx, res)
-			}(localCtx, binding, results[idx])
+				rb.PreLoad(ctx, address, res)
```
ilija42, line 169: Is Preload similar to evm contract read? Why not call ReadAsync?

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: I think eventually, yes, but most of this will change when log events get rolled in.

ilija42 (approved):

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: For the reader, `PreLoad` does do an async read, but it stores the value in the reader and must be followed by `GetLatestValue`. It allows multiple readers to be read async, but in order.

aalu1418 (approved):

PR #852, opened 2024-09-12
update: use anza-xyz agave

aalu1418:
### Description

solanalabs/solana will be deprecated with the 2.0.X release - migrating to anza-xyz/agave before to be ready for incoming 2.0.X releases


jadepark-dev (approved):
Looks good to me!

PR #857, opened 2024-09-13
som: handle pruned transactions for slower updating feeds

aalu1418:
### Description

some solana RPC nodes will prune old transactions. if there are feeds that update slowly, transactions may be pruned from the rpc node and SOM would be unable to find a previous tx and throw an error. 

that error blocks the data packed in the envelope to be used as an update. if this occurs during startup, it would cause price data to be missing until SOM sees a transaction.

this PR adds a cached value in the case that transactions are pruned. on startup with pruned transactions, it would return 0 rather than block the entire metrics update

### DIFF
--- a/pkg/monitoring/source_envelope.go
+++ b/pkg/monitoring/source_envelope.go
@@ -74,7 +74,7 @@ func (s *envelopeSource) setJuelsPerLamport(v uint64) error {
 
 func (s *envelopeSource) readJuelsPerLamport() uint64 {
 	s.lock.RLock()
-	s.lock.RUnlock()
+	defer s.lock.RUnlock()
 	return s.cacheJuelsPerLamport
 }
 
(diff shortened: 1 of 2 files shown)


pkg/monitoring/source_envelope.go
```diff
@@ -57,6 +57,25 @@ type envelopeSource struct {
...
+	s.lock.Lock()
+	s.cacheJuelsPerLamport = v
+	s.lock.Unlock()
+
+	return nil
+}
+
+func (s *envelopeSource) readJuelsPerLamport() uint64 {
+	s.lock.RLock()
+	s.lock.RUnlock()
```
jmank88:
```suggestion
	defer s.lock.RUnlock()
```

aalu1418: 🤦🏼

jmank88 (dismissed):

jmank88 (approved):

PR #864, opened 2024-09-23
add dependabot for CTF

skudasov:


jmank88 (approved):

aalu1418:
is dependabot robust enough to handle the bumps?

often times I need to bump the core dep at the same time otherwise the test won't compile

that's why I have this workflow that automates it: https://github.com/smartcontractkit/chainlink-solana/blob/6ad738f78dd5dfacbdf986237cdd0fd9a54bb0a4/.github/workflows/dependency-updates.yml#L22

i guess the main question is - do we still need this?

PR #875, opened 2024-10-01
[BCFR-147][solana] - Add codec chain agnostic modifier for converting byte array address to string

Farber98:
### **Description**
This PR focuses on adding the `AddressBytesToStringModifierConfig` inside chain reader service tests.

### Resources
- [BCFR-147](https://smartcontract-it.atlassian.net/browse/BCFR-147)

### Related PRs
-  https://github.com/smartcontractkit/chainlink-common/pull/818
-  https://github.com/smartcontractkit/chainlink/pull/14620

[BCFR-147]: https://smartcontract-it.atlassian.net/browse/BCFR-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -652,6 +662,11 @@ func (r *wrappedTestChainReader) GetLatestValue(ctx context.Context, readIdentif
 
 		fallthrough
 	default:
+		// Isolate test until we have InputModifications
+		if strings.Contains(r.test.Name(), "wraps_config_with_modifiers_using_its_own_mapstructure_overrides") {
```
ilija42: why are we skipping this whole test case, wdym by until we have InputModifications?

ilija42 (commented):
Looks good, I just don't understand the skipped test case

pkg/solana/codec/byte_string_modifier.go
```diff
@@ -0,0 +1,42 @@
...
+func (s SolanaAddressModifier) EncodeAddress(bytes []byte) (string, error) {
+	if len(bytes) != s.Length() {
+		return "", fmt.Errorf("%w: got length %d, expected 32 for bytes %x", commontypes.ErrInvalidType, len(bytes), bytes)
+	}
+	return solana.PublicKeyFromBytes(bytes).String(), nil
+}
+
+// DecodeAddress decodes a Base58-encoded Solana address into a 32-byte array.
+func (s SolanaAddressModifier) DecodeAddress(str string) ([]byte, error) {
+	// Validate the length of the address
```
ilija42: comment the why, not the what, a lot of the comments in this file can be very easily derived from the code

pkg/solana/chainreader/chain_reader_test.go
Farber98: Was the only way I found to `isolate / skip` this failing test that we talked

I think the same happens in `core` but it is handled with InputModifications:
- [common test](https://github.com/smartcontractkit/chainlink-common/blob/167715aa8613ebdaa17f1b5b4b003a31c120d455/pkg/types/interfacetests/chain_components_interface_tests.go#L272-L273)
- [core config](https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/evmtesting/chain_components_interface_tester.go#L185-L191)

I did the same with this new modifier to make that test pass in core:
- [input modification address byte string modifier](https://github.com/smartcontractkit/chainlink/blob/8830519a04ffae6923f5c63130d97fde4a03dd5a/core/services/relay/evm/evmtesting/chain_components_interface_tester.go#L209-L212)

Didn't find a way to achieve the same here, let me know if you think of a better alternative

ilija42: We shouldn't break unrelated tests like this. Can the struct be manually "retyped" with an added TODO when this test case is encountered instead of using an input modifier

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -991,6 +1041,11 @@ func (s *skipEventsChainReader) GetLatestValue(ctx context.Context, readIdentifi
 		s.t.Skip("Events are not yet supported in Solana")
 	}
 
+	// Isolate test until we have InputModifications
+	if strings.Contains(s.t.Name(), "wraps_config_with_modifiers_using_its_own_mapstructure_overrides") {
+		s.t.Skip("Isolating test until we have InputModifications")
+	}
+
```
ilija42: The test is still being skipped, just delete this Skip and everything should work
```suggestion
```

ilija42 (changes requested):

ilija42 (dismissed):

silaslenihan (dismissed):

pkg/solana/chainreader/chain_reader.go
```diff
@@ -311,6 +314,16 @@ func (s *SolanaChainReaderService) init(namespaces map[string]config.ChainReader
 	return nil
 }
 
+// injectSolanaSpecificCodecModifiers injects the AddressModifier into OutputModifications for Solana procedures.
+func injectSolanaSpecificCodecModifiers(procedure *config.ChainReaderProcedure) {
```
jmank88: Will this be used from anywhere else? IMHO it would read easier inline. Otherwise, this func could accept just the `OutputModifications` instead of a reference to the enclosing `ChainReaderProcedure`

EasterTheBunny: Well the init function is already 3 deep on nested for loops. This just adds another. Layering like that gets hard to read inline too. Kind of a style issue. Linters would normalize this though.

EasterTheBunny (dismissed):

pkg/solana/chainreader/chain_reader.go
jmank88: Agree with limiting nesting, but this is not the only way. IMHO outside-in would yield cleaner results. How about using a helper for everything after:
```go
for namespace, methods := range namespaces {
		for methodName, method := range methods.Methods {
```

jmank88: If not, then please simplify this func to accept only what is necessary, and find a clearer name

ilija42 (dismissed):

silaslenihan (dismissed):

pkg/solana/chainreader/chain_reader_test.go
```diff
@@ -666,9 +679,44 @@ func (r *wrappedTestChainReader) GetLatestValue(ctx context.Context, readIdentif
 		// split into two encoded parts to test the preloading function
 		cdc := makeTestCodec(r.test, fullStructIDL(r.test), config.EncodingTypeBorsh)
 
-		bts, err := cdc.Encode(ctx, nextTestStruct, "TestStructB")
-		if err != nil {
-			r.test.FailNow()
+		var bts []byte
+		var err error
+		if strings.Contains(r.test.Name(), "wraps_config_with_modifiers_using_its_own_mapstructure_overrides") {
+			// TODO: This is a temporary solution. We are manually retyping this struct to avoid breaking unrelated tests.
```
jmank88, line 685: Do we have a ticket to link?

jmank88 (dismissed):

aalu1418 (dismissed):
lgtm! i will enable auto-merge once the common PR is merged and ref updated

ilija42 (approved):

aalu1418 (approved):

PR #879, opened 2024-10-04
MultiNode Transaction Sender

DylanTinianov:
### Description
Ticket: https://smartcontract-it.atlassian.net/browse/BCFR-300

Integrated the use of MultiNode TransactionSender. 
- Added a `sendTx()` override to txm which is used to call TransactionSender instead of a single rpc client.
- Added Error Classification
- Created Loader component to swap out lazy loading with direct calls to MultiNode


pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -28,31 +28,32 @@ var (
 // (e.g. Successful, Fatal, Retryable, etc.)
 type TxErrorClassifier[TX any] func(tx TX, err error) SendTxReturnCode
 
-type sendTxResult struct {
+type sendTxResult[RESULT any] struct {
 	Err        error
 	ResultCode SendTxReturnCode
+	Result     *RESULT
```
jmank88: Typically generics are single letters, unless they need to be longer to distinguish from one another. In this case it is easy to see the only usage is the field named `Result`:
```suggestion
type sendTxResult[R any] struct {
	Err        error
	ResultCode SendTxReturnCode
	Result     *R
```

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -63,7 +64,7 @@ func NewTransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]](
 	}
 }
 
-type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
+type TransactionSender[TX any, RESULT any, CHAIN_ID ID, RPC SendTxRPCClient[TX, RESULT]] struct {
```
jmank88: IMHO these names are too verbose. Single letters could be used without conflicts:
```suggestion
type TransactionSender[T any, R any, I ID, C SendTxRPCClient[T, R]] struct {
```
I realize that some of those are not intuitive as-is, but that's what docs are for :shrug:

patrick-dowell: Isn't CHAIN_ID something we use elsewhere, like in the TX? I understand wanting to use the abbreviations but we should be consistent about the naming.

jmank88: Well single var everywhere would be great, but each case is different. Enforcing wide consistency actually induces this problem because suddenly we have to operate in a global name space. These generic types are just meant to be shorthand to reference from elsewhere though, and when you have many methods they get repeated a lot and make things unreadable.

jmank88: The actual type constraint is much more interesting. In this case, `ID`.

DylanTinianov: Sounds good, I agree with changing the generics to single letters. However, I think we should apply this across all of the common components and not just TransactionSender. I'm going to consider this out of scope for this PR. 

I will be extracting MultiNode and corresponding common components to our new chainlink-framework repo soon, which will be a better time to make this change in a single place rather than for core and Solana. Here's a ticket for after I do the extraction: https://smartcontract-it.atlassian.net/browse/BCFR-1015

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: Replied to your other comment, will resolve this one.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -28,31 +28,32 @@ var (
...
 }
 
 const sendTxQuorum = 0.7
 
 // SendTxRPCClient - defines interface of an RPC used by TransactionSender to broadcast transaction
-type SendTxRPCClient[TX any] interface {
+type SendTxRPCClient[TX any, RESULT any] interface {
 	// SendTransaction errors returned should include name or other unique identifier of the RPC
-	SendTransaction(ctx context.Context, tx TX) error
+	SendTransaction(ctx context.Context, tx TX) (*RESULT, error)
```
dhaidashenko: I wonder what we'll return instead of `RESULT` for EVMs? Is it going to be `interface{}` and we'll always return `nil`?

One of the options is to replace (*RESULT, error) with a single param. So it will look something like this:
```go
type SendTxResult interface {
	Code() SendTxReturnCode
	Error() error
}

// SendTxRPCClient - defines interface of an RPC used by TransactionSender to broadcast transaction
type SendTxRPCClient[TX any, R SendTxResult] interface {
	// SendTransaction errors returned should include name or other unique identifier of the RPC
	SendTransaction(ctx context.Context, tx TX) SendTxResult
}
```

Solana's implementation of `SendTxResult` will additionally have `Signature() solanaGo.Signature` method.

pkg/solana/txm/txm.go
```diff
@@ -84,9 +89,24 @@ func NewTxm(chainID string, tc func() (client.ReaderWriter, error), cfg config.C
 		txs:    newPendingTxContextWithProm(chainID),
 		ks:     ks,
 		client: utils.NewLazyLoad(tc),
```
dhaidashenko: LazyLoad here and in BlockHistoryEstimator forces an RPC to be reused for all read operations until it's reset on transaction error. With `TransactionSneder` we might not observe the error and will keep using unhealthy RPC.
We should replace `*utils.LazyLoad[client.ReaderWriter]` with a corresponding interface and fully inject MultiNode when it's enabled.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -154,80 +155,81 @@ func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx contex
...
+func (txSender *TransactionSender[TX, RESULT, CHAIN_ID, RPC]) broadcastTxAsync(ctx context.Context, rpc RPC, tx TX) RESULT {
+	result := rpc.SendTransaction(ctx, tx)
+	txSender.lggr.Debugw("Node sent transaction", "tx", tx, "err", result.TxError())
+	resultCode := txSender.txErrorClassifier(tx, result.TxError())
 	if !slices.Contains(sendTxSuccessfulCodes, resultCode) {
-		txSender.lggr.Warnw("RPC returned error", "tx", tx, "err", txErr)
+		txSender.lggr.Warnw("RPC returned error", "tx", tx, "err", result.TxError())
 	}
-	return sendTxResult{Err: txErr, ResultCode: resultCode}
+	result.SetCode(resultCode)
```
dhaidashenko: There is no need for us to `SetCode` here. It can be done by the RPC inside `SendTransaction` impl

pkg/solana/chain_test.go
```diff
@@ -371,3 +318,305 @@ func TestChain_Transact(t *testing.T) {
...
+	sender, err := solana.NewRandomPrivateKey()
+	require.NoError(t, err)
+	receiver, err := solana.NewRandomPrivateKey()
+	require.NoError(t, err)
+	client.FundTestAccounts(t, solana.PublicKeySlice{sender.PublicKey()}, url)
+
+	// configuration
+	cfg := solcfg.NewDefault()
+	cfg.MultiNode.MultiNode.Enabled = ptr(true)
+	cfg.Nodes = append(cfg.Nodes,
```
dhaidashenko, line 392: each of the nodes here is a separate network. This could lead to a flaky behaviour or confusion in the future

pkg/solana/chain_test.go
```diff
@@ -371,3 +318,305 @@ func TestChain_Transact(t *testing.T) {
...
+		require.Error(t, (*result).TxError())
+		require.Equal(t, mn.Fatal, (*result).Code())
+		require.Empty(t, (*result).Signature())
+	})
+}
+
+func TestSolanaChain_MultiNode_Txm(t *testing.T) {
+	cfg := solcfg.NewDefault()
+	cfg.MultiNode.MultiNode.Enabled = ptr(true)
+	cfg.Nodes = []*solcfg.Node{
```
dhaidashenko, line 494: same here. What are we testing by specifying multiple RPCs?

pkg/solana/chain_test.go
```diff
@@ -371,3 +318,305 @@ func TestChain_Transact(t *testing.T) {
...
+			hash.Value.Blockhash,
+			solana.TransactionPayer(signer),
+		)
+		require.NoError(t, txErr)
+		return tx
+	}
+
+	// Send funds twice, along with an invalid transaction
+	require.NoError(t, testChain.txm.Enqueue(tests.Context(t), "test_success", createTx(pubKey, pubKey, pubKeyReceiver, solana.LAMPORTS_PER_SOL)))
+	time.Sleep(500 * time.Millisecond) // pause 0.5s for new blockhash
```
dhaidashenko: What happens if block production is a bit slower?

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,9 +87,9 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
 // * If there is at least one terminal error - returns terminal error
 // * If there is both success and terminal error - returns success and reports invariant violation
 // * Otherwise, returns any (effectively random) of the errors.
-func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx context.Context, tx TX) (SendTxReturnCode, error) {
-	txResults := make(chan sendTxResult)
-	txResultsToReport := make(chan sendTxResult)
+func (txSender *TransactionSender[TX, RESULT, CHAIN_ID, RPC]) SendTransaction(ctx context.Context, tx TX) (*RESULT, error) {
```
dhaidashenko: We should not return a pointer to the interface. It eventually causes code like `if (*result).Signature().IsZero() {`.

pkg/solana/internal/loader.go
```diff
@@ -0,0 +1,41 @@
+package internal
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/utils"
+)
+
+type Loader[T any] interface {
```
dhaidashenko, line 3: Interface definition should be near usage not near implementation definition.

pkg/solana/internal/loader.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (c *loader[T]) Reset() {
+	if c.lazyLoader != nil {
+		c.lazyLoader.Reset()
+	}
+}
+
+func NewLoader[T any](lazyLoad bool, getClient func() (T, error)) *loader[T] {
+	var lazyLoader *utils.LazyLoad[T]
+	if lazyLoad {
```
dhaidashenko: `*utils.LazyLoad` itself implements `Loader[T any]`, why do you need to use it here?

pkg/solana/monitor/balance.go
```diff
@@ -26,34 +28,33 @@ type BalanceClient interface {
 }
 
 // NewBalanceMonitor returns a balance monitoring services.Service which reports the SOL balance of all ks keys to prometheus.
-func NewBalanceMonitor(chainID string, cfg Config, lggr logger.Logger, ks Keystore, newReader func() (BalanceClient, error)) services.Service {
-	return newBalanceMonitor(chainID, cfg, lggr, ks, newReader)
+func NewBalanceMonitor(chainID string, cfg Config, lggr logger.Logger, ks Keystore, reader internal.Loader[BalanceClient]) services.Service {
```
dhaidashenko, line 31: why these changes?

dhaidashenko (changes requested):

pkg/solana/monitor/balance.go
DylanTinianov: BalanceMonitor was also doing caching of a single RPC which this allows us to disable. It also had its own lazy loading logic which should instead use the same lazyloader as everything else.

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: Ahh I can use a named return value and not need to have a pointer!

pkg/solana/chain_test.go
DylanTinianov: Since this is using the **Solana Test Validator** it should always work as expected. The same approach is being used in the current txm tests.

pkg/solana/chain.go
```diff
@@ -288,13 +284,34 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+			}
+			return result.Signature(), nil
+		}
 	}
 
-	tc := func() (client.ReaderWriter, error) {
-		return ch.getClient()
+	var tc internal.Loader[client.ReaderWriter]
+	var bc internal.Loader[monitor.BalanceClient]
+	if cfg.MultiNode.Enabled() {
```
dhaidashenko: nit: do we need two `if cfg.MultiNode.Enabled() {` blocks?

pkg/solana/chain_test.go
dhaidashenko: If the test fails on failure to produce a new block within 0.5s we should avoid using `time.Sleep` and instead explicitly wait for a new block to be produced. 
Otherwise, we might experience flaky behavior due to limited resources when the whole test suite runs in the CI.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,16 +87,16 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
 // * If there is at least one terminal error - returns terminal error
 // * If there is both success and terminal error - returns success and reports invariant violation
 // * Otherwise, returns any (effectively random) of the errors.
-func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx context.Context, tx TX) (SendTxReturnCode, error) {
-	txResults := make(chan sendTxResult)
-	txResultsToReport := make(chan sendTxResult)
+func (txSender *TransactionSender[TX, RESULT, CHAIN_ID, RPC]) SendTransaction(ctx context.Context, tx TX) (result RESULT, err error) {
```
dhaidashenko: SendTransaction should only return `RESULT` to avoid bug where caller will only handle error and won't handle error returned from `result.Err()`.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -154,80 +147,79 @@ func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx contex
...
 	// return temporary error
-	for code, result := range resultsByCode {
-		return code, result[0], nil
+	for _, r := range resultsByCode {
+		return r[0], nil
 	}
 
 	err = fmt.Errorf("expected at least one response on SendTransaction")
-	return Retryable, err, err
+	return result, err
```
dhaidashenko: here we should create result from error

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -145,7 +138,7 @@ func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx contex
 	}()
 
 	if err != nil {
-		return Retryable, err
+		return result, err
```
dhaidashenko: here we should create result from err

pkg/solana/txm/txm.go
```diff
@@ -88,10 +94,25 @@ func NewTxm(chainID string, tc func() (client.ReaderWriter, error), cfg config.C
...
+		return txm.sendTx(ctx, tx)
+	}
+
+	// Send tx using a single RPC client
+	client, err := txm.client.Get()
+	if err != nil {
+		return solanaGo.Signature{}, err
+	}
+	return client.SendTx(ctx, tx)
+}
```
aalu1418: this feels weird to have two sending logics?

is it possible to have the multinode conform to a similar interface as the client?
or wrap it so it can run `Get()` (could be a no-op) and then `SendTx()`?

pkg/solana/chain.go
```diff
@@ -288,13 +284,34 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
+		tc = internal.NewLoader[client.ReaderWriter](func() (client.ReaderWriter, error) { return ch.multiNode.SelectRPC() })
+		bc = internal.NewLoader[monitor.BalanceClient](func() (monitor.BalanceClient, error) { return ch.multiNode.SelectRPC() })
+	} else {
+		tc = utils.NewLazyLoad(func() (client.ReaderWriter, error) { return ch.getClient() })
+		bc = utils.NewLazyLoad(func() (monitor.BalanceClient, error) { return ch.getClient() })
```
aalu1418: this can be cleaned up to initialize the variables with the `else` case and then override if `true`

pkg/solana/chain_test.go
aalu1418: using a `time.Sleep` also can lead to flaky tests - please change this

it would be better to run a loop with a timeout that looks for a new blockchash

pkg/solana/client/classify_errors.go
```diff
@@ -0,0 +1,109 @@
...
+	ErrMaxLoadedAccountsDataSizeExceeded     = regexp.MustCompile(`Transaction exceeded max loaded accounts data size cap`)
+	ErrInvalidLoadedAccountsDataSizeLimit    = regexp.MustCompile(`LoadedAccountsDataSizeLimit set for transaction must be greater than 0\.`)
+	ErrResanitizationNeeded                  = regexp.MustCompile(`Sanitized transaction differed before/after feature activation\. Needs to be resanitized\.`)
+	ErrProgramExecutionTemporarilyRestricted = regexp.MustCompile(`Execution of the program referenced by account at index \d+ is temporarily restricted\.`)
+	ErrUnbalancedTransaction                 = regexp.MustCompile(`Sum of account balances before and after transaction do not match`)
+	ErrProgramCacheHitMaxLimit               = regexp.MustCompile(`Program cache hit max limit`)
+)
+
+// errCodes maps regex patterns to corresponding return code
+var errCodes = map[*regexp.Regexp]mn.SendTxReturnCode{
```
aalu1418, line 56: can you list reasons why each on is retryable or fatal?

aalu1418 (changes requested):

aalu1418:
> Note: Soak testing was successful

links to the soak test? was this against a live network? please list the conditions of the test

pkg/solana/client/multinode/transaction_sender.go
aalu1418, line 1: has the multinode code been extracted?

DylanTinianov: Not yet, still finalizing the TransactionSender code as there were changes needed for Solana. Once this is merged I will do the extraction.

pkg/solana/client/classify_errors.go
DylanTinianov: Just removed all fatal error codes. By default a transaction will be `Retryable`, so now I only added the obvious exceptions with matching codes.

In the future if we determine that some of these error codes cause `Retryable` transactions to fail on retry consistently, then we can add them to the map with a non-retryable code.

DylanTinianov:
@aalu1418 Updated description with link to soak test details: https://github.com/smartcontractkit/chainlink-solana/pull/894

pkg/solana/txm/txm.go
DylanTinianov: Maybe it would make more sense to pass the `TransactionSender` directly rather than an override function? My thought was to make it generic, but it might be more clear to just use the component itself. 

I don't see a nice way to wrap it since `client.Get()` is used for returning a `ReaderWriter` which is essentially a single RPC, while using the `TransactionSender` is a layer higher using all available RPCs.

pkg/solana/client/classify_errors.go
```diff
@@ -0,0 +1,75 @@
...
+		return mn.Successful
+	}
+
+	errMsg := err.Error()
+	for pattern, code := range errCodes {
+		if pattern.MatchString(errMsg) {
+			return code
+		}
+	}
+	return mn.Retryable
```
dhaidashenko, line 74: Eventually, we want this to be unknown. It will help us spot misconfigured RPC more easily. And ensure we properly handle transactions. It does not seem to be a blocker for the launch. So lets just create a ticket to track that effort.

pkg/solana/client/multinode_client.go
```diff
@@ -300,8 +300,40 @@ func (m *MultiNodeClient) GetInterceptedChainInfo() (latest, highestUserObservat
...
+	txErr error
+	code  mn.SendTxReturnCode
+	sig   solana.Signature
+}
+
+var _ mn.SendTxResult = (*SendTxResult)(nil)
+
+func NewSendTxResult(err error) *SendTxResult {
+	return &SendTxResult{
+		err: err,
```
dhaidashenko: We need to classify the error, otherwise the mn.SendTxReturnCode is 0, which does not match any value

pkg/solana/client/classify_errors.go
aalu1418: what does defaulting to retryable imply?

is `retryable` only something within multinode/txsender or is this a signal to the TXM?

if it's a signal to the TXM that's a bit more concerning as we can burn tokens really fast that way

pkg/solana/client/classify_errors.go
DylanTinianov: Created a ticket: https://smartcontract-it.atlassian.net/browse/NONEVM-796

pkg/solana/client/multinode_client.go
DylanTinianov: Added the error classification. However, result.Error() should be checked and handled before result.TxError() or result.Code() are used.

aalu1418 (dismissed):
looks good to me! I'll wait for for more folks on the framework side to review

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,13 +91,16 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
...
+	txResultsToReport := make(chan RESULT)
 	primaryNodeWg := sync.WaitGroup{}
 
-	ctx, cancel := txSender.chStop.Ctx(ctx)
-	defer cancel()
+	if txSender.State() != "Started" {
+		return txSender.newResult(errors.New("TransactionSender not started"))
+	}
+
+	ctx, _ = txSender.chStop.Ctx(ctx)
```
dhaidashenko: This seems like a bug in the initial version of the txSender, so feel free to create it as a separate PR (but we'll have to merge it first)
We should only pass caller's context into `txSender.collectTxResults`. All other calls are background and can not be canceled by the caller, so we should pass txSender's root context.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -154,80 +154,79 @@ func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx contex
...
 }
 
-type sendTxResults map[SendTxReturnCode][]error
+type sendTxResults[RESULT any] map[SendTxReturnCode][]RESULT
 
-func aggregateTxResults(resultsByCode sendTxResults) (returnCode SendTxReturnCode, txResult error, err error) {
-	severeCode, severeErrors, hasSevereErrors := findFirstIn(resultsByCode, sendTxSevereErrors)
-	successCode, successResults, hasSuccess := findFirstIn(resultsByCode, sendTxSuccessfulCodes)
+func aggregateTxResults[RESULT any](resultsByCode sendTxResults[RESULT]) (result RESULT, criticalErr error) {
+	_, severeErrors, hasSevereErrors := findFirstIn(resultsByCode, sendTxSevereErrors)
```
dhaidashenko: nit: all calls to `findFirstIn` ignore the first result param. Can we drop it?

pkg/solana/chain.go
```diff
@@ -288,13 +289,25 @@ func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.L
...
 		// clientCache will not be used if multinode is enabled
 		ch.clientCache = nil
-	}
 
-	tc := func() (client.ReaderWriter, error) {
-		return ch.getClient()
+		// Send tx using MultiNode transaction sender
+		sendTx = func(ctx context.Context, tx *solanago.Transaction) (solanago.Signature, error) {
+			// Use empty context since individual RPC timeouts are handled by the client
+			result := ch.txSender.SendTransaction(context.Background(), tx)
```
dhaidashenko: We should not use context.Background. 
I agree that it does not make sense to override RPC timeouts, but the fact that we can not handle this case signals a deeper problem. 
I've described what seems to be a root cause in another comment.

jmank88: Why doesn't it make sense to pass the parent context here? We can at least use `context.WithoutCancel` to maintain the request scoped data, but that must be justified, and all behavior should be cancelled on shutdown besides.
```suggestion
			result := ch.txSender.SendTransaction(context.WithoutCancel(ctx), tx)
```

pkg/solana/client/multinode/transaction_sender.go
jmank88: Close funcs must not be ignored because they leak goroutines/memory:
```suggestion
	ctx, close = txSender.chStop.Ctx(ctx)
	defer close()
```

jmank88 (changes requested):

pkg/solana/client/classify_errors.go
DylanTinianov: Currently Solana TXM doesn't care which type of error is returned. It only looks for `context cancelled` to determine if it should retry. We can certainly optimize this later and make use of the error codes.

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,16 +91,26 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
...
-	txResults := make(chan sendTxResult)
-	txResultsToReport := make(chan sendTxResult)
+func (txSender *TransactionSender[TX, RESULT, CHAIN_ID, RPC]) SendTransaction(ctx context.Context, tx TX) RESULT {
+	txResults := make(chan RESULT)
+	txResultsToReport := make(chan RESULT)
 	primaryNodeWg := sync.WaitGroup{}
 
-	ctx, cancel := txSender.chStop.Ctx(ctx)
-	defer cancel()
+	if txSender.State() != "Started" {
```
dhaidashenko, line 99: nit: we should use `txSender.IfStarted`

dhaidashenko (dismissed):
Nice work!

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -145,89 +152,92 @@ func (txSender *TransactionSender[TX, CHAIN_ID, RPC]) SendTransaction(ctx contex
...
-		return Retryable, err
+		return txSender.newResult(err)
 	}
 
 	txSender.wg.Add(1)
-	go txSender.reportSendTxAnomalies(tx, txResultsToReport)
+	reportWg.Add(1)
+	go func() {
+		txSender.reportSendTxAnomalies(tx, txResultsToReport)
+		reportWg.Done()
```
jmank88: It's important to defer these so that they still execute on panic:
```suggestion
		defer reportWg.Done()
		txSender.reportSendTxAnomalies(tx, txResultsToReport)
```

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,16 +91,26 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
...
+	}
+
+	txSenderCtx, cancel := txSender.chStop.NewCtx()
+	reportWg := sync.WaitGroup{}
+	defer func() {
+		go func() {
+			reportWg.Wait()
+			cancel()
+		}()
+	}()
```
jmank88, line 110: This is incredibly unusual. Does `reportWg` only exist in order to coordinate this cancellation?

jmank88 (dismissed):

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -94,16 +91,26 @@ type TransactionSender[TX any, CHAIN_ID ID, RPC SendTxRPCClient[TX]] struct {
...
 
 	healthyNodesNum := 0
-	err := txSender.multiNode.DoAll(ctx, func(ctx context.Context, rpc RPC, isSendOnly bool) {
+	err := txSender.multiNode.DoAll(txSenderCtx, func(ctx context.Context, rpc RPC, isSendOnly bool) {
```
jmank88, line 113: Why can't we pass the standard parent `ctx` here?
If the go routines inside of the func are the problem, then they should handled directly, e.g. via `context.WithoutCancel(ctx)`

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: Yeah, we need to ensure `cancel` is called after the concurrent call to `reportSendTxAnomalies`, as well as the entire `SendTransaction` function, returns.

dhaidashenko (approved):

aalu1418 (approved):

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: We only want to cancel this context when transaction sender is closed down. We still need the parent ctx for collecting the results, but the context we sent to multinode is separate.

pkg/solana/client/multinode/transaction_sender.go
jmank88: IMHO this is more than a nit. As written, it is racey.

pkg/solana/client/multinode/transaction_sender.go
jmank88: I only see one caller of Add/Done on `reportWg`. Why do we need to defer spawning a goroutine to wait for one thing? Why not just call `cancel()` inline from the other goroutine that calls `Done`?

pkg/solana/client/multinode/transaction_sender.go
jmank88: What would happen if you passed `ctx` here, but you used `context.WithoutCancel(ctx)` from the goroutines? Why doesn't that work?

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: We're also waiting for `collectTxResults` to finish as well, which is why it's deferred. We can't call `cancel()` after `collectTxResults` directly nor after `reportSendTxAnomalies` since we must wait for both to complete.

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: If we use `context.WithoutCancel(ctx)` we would be ignoring the TransactionSender being closed inside of the `broadcastTxAsync` and in turn the RPC `SendTx` functions. We want the RPC client to be responsible for the `SendTx` timeout, but still include the case where Transaction Sender if closed.

pkg/solana/client/multinode/transaction_sender.go
jmank88: Why do they have to _complete_ before we cancel the context which is wired to the goroutines that fed them inputs? We already wait on those goroutines to close the channels some where else. Can we just cancel there?
```diff
		primaryNodeWg.Wait()
		close(txResultsToReport)
		close(txResults)
+		cancel()
```

pkg/solana/client/multinode/transaction_sender.go
jmank88: I agree that we should not orphan the goroutines, but this illustrates the point that we can in fact pass the `ctx` normally as a starting point without causing any inherent problems.

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: Is it guaranteed in this case that `collectTxResults` would read the last result before exiting due to `ctx.Done()`?

I see that `case r := <-txResults:` would be waiting for a result which would come before we call `cancel`, however couldn't `ctx.Done()` trigger first if both cases are ready?

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: The only potential problem is some RPC calls might timeout slightly early, but this shouldn't be an issue since the timeout should be large enough.

pkg/solana/client/multinode/transaction_sender.go
jmank88: But `cancel` is associated with `txSenderCtx` which is not passed to `collectTxResults`.

pkg/solana/client/multinode/transaction_sender.go
jmank88: What do you mean that RPC calls might timeout early? There is no need to compromise anything with this approach.

dhaidashenko: > What would happen if you passed ctx here, but you used context.WithoutCancel(ctx) from the goroutines? Why doesn't that work?

@jmank88 if we use `context.WithoutCancel(ctx)` from the goroutines we won't be able to cancel them when txSender is closed.

jmank88: `context.WithCancel` will only disconnect the parent context cancellation. It does not preclude us from cancelling the resulting context. The point is that we _do_ want `multiNode.DoAll` to be responsive to cancellation of the parent context (e.g. don't even launch the goroutines if already cancelled) - only the goroutines workers need to managed separately.

pkg/solana/client/multinode/transaction_sender.go
dhaidashenko: The initial idea was to call cancel when `reportSendTxAnomalies` - the last background job spawned within the call, is done. 
But as `reportSendTxAnomalies` ignores context we can simplify the code and cancel after `primaryNodeWg.Wait()`.
@DylanTinianov my bad, sorry for overcomplicating.

pkg/solana/client/multinode/transaction_sender.go
dhaidashenko: > The point is that we do want multiNode.DoAll to be responsive to cancellation of the parent context

Agree.
Are you suggesting something like the following:
```go
err := txSender.multiNode.DoAll(ctx, func(ctx context.Context, rpc RPC, isSendOnly bool) {
		// Primary Nodes
		healthyNodesNum++
		primaryNodeWg.Add(1)
		go func() {
			ctx, cancel := txSender.chStop.Ctx(context.WithoutCancel(ctx))
			defer cancel()
			defer primaryNodeWg.Done()
			r := txSender.broadcastTxAsync(ctx, rpc, tx)
			select {
			case <-ctx.Done():
				return
			case txResults <- r:
			}

			select {
			case <-ctx.Done():
				return
			case txResultsToReport <- r:
			}
		}()
	})
```

jmank88: Yeah, and we could cancel those internal contexts additional ways if we want/need to.

PR #884, opened 2024-10-09
Add EstimateComputeUnitLimit feature

amit-momin:
[BCFR-1012](https://smartcontract-it.atlassian.net/browse/BCFR-1012)

- Added feature to estimate compute unit limits for transactions if `EstimateComputeUnitLimit` config is enabled
- Required for more complex transactions that can use varying amounts of compute units unlike Data Feed transactions

Chainlink Repo PR: https://github.com/smartcontractkit/chainlink/pull/14741

[BCFR-1012]: https://smartcontract-it.atlassian.net/browse/BCFR-1012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/txm/txm.go
```diff
@@ -176,6 +185,12 @@ func (txm *Txm) sendWithRetry(chanCtx context.Context, baseTx solanaGo.Transacti
...
 
+	if txcfg.EstimateComputeUnitLimit {
+		if computeUnitLimitErr := txm.EstimateAndSetComputeUnitLimit(chanCtx, &baseTx); computeUnitLimitErr != nil {
+			return solanaGo.Transaction{}, uuid.Nil, solanaGo.Signature{}, fmt.Errorf("failed to estimate and set compute unit limit: %w", computeUnitLimitErr)
+		}
+	}
+
 	// add compute unit limit instruction - static for the transaction
 	// skip if compute unit limit = 0 (otherwise would always fail)
 	if txcfg.ComputeUnitLimit != 0 {
```
aalu1418: would this make more sense in the `Enqueue` function and then setting the `txcfg.ComputeUnitLimit`?

it would deduplicate some of the logic of building the transaction multiple times?

pkg/solana/config/config.go
```diff
@@ -53,26 +54,28 @@ type Config interface {
...
+	Commitment               *string
+	MaxRetries               *int64
+	FeeEstimatorMode         *string
+	ComputeUnitPriceMax      *uint64
+	ComputeUnitPriceMin      *uint64
+	ComputeUnitPriceDefault  *uint64
+	FeeBumpPeriod            *config.Duration
+	BlockHistoryPollPeriod   *config.Duration
+	ComputeUnitLimitDefault  *uint32
+	EstimateComputeUnitLimit *bool
```
aalu1418, line 78: note: changes to config also require changes to core tests + documentation

pkg/solana/txm/txm_test.go
aalu1418, line 1: nit: can we name this file to something different?

using the same named file creates a giant diff that makes it look like a ton more changes than it actually is

pkg/solana/txm/txm.go
```diff
@@ -135,14 +140,18 @@ func (txm *Txm) run() {
...
+			if !msg.cfg.EstimateComputeUnitLimit {
+				// send tx + signature to simulation queue
+				msg.tx = &tx
+				msg.signature = sig
+				msg.id = id
+				select {
+				case txm.chSim <- msg:
+				default:
+					txm.lggr.Warnw("failed to enqeue tx for simulation", "queueFull", len(txm.chSend) == MaxQueueLen, "tx", msg)
+				}
```
aalu1418: leaving the extra simulate loop should be ok?

otherwise the retry can only ever be cancelled by timeout or confirmed transaction - rather than something like a duplicate report or something

pkg/solana/txm/txm.go
```diff
@@ -568,6 +583,65 @@ func (txm *Txm) Enqueue(accountID string, tx *solanaGo.Transaction, txCfgs ...Se
 	return nil
 }
 
+func (txm *Txm) EstimateAndSetComputeUnitLimit(ctx context.Context, tx *solanaGo.Transaction) (err error) {
```
aalu1418: can we share logic with the existing simulate call here?

aalu1418 (changes requested):

pkg/solana/txm/txm.go
amit-momin: Could a second simulation pick up an error that the original didn't? I was thinking if a transaction passes simulation for the compute unit limit then timeout or confirmation should be the only reasons to cancel retry. I guess it doesn't hurt to keep it in case there's some state change in between

pkg/solana/txm/txm.go
amit-momin: Agree this is probably cleaner. I've updated this in the latest commit

pkg/solana/txm/txm.go
amit-momin: Good point. I've updated the code to remove some duplicity here

pkg/solana/txm/txm.go
```diff
@@ -568,6 +550,82 @@ func (txm *Txm) Enqueue(accountID string, tx *solanaGo.Transaction, txCfgs ...Se
...
+
+func (txm *Txm) simulateTx(ctx context.Context, tx *solanaGo.Transaction) (res *rpc.SimulateTransactionResult, err error) {
+	// get client
+	client, err := txm.client.Get()
+	if err != nil {
+		txm.lggr.Errorw("failed to get client", "error", err)
+		return
+	}
+
+	res, err = client.SimulateTx(ctx, tx, nil) // use default options (does not verify signatures)
```
Farber98, line 604: Do we always want `finalized` [commitment](https://solana.com/docs/rpc#configuring-state-commitment) or may be worth adding the option to specify it?

pkg/solana/txm/txm.go
```diff
@@ -568,6 +550,82 @@ func (txm *Txm) Enqueue(accountID string, tx *solanaGo.Transaction, txCfgs ...Se
...
+	if unitsConsumed > math.MaxUint32 {
+		txm.lggr.Debug("compute units used with buffer greater than uint32 max", "unitsConsumed", unitsConsumed)
+		// Do not return error to allow falling back to default compute unit limit
+		return 0, nil
+	}
+
+	return uint32(unitsConsumed), nil
+}
+
+func (txm *Txm) simulateTx(ctx context.Context, tx *solanaGo.Transaction) (res *rpc.SimulateTransactionResult, err error) {
```
Farber98, line 596: We will `try` to simulate every tx before broadcasting it right? Will this be handled gracefully in terms of performance?

Farber98 (commented):
Good changes! The refactor looks good and makes the code easier to follow

pkg/solana/txm/txm.go
aalu1418: simulation before broadcasting should only happen if that is configured - I believe default is `false`

pkg/solana/txm/txm.go
aalu1418: If you're using the client defined in the solana package, it should use the commitment level from the network config

https://github.com/smartcontractkit/chainlink-solana/blob/4e1d0ae8315a0a0abe828ffddf4a45884349f183/pkg/solana/client/client.go#L289

you can pass in custom simulation options

pkg/solana/txm/txm.go
aalu1418: hmmmm - fair point

technically we are modifying the transaction with the limit and the price - so technically it could still fail if estimation/tx building for either was broken 🤔 - unlikely but possible

aalu1418 (commented):
changes look good to me!

i'd like to see the core side PR before approval / merge :)

also CI should fully pass on this side - but i think you might need this PR to be merged for that to happen: https://github.com/smartcontractkit/chainlink/pull/14011 (or need to use the `core_ref` inside the PR description)

pkg/solana/txm/txm.go
amit-momin: Oh ya good point. Makes sense to keep it then. I've undone this in the latest commit

pkg/solana/txm/txm.go
```diff
@@ -568,6 +550,82 @@ func (txm *Txm) Enqueue(accountID string, tx *solanaGo.Transaction, txCfgs ...Se
 	return nil
 }
 
+func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Transaction) (uint32, error) {
```
silaslenihan, line 559: Would be nice to add comments for above method and `simulateTx` to summarize their usage to continue the good documentation in this package

silaslenihan (dismissed):

pkg/solana/txm/txm.go
amit-momin: Good point, added comments in the latest commit

silaslenihan (dismissed):

Farber98 (dismissed):

aalu1418 (dismissed):

aalu1418 (approved):

PR #894, opened 2024-10-17
MultiNode Soak Testing

DylanTinianov:
## Changes
- Enabled the use of Multiple RPCs for Soak Testing
- Added `run_soak_test.sh` to easily trigger soak tests and exit the terminal at the correct time
- Wrapped `SendTransaction()` functionality within `txSender.IfStarted(func() {...})`
- Improved `TransactionSender` context handling
- Slightly relaxed some default configurations
- Added reader getters to ensure a single RPC is not passed around negating MultiNode

Closes: https://smartcontract-it.atlassian.net/browse/NONEVM-842

## Latest Soak Test (Devnet)

Soak tests are being run using a [local set up](https://smartcontract-it.atlassian.net/wiki/spaces/QA/pages/601097421/Solana+testing+wiki#Integration-tests).

`CommitHash` ca7c98259f21439e004507edc03046347bcf9766 (latest) 
--- |
Namespace: `solana-gauntlet-embedded-d7efd `
Num RPCs  Latest Test SyncThreshold 
[Grafana Logs](https://grafana.ops.prod.cldev.sh/explore?schemaVersion=1&panes=%7B%22u3i%22:%7B%22datasource%22:%22P8E80F9AEF21F6940%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22solana-gauntlet-embedded-d7efd%5C%22%7D%20%7C%3D%20%20%60Solana.devnet.Node%60%20or%20%60Solana.devnet.TransactionSender%60%20or%20%60Solana.devnet.MultiNode%60%20%7C%20json%20%7C%20level%3D~%5C%22%7Cerror%7Ccrit%7Cwarn%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P8E80F9AEF21F6940%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

### Config
```
# This is the default configuration so OCR2 tests can run without issues
[ChainlinkImage]
version=<Latest CI Build Image>

[Logging]
test_log_collect=false

[Logging.LogStream]
log_targets=["file"]
log_producer_timeout="10s"
log_producer_retry_limit=10

[Common]
internal_docker_repo = ***
inside_k8 = true
network = "devnet"
user = "DylanTinianov"
stateful_db = true
devnet_image = "layerzerolabs/solana:1.17.31"

[OCR2]
node_count = 6
test_duration = "72h"
number_of_rounds = 3
```

### RPCs
Using 3 internal solana-devnet RPCs


pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -164,10 +165,11 @@ func (n *node[CHAIN_ID, HEAD, RPC]) aliveLoop() {
...
 				return
 			}
-			_, latestChainInfo := n.StateAndLatest()
-			if outOfSync, liveNodes := n.isOutOfSyncWithPool(latestChainInfo); outOfSync {
+			if outOfSync, liveNodes := n.isOutOfSyncWithPool(); outOfSync {
 				// note: there must be another live node for us to be out of sync
-				lggr.Errorw("RPC endpoint has fallen behind", "blockNumber", latestChainInfo.BlockNumber, "totalDifficulty", latestChainInfo.TotalDifficulty, "nodeState", n.getCachedState())
+				_, highest := n.poolInfoProvider.LatestChainInfo()
+				_, latestChainInfo := n.StateAndLatest()
+				lggr.Errorw("RPC endpoint has fallen behind", "blockNumber", latestChainInfo.BlockNumber, "bestLatestBlockNumber", highest.BlockNumber, "totalDifficulty", latestChainInfo.TotalDifficulty, "nodeState", n.getCachedState())
```
dhaidashenko: Consider moving the log line into `isOutOfSyncWithPool` and logging it as an error only if the node's cached state is alive. Otherwise higher and latest chain info may be inaccurate at the moment of writing the log

pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -192,7 +194,8 @@ func (n *node[CHAIN_ID, HEAD, RPC]) aliveLoop() {
 		case <-headsSub.NoNewHeads:
 			// We haven't received a head on the channel for at least the
 			// threshold amount of time, mark it broken
-			lggr.Errorw(fmt.Sprintf("RPC endpoint detected out of sync; no new heads received for %s (last head received was %v)", noNewHeadsTimeoutThreshold, localHighestChainInfo.BlockNumber), "nodeState", n.getCachedState(), "latestReceivedBlockNumber", localHighestChainInfo.BlockNumber, "noNewHeadsTimeoutThreshold", noNewHeadsTimeoutThreshold)
+			_, latestChainInfo := n.StateAndLatest()
+			lggr.Errorw(fmt.Sprintf("RPC endpoint detected out of sync; no new heads received for %s (last head received was %v)", noNewHeadsTimeoutThreshold, latestChainInfo.BlockNumber), "nodeState", n.getCachedState(), "latestReceivedBlockNumber", latestChainInfo.BlockNumber, "noNewHeadsTimeoutThreshold", noNewHeadsTimeoutThreshold)
```
dhaidashenko: We decide to reset the noNewHeads timer based on `localHighestChainInfo` values. Why do you prefer logging latestChainInfo instead?

pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -365,13 +368,14 @@ func (n *node[CHAIN_ID, HEAD, RPC]) isOutOfSyncWithPool(localState ChainInfo) (o
 	}
 	// Check against best node
 	ln, ci := n.poolInfoProvider.LatestChainInfo()
+	_, localChainInfo := n.StateAndLatest()
```
dhaidashenko:
```suggestion
	localChainInfo, _ := n.rpc.GetInterceptedChainInfo()
```
`StateAndLatest` recalculates Node's state on every call. It's preferred to use `GetInterceptedChainInfo`

pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -500,7 +504,8 @@ func (n *node[CHAIN_ID, HEAD, RPC]) outOfSyncLoop(syncIssues syncStatus) {
 				continue
 			}
 
-			receivedNewHead := n.onNewFinalizedHead(lggr, &localHighestChainInfo, latestFinalized)
+			_, latestChainInfo := n.StateAndLatest()
+			receivedNewHead := n.onNewFinalizedHead(lggr, &latestChainInfo, latestFinalized)
```
dhaidashenko: Why this change?
An RPC can't recover from out of sync state as `latestChainInfo` already updated by `latestFinalized` block before the check if `latestFinalized` is a new finalized block.

pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -511,7 +516,9 @@ func (n *node[CHAIN_ID, HEAD, RPC]) outOfSyncLoop(syncIssues syncStatus) {
 				finalizedHeadsSub.ResetTimer(noNewFinalizedBlocksTimeoutThreshold)
 			}
 
-			lggr.Debugw(msgReceivedFinalizedBlock, "blockNumber", latestFinalized.BlockNumber(), "syncIssues", syncIssues)
+			_, highestSeen := n.poolInfoProvider.LatestChainInfo()
```
dhaidashenko: use `HighestUserObservations` to avoid state recalculation for all nodes.

pkg/solana/client/multinode/node_lifecycle.go
```diff
@@ -511,7 +516,9 @@ func (n *node[CHAIN_ID, HEAD, RPC]) outOfSyncLoop(syncIssues syncStatus) {
...
+
+			lggr.Debugw(msgReceivedFinalizedBlock, "blockNumber", latestFinalized.BlockNumber(), "highestBlockNumber", highestSeen.FinalizedBlockNumber, "syncIssues", syncIssues)
```
dhaidashenko:
```suggestion
			lggr.Debugw(msgReceivedFinalizedBlock, "blockNumber", latestFinalized.BlockNumber(), "poolHighestBlockNumber", highestSeen.FinalizedBlockNumber, "syncIssues", syncIssues)
```

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -92,77 +92,73 @@ type TransactionSender[TX any, RESULT SendTxResult, CHAIN_ID ID, RPC SendTxRPCCl
...
+	var result RESULT
+	if !txSender.IfStarted(func() {
+		txResults := make(chan RESULT)
+		txResultsToReport := make(chan RESULT)
+		primaryNodeWg := sync.WaitGroup{}
+
+		txSenderCtx, cancel := txSender.chStop.Ctx(ctx)
+
+		healthyNodesNum := 0
+		err := txSender.multiNode.DoAll(txSenderCtx, func(ctx context.Context, rpc RPC, isSendOnly bool) {
```
dhaidashenko: Could you update this according to latest discussion with Jordan from TxSender PR?

pkg/solana/client/multinode/node_lifecycle.go
DylanTinianov: Changed during debugging, just changed it back now

pkg/solana/client/multinode/transaction_sender.go
DylanTinianov: Done

pkg/solana/client/multinode/node_lifecycle.go
DylanTinianov: Changed back

pkg/solana/client/multinode/transaction_sender.go
```diff
@@ -92,89 +92,88 @@ type TransactionSender[TX any, RESULT SendTxResult, CHAIN_ID ID, RPC SendTxRPCCl
...
+				case txResults <- r:
+				}
+
+				select {
+				case <-ctx.Done():
+					txSender.lggr.Debugw("Failed to send tx results to report", "err", ctx.Err())
+					return
+				case txResultsToReport <- r:
+				}
+			}()
```
jmank88:
```suggestion
			go func(ctx context.Context) {
				ctx, cancel := txSender.chStop.Ctx(context.WithoutCancel(ctx))
				defer cancel()
				defer primaryNodeWg.Done()
				r := txSender.broadcastTxAsync(ctx, rpc, tx)
				select {
				case <-ctx.Done():
					txSender.lggr.Debugw("Failed to send tx results", "err", ctx.Err())
					return
				case txResults <- r:
				}

				select {
				case <-ctx.Done():
					txSender.lggr.Debugw("Failed to send tx results to report", "err", ctx.Err())
					return
				case txResultsToReport <- r:
				}
			}(ctx)
```

integration-tests/common/common.go
```diff
@@ -51,7 +51,7 @@ type TestEnvDetails struct {
 type ChainDetails struct {
 	ChainName             string
 	ChainID               string
-	RPCUrl                string
+	RPCUrl                []string
```
jmank88: Should we make it plural?
```suggestion
	RPCUrls                []string
```

integration-tests/common/test_common.go
jmank88, line 1: Do we have to choose one here? Why the first? Might we prefer to pass the slice to be selected from later?

integration-tests/scripts/run_soak_test.sh
jmank88, line 1: This looks very complicated. Why do we need to manually manage timeouts and kill processes? Why not just set the timeout that we want to use on the go command instead of setting `24h` but killing it after 15 minutes?

jmank88: Oh it's more complicated. We pass `-timeout 24h`, `-test.timeout 30m`, and then `sleep 900` seconds before killing.

pkg/solana/client/multinode/multi_node.go
```diff
@@ -372,6 +372,6 @@ func (c *MultiNode[CHAIN_ID, RPC]) report(nodesStateInfo []nodeWithState) {
 		c.lggr.Criticalw(rerr.Error(), "nodeStates", nodesStateInfo)
 		c.SvcErrBuffer.Append(rerr)
 	} else if dead > 0 {
-		c.lggr.Errorw(fmt.Sprintf("At least one primary node is dead: %d/%d nodes are alive", live, total), "nodeStates", nodesStateInfo)
```
dhaidashenko, line 375: why this change?

dhaidashenko (dismissed):

pkg/solana/client/multinode/multi_node.go
DylanTinianov: MultiNode should gracefully handle when RPCs are dead. I think we should only log an error when there is a problem that could hinder a user, or an error in the MultiNode itself. Having some RPCs out of sync or non-responsive is expected to happen from time to time, and thus I believe a warning is sufficient here. As long as there are enough healthy RPCs to select from, then there shouldn't be an error. I can change this back if you have a strong opinion on this.

integration-tests/scripts/run_soak_test.sh
DylanTinianov: Yeah I made the script so I didn't have to babysit the terminal, and just used the command outlined in the [soak test docs ](https://smartcontract-it.atlassian.net/wiki/spaces/QA/pages/601097421/Solana+testing+wiki#Prerequisites)

You have to exit the terminal once`ocr2:inspect:responses` occurs, and if you exit too early or too late then the namespace is terminated. I haven't looked too deeply into improving the logic itself, but I think it will be out of scope for this PR.

jmank88: But `-timeout` and `-test.timeout` are different ways of setting the same flag:
> Each of these flags is also recognized with an optional 'test.' prefix,
as in -test.v.

https://pkg.go.dev/cmd/go/internal/test

pkg/monitoring/chain_reader.go
```diff
@@ -31,11 +32,13 @@ type chainReader struct {
...
 func (c *chainReader) GetState(ctx context.Context, account solana.PublicKey, commitment rpc.CommitmentType) (state pkgSolana.State, blockHeight uint64, err error) {
-	return pkgSolana.GetState(ctx, c.client, account, commitment)
+	getReader := func() (client.AccountReader, error) { return c.client, nil }
+	return pkgSolana.GetState(ctx, getReader, account, commitment)
 }
 
 func (c *chainReader) GetLatestTransmission(ctx context.Context, account solana.PublicKey, commitment rpc.CommitmentType) (answer pkgSolana.Answer, blockHeight uint64, err error) {
-	return pkgSolana.GetLatestTransmission(ctx, c.client, account, commitment)
+	getReader := func() (client.AccountReader, error) { return c.client, nil }
+	return pkgSolana.GetLatestTransmission(ctx, getReader, account, commitment)
```
dhaidashenko, line 41: Could you double check that `pkgSolana.GetLatestTransmission` is not used in other repos to ensure we are not breaking anything

pkg/solana/transmitter.go
```diff
@@ -32,7 +30,12 @@ func (c *Transmitter) Transmit(
 	report types.Report,
 	sigs []types.AttributedOnchainSignature,
 ) error {
-	blockhash, err := c.reader.LatestBlockhash(ctx)
+	reader, err := c.reader()
```
dhaidashenko: nit: we should not call variable that holds method to get reader as reader

dhaidashenko (dismissed):

pkg/monitoring/chain_reader.go
DylanTinianov: Yeah good call, I'll check before merging

dhaidashenko (commented):
Why in TxSender Result we need both TxError and Error?

dhaidashenko (dismissed):

integration-tests/common/test_common.go
DylanTinianov: Going to resolve this for now as modifying this is not required for anything at the moment. Will keep this in mind for later.

dhaidashenko (dismissed):

pkg/solana/client/multinode/poller.go
```diff
@@ -65,7 +65,8 @@ func (p *Poller[T]) Err() <-chan error {
 }
 
 func (p *Poller[T]) pollingLoop(ctx context.Context) {
-	ticker := time.NewTicker(p.pollingInterval)
+	tickerCfg := services.TickerConfig{Initial: 0, JitterPct: services.DefaultJitter}
+	ticker := tickerCfg.NewTicker(p.pollingInterval)
```
aalu1418: isn't this redundant with the default? [`services.NewTicker`](https://github.com/smartcontractkit/chainlink-common/blob/ad84e3712352552a7c8720860aed7942d01228c8/pkg/services/ticker.go#L16)

pkg/solana/config_tracker.go
```diff
@@ -5,13 +5,11 @@ import (
...
 	"github.com/smartcontractkit/libocr/offchainreporting2/reportingplugin/median"
 	"github.com/smartcontractkit/libocr/offchainreporting2/types"
-
-	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
 )
 
 type ConfigTracker struct {
 	stateCache *StateCache
-	reader     client.Reader
+	reader     GetReader
```
aalu1418:
```suggestion
	getReader     GetReader
```

other places also use `getReader` - naming is also clearer

pkg/solana/cache_test.go
```diff
@@ -106,8 +106,9 @@ func TestGetState(t *testing.T) {
 	}))
 	defer mockServer.Close()
 
+	reader := func() (client.AccountReader, error) { return testSetupReader(t, mockServer.URL), nil }
```
aalu1418: why is this recreating a client/reader each time?

it would be cleaner to create a reader and then return it with a `getReader` function

pkg/solana/relay.go
```diff
@@ -187,6 +187,8 @@ func (r *Relayer) NewAutomationProvider(ctx context.Context, rargs relaytypes.Re
 
 var _ relaytypes.ConfigProvider = &configProvider{}
 
+type GetReader func() (client.Reader, error)
+
```
aalu1418: why is this defined here?

`GetAccountReader` is defined where it is being used...

pkg/solana/state_cache.go
```diff
@@ -23,16 +23,24 @@ type StateCache struct {
 	*client.Cache[State]
 }
 
-func NewStateCache(stateID solana.PublicKey, chainID string, cfg config.Config, reader client.Reader, lggr logger.Logger) *StateCache {
+type GetAccountReader func() (client.AccountReader, error)
+
+func NewStateCache(stateID solana.PublicKey, chainID string, cfg config.Config, reader GetReader, lggr logger.Logger) *StateCache {
```
aalu1418: please double check the naming across your PR - it should be `getReader` not `reader`

pkg/solana/transmissions_cache.go
```diff
@@ -19,19 +19,26 @@ type TransmissionsCache struct {
 	*client.Cache[Answer]
 }
 
-func NewTransmissionsCache(transmissionsID solana.PublicKey, chainID string, cfg config.Config, reader client.Reader, lggr logger.Logger) *TransmissionsCache {
+func NewTransmissionsCache(transmissionsID solana.PublicKey, chainID string, cfg config.Config, getReader GetReader, lggr logger.Logger) *TransmissionsCache {
```
aalu1418: why are we passing `getReader` to just wrap it into `getAccountReader`? it would be cleaner to just pass account reader

aalu1418 (changes requested):
overall looks good - just some code cleanliness things

dhaidashenko (dismissed):

pkg/solana/state_cache.go
```diff
@@ -23,16 +23,25 @@ type StateCache struct {
...
 
-func NewStateCache(stateID solana.PublicKey, chainID string, cfg config.Config, reader client.Reader, lggr logger.Logger) *StateCache {
+type GetReader func() (client.Reader, error)
+type GetAccountReader func() (client.AccountReader, error)
+
+func NewStateCache(stateID solana.PublicKey, chainID string, cfg config.Config, getReader GetReader, lggr logger.Logger) *StateCache {
 	name := "ocr2_median_state"
 	getter := func(ctx context.Context) (State, uint64, error) {
-		return GetState(ctx, reader, stateID, cfg.Commitment())
+		getAccountReader := func() (client.AccountReader, error) { return getReader() }
```
aalu1418: please remove this - similar to NewTransmissionsCache

aalu1418 (changes requested):

aalu1418 (approved):

dhaidashenko (approved):

skippaDaBitFlippa (approved):

PR #896, opened 2024-10-18
[NONEVM-714] [solana] - Add multiple blocks aggregation functionality to Block History Estimator 

Farber98:
### Description:
This PR Introduces the `MultipleBlocksEstimator` that **tries** to get the `desiredBlockCount` most recent blocks to determine a more stable compute unit price. 

### How It Works:

- **Configuration:** `BlockHistorySize` is used to configure this estimator.

- **Fetch Recent Blocks:** Tries to retrieve specified number of the latest confirmed blocks from latest slots based on the configured `BlockHistorySize`. Not all slots have a block.

- **Concurrent Processing:**: Fetches each block concurrently while limiting the number of simultaneous operations to prevent rate limiting.

- **Extract and Aggregate Fees:** Parses each block to extract compute unit prices from transactions. Calculates the median compute unit price for each block. 

- **Determine Overall Average:** Aggregates the median prices from all fetched blocks. Computes the overall average to set as the current base compute unit price, ensuring it stays within predefined minimum and maximum config limits.

### Merge together with: 
- https://github.com/smartcontractkit/chainlink/pull/15029


Farber98:
SonarQube thinks some account addresses in the json file are IBM API Keys. That's why the low score.

image

pkg/solana/fees/block_history.go
```diff
@@ -99,39 +121,161 @@ func (bhe *blockHistoryEstimator) readRawPrice() uint64 {
...
+}
+
+func (bhe *blockHistoryEstimator) calculatePriceFromMultipleBlocks(ctx context.Context, desiredBlockCount uint64) error {
+	if desiredBlockCount == 0 {
+		return fmt.Errorf("desiredBlockCount is zero")
+	}
+
+	if desiredBlockCount > MAX_BLOCK_HISTORY_DEPTH {
+		// Limit the desired block count to maxblockHistoryDepth
+		desiredBlockCount = MAX_BLOCK_HISTORY_DEPTH
```
silaslenihan: Maybe we should log a warning here

Farber98: The thing is this code executes once every `BlockHistoryPollPeriod`, so it may become really noisy. 

Wdyt about refactoring and checking/logging during initialization?

pkg/solana/config/config.go
```diff
@@ -31,6 +31,7 @@ var defaultConfigSet = Chain{
 	BlockHistoryPollPeriod:   config.MustNewDuration(5 * time.Second),
 	ComputeUnitLimitDefault:  ptr(uint32(200_000)), // set to 0 to disable adding compute unit limit
 	EstimateComputeUnitLimit: ptr(false),           // set to false to disable compute unit limit estimation
+	BlockHistoryDepth:        ptr(uint64(20)),      // number of blocks to fetch for multiple blocks fee estimation
```
amit-momin: Just to keep things consistent with EVM where we can, could we rename this config to `BlockHistorySize` instead?

pkg/solana/fees/block_history.go
```diff
@@ -16,6 +16,26 @@ import (
...
+func (em EstimationMethod) String() string {
+	switch em {
+	case LatestBlockEstimator:
+		return "LatestBlock"
+	case MultipleBlocksEstimator:
+		return "MultipleBlocks"
+	default:
+		return "UnknownMethod"
+	}
+}
```
amit-momin: I don't think we need to introduce an estimation method here. I think we can infer the same thing with using the `BlockHistorySize` config. Like if `BlockHistorySize=1` then we should do the latest block logic, otherwise do the multiple blocks logic

pkg/solana/fees/block_history.go
amit-momin: I think it might be ok not setting a max block history unless during testing we find some sort of definitive upper limit. Otherwise, we could make a recommendation to CCIP if they want to set a high value to not go over a certain value but we shouldn't make it a hard requirement in the code.

amit-momin: > Wdyt about refactoring and checking/logging during initialization?

If we do want to keep an upper bound, I think it's better to move this to initialization. But instead of just logging a warning, we should fail initialization so the person running the node can update the configs accordingly.

pkg/solana/config/config.go
amit-momin: Also to keep existing behavior of the BHE, we should maybe set this default to 1 to mimic latest block behavior.

pkg/solana/fees/multiple_blocks_data.json
amit-momin, line 1: I think just for the unit tests we probably shouldn't commit such a big file. Feel like we should be able to get away with mocking a few blocks in code in `block_history_test.go` instead.

pkg/solana/fees/block_history_test.go
```diff
@@ -88,9 +90,197 @@ func TestBlockHistoryEstimator(t *testing.T) {
...
+			Return(&testSlotsResult, nil).Times(1)
+		for i, slot := range testSlots {
+			rw.On("GetBlock", mock.Anything, slot).
+				Return(testBlocks[i], nil).Times(1)
+		}
+		lgr, logs := logger.TestObserved(t, zapcore.DebugLevel)
+		estimator := initializeEstimator(t, rwLoader, cfg, lgr, tests.Context(t))
+
+		// Calculate expected median price from all the blocks and check estimated price
+		tests.AssertLogEventually(t, logs, "BlockHistoryEstimator: updated")
```
amit-momin: I know we use this in other parts of our code base but I think we want to minimize our use of eventually. Think it leads to potential flakiness. Could we instead make `calculatePrice` price a public method and call that directly in the test? Then we can be sure the price is set and check `BaseComputeUnitPrice` after.

pkg/solana/fees/block_history.go
Farber98: I'll remove it for now. Let's see what happens when testing with different values

I got a limited around 20slots/5s for the public rpc endpoint for devnet

I'm not sure what the limit will be in terms of mainnet

pkg/solana/fees/multiple_blocks_data.json
Farber98: Did it that way to be consistent on how the previous estimator was tested: https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/fees/blockdata.json.

I can try mocking or making the file smaller so we have small blocks with useful data only

amit-momin: Ah I see, that's ok then. I was thinking of how we test it for EVM. We can maybe keep this as something we could clean up later if we have time and maybe even do the same for `blockdata.json`

pkg/solana/fees/block_history.go
```diff
@@ -99,28 +99,39 @@ func (bhe *blockHistoryEstimator) readRawPrice() uint64 {
 }
 
 func (bhe *blockHistoryEstimator) calculatePrice(ctx context.Context) error {
+	switch {
+	case bhe.cfg.BlockHistorySize() == 1:
+		return bhe.calculatePriceFromLatestBlock(ctx)
+	case bhe.cfg.BlockHistorySize() > 1:
+		return bhe.calculatePriceFromMultipleBlocks(ctx, bhe.cfg.BlockHistorySize())
+	default:
+		return fmt.Errorf("invalid block history depth: %d", bhe.cfg.BlockHistorySize())
```
amit-momin: I think it might be better to validate the `BlockHistorySize` during initialization so we fail fast and problems with configs can be noticed quicker. So I guess just a `if BlockHistorySize < 1 { error }`

pkg/solana/fees/block_history.go
```diff
@@ -133,5 +144,105 @@ func (bhe *blockHistoryEstimator) calculatePrice(ctx context.Context) error {
...
+		wg.Add(1)
+		go func(s uint64) {
+			defer wg.Done()
+			semaphore <- struct{}{}
+			defer func() { <-semaphore }()
+
+			// Fetch the block details
+			block, errGetBlock := c.GetBlock(ctx, s)
+			if errGetBlock != nil || block == nil {
+				// Failed to get block at slot || no block found at slot: skip.
```
amit-momin: Might be worth logging an error here if `errGetBlock != nil` since it could help surface an RPC issue to the user. I'm assuming we wouldn't error in normal use cases so this should flood the log unless you've noticed this endpoint be flakey.

pkg/solana/fees/block_history.go
```diff
@@ -133,5 +144,105 @@ func (bhe *blockHistoryEstimator) calculatePrice(ctx context.Context) error {
...
+				return
+			}
+
+			// Parse the block to extract compute unit prices
+			feeData, errParseBlock := ParseBlock(block)
+			if errParseBlock != nil {
+				// Failed to parse block at slot: skip.
```
amit-momin: Same here. Might be worth adding an error log in case there's some compatibility issue that comes up so we know what's breaking.

pkg/solana/fees/block_history.go
```diff
@@ -133,5 +144,105 @@ func (bhe *blockHistoryEstimator) calculatePrice(ctx context.Context) error {
...
+				return
+			}
+
+			// Calculate the median compute unit price for the block
+			blockMedian, errMedian := mathutil.Median(feeData.Prices...)
```
amit-momin, line 227: Looks like we take the median of the medians with this logic. Don't think there's anything wrong with that but if we're trying to mirror the EVM BHE's behavior, we may want to take the median of all aggregated transactions across all of the desired blocks. Instead of taking the median here, would it be a lot if we just aggregate all of the `feeData.Prices` into one big list? If there are a ton of transactions per block and this would be a lot, we could keep it the way it is.

pkg/solana/fees/block_history_test.go
```diff
@@ -15,82 +16,442 @@ import (
...
 	cfg := cfgmock.NewConfig(t)
-	cfg.On("ComputeUnitPriceDefault").Return(uint64(100))
-	cfg.On("ComputeUnitPriceMin").Return(min)
-	cfg.On("ComputeUnitPriceMax").Return(max)
-	cfg.On("BlockHistoryPollPeriod").Return(100 * time.Millisecond)
-	lgr, logs := logger.TestObserved(t, zapcore.DebugLevel)
+	setupConfigMock(cfg, defaultPrice, min, max, pollPeriod, invalidDepth)
+	lgr, _ := logger.TestObserved(t, zapcore.DebugLevel)
 	ctx := tests.Context(t)
+	estimator := initializeEstimator(t, rwLoader, cfg, lgr, ctx)
```
amit-momin: I think since you've removed the eventually's I don't think we need an observed logger anymore. Think you can just do something like this.
```suggestion
	ctx := tests.Context(t)
	estimator := initializeEstimator(t, rwLoader, cfg, logger.Test(t), ctx)
```

pkg/solana/fees/block_history_test.go
```diff
@@ -15,82 +16,442 @@ import (
...
+
+	t.Run("Successful Estimation", func(t *testing.T) {
+		// Setup
+		rw := clientmock.NewReaderWriter(t)
+		rwLoader := utils.NewLazyLoad(func() (client.ReaderWriter, error) {
+			return rw, nil
+		})
+		cfg := cfgmock.NewConfig(t)
+		setupConfigMock(cfg, defaultPrice, min, max, pollPeriod, depth)
+		rw.On("GetLatestBlock", mock.Anything).Return(lastBlock, nil)
```
amit-momin: To reduce some of the duplication here, could we move this mock ReaderWriter to the parent test with the mock for `rw.On("GetLatestBlock", mock.Anything).Return(lastBlock, nil)`. Seems like this is the same across a few tests. It's ok if you mock new ReaderWriters for the RPC failure tests further down though

pkg/solana/fees/block_history_test.go
```diff
@@ -15,82 +16,442 @@ import (
...
+		cfg := cfgmock.NewConfig(t)
+		ctx := tests.Context(t)
+		setupConfigMock(cfg, defaultPrice, min, max, pollPeriod, depth)
+		rw.On("SlotHeight", mock.Anything).Return(testSlots[len(testSlots)-1], nil)
+		rw.On("GetBlocksWithLimit", mock.Anything, mock.Anything, mock.Anything).
+			Return(&testSlotsResult, nil)
+		for i, slot := range testSlots {
+			rw.On("GetBlock", mock.Anything, slot).
+				Return(testBlocks[i], nil)
+		}
```
amit-momin: Same as the latest block BHE tests, could we move this ReaderWriter mock to the parent test? And just create new mocks for the tests for RPC failures?

pkg/solana/fees/block_history.go
Farber98: Still need to discuss if we use:
- median of medians
- average of medians

Farber98: We decided to go with:
- average of medians

pkg/solana/fees/block_history.go
```diff
@@ -133,5 +146,113 @@ func (bhe *blockHistoryEstimator) calculatePrice(ctx context.Context) error {
...
+
+			// Fetch the block details
+			block, errGetBlock := c.GetBlock(ctx, s)
+			if errGetBlock != nil || block == nil {
+				// Failed to get block at slot || no block found at slot
+				if errGetBlock != nil {
+					bhe.lgr.Errorw("BlockHistoryEstimator: failed to get block", "slot", s, "error", errGetBlock)
+				}
+				return
+			}
```
amit-momin: nit: maybe we can split this into 2 different if's to avoid extra nesting. We can also log an error for the other scenario too
```suggestion
			if errGetBlock != nil {
				// Failed to get block at slot
			        bhe.lgr.Errorw("BlockHistoryEstimator: failed to get block", "slot", s, "error", errGetBlock)
				return
			}
			if block == nil {
			  // No block found at slot
			  bhe.lgr.Errorw("BlockHistoryEstimator: failed to get block at slot", "slot", s)
			  return
			}
```

pkg/solana/fees/block_history_test.go
```diff
@@ -11,86 +12,366 @@ import (
...
+		cfg := cfgmock.NewConfig(t)
+		setupConfigMock(cfg, defaultPrice, min, max, pollPeriod, depth)
+		estimator := initializeEstimator(tests.Context(t), t, rwFailLoader, cfg, logger.Test(t))
+
+		// Call calculatePrice and expect an error
+		// Ensure the price remains unchanged
+		require.Error(t, estimator.calculatePrice(tests.Context(t)), "Expected error when getting client fails")
+		cfg.On("ComputeUnitPriceMax").Return(max)
+		assert.Equal(t, defaultPrice, estimator.BaseComputeUnitPrice(), "Price should remain at default when client fails")
+	})
```
amit-momin, line 171: Could you add a similar test to the multiple block `no compute unit prices collected` for latest block tests? Looking at the current code, we're relying on the `mathutil.Median` to throw an error if the slice is empty but think it might be better to have an explicit check somewhere in between [here](https://github.com/smartcontractkit/chainlink-solana/blob/1cf95ab5742b081f1d35e6c7c7ba50fc02679851/pkg/solana/fees/block_history.go#L115-L124) with its own unique error string if no prices are returned.

pkg/solana/fees/block_history.go
Farber98: Thought the `block == nil` case would be too noisy as not every slot has a block. Just tested with my custom script across 100 txes and didn't return any log so I'll include it

amit-momin: Actually that's a fair point. You can skip the log for `block == nil` but we should still split it into different if's

pkg/solana/fees/block_history_test.go
```diff
@@ -11,86 +12,366 @@ import (
...
-	require.NoError(t, err)
-	blockRes := &rpc.GetBlockResult{}
-	require.NoError(t, json.Unmarshal(testBlockData, blockRes))
+	rw.On("GetLatestBlock", mock.Anything).Return(lastBlock, nil)
+
+	t.Run("Successful Estimation", func(t *testing.T) {
+		// Setup
+		cfg := cfgmock.NewConfig(t)
+		setupConfigMock(cfg, defaultPrice, min, max, pollPeriod, depth)
+		estimator := initializeEstimator(tests.Context(t), t, rwLoader, cfg, logger.Test(t))
```
amit-momin: Sorry one more nit. Could you initialize a context at the top like `ctx := tests.Context(t)` and use that in all of the calls in these sub tests? Same for the multiple blocks tests?
```suggestion
		estimator := initializeEstimator(ctx, t, rwLoader, cfg, logger.Test(t))
```

pkg/solana/txm/txm.go
```diff
@@ -170,9 +170,10 @@ func (txm *Txm) sendWithRetry(ctx context.Context, baseTx solanaGo.Transaction,
 
 	// base compute unit price should only be calculated once
 	// prevent underlying base changing when bumping (could occur with RPC based estimation)
+	baseFee := txm.fee.BaseComputeUnitPrice()
 	getFee := func(count int) fees.ComputeUnitPrice {
 		fee := fees.CalculateFee(
-			txcfg.BaseComputeUnitPrice,
+			baseFee,
```
Farber98: Eyes 👀. Added this change here so we avoid merging 2 PRs: https://github.com/smartcontractkit/chainlink-solana/pull/892

pkg/solana/txm/txm_internal_test.go
```diff
@@ -680,6 +680,8 @@ func TestTxm_Enqueue(t *testing.T) {
 	lggr := logger.Test(t)
 	cfg := config.NewDefault()
 	mc := mocks.NewReaderWriter(t)
+	mc.On("GetLatestBlock", mock.Anything).Return(&rpc.GetBlockResult{}, nil).Maybe()
```
Farber98: this will be needed when we change the configs. Detected when running Soak Tests. One is needed for `LatestBlock` and the other one for `MultipleBlocks`. Same idea, adding changes here so we merge only 1 PR

pkg/solana/txm/txm.go
amit-momin: Sent a message async with a better explanation but don't think we need this change

pkg/solana/txm/txm_internal_test.go
amit-momin: We'll actually keep the default estimator as `fixed` to avoid any changes in the existing behavior. Would we really need these?

Farber98: We will need them to be able to set the `FeeEstimatorMode = "blockhistory"` without tests failing: 
- When `BlockHistorySize = 1`, we [need](https://github.com/smartcontractkit/chainlink-solana/blob/213f7bc4d55149410327edc95c6b421d0458775f/pkg/solana/fees/block_history.go#L124) `GetLatestBlock` mock
- When `BlockHistorySize > 1`, we [need](https://github.com/smartcontractkit/chainlink-solana/blob/213f7bc4d55149410327edc95c6b421d0458775f/pkg/solana/fees/block_history.go#L167) `SlotHeight` mock

amit-momin: Isn't this test just using the default fixed price estimator though so don't think these mocks will be used. We could add these if we were explicitly adding tests for BHE but don't think we are?

Farber98: ohh now I get it. Yes, you are right

pkg/solana/txm/txm.go
Farber98: reverted

amit-momin (approved):

silaslenihan (approved):

PR #909, opened 2024-10-28
Updated Solana TXM in-memory storage layer to track transactions states

amit-momin:
[NONEVM-708](https://smartcontract-it.atlassian.net/browse/NONEVM-708)

- Added internal transaction states that largely map to Solana tx states: `Broadcasted`, `Processed`, `Confirmed`, `Finalized`, `Errored`
- Added 3 state maps
  - broadcasted/processed map: the map where `Broadcasted` and `Processed` transactions are stored
    - transactions in these states continue to retry and bump
  - confirmed map: the map where `Confirmed` transactions are stored
    - transaction in this state stop retrying and bumping but continue to track state changes till finalization
  - finalized/errored map: the map where `Finalized` and `Errored` transactions are stored
    - transactions in these states stop retrying/bumping and requesting the RPC for state changes
    - transactions are stored here for a configurable amount of time to allow their statuses to be queryable for external callers
- Added new `TxRetentionTimeout` config that determines how long finalized/errored transactions are held onto before being reaped
- Introduced a new prom metric `solana_txm_tx_finalized` to track the amount of finalized transactions

[NONEVM-708]: https://smartcontract-it.atlassian.net/browse/NONEVM-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/txm/pendingtx.go
```diff
@@ -84,48 +128,62 @@ func (c *pendingTxContext) Add(id uuid.UUID, sig solana.Signature) error {
 	if _, exists := c.sigToID[sig]; exists {
 		return errors.New("signature already exists")
 	}
```
silaslenihan: Why do we need to check this twice? (applies to other occurances as well)

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
-	// upgrade to write lock if sig does not exist
+	// upgrade to write lock if sig or id do not exist
 	c.lock.Lock()
 	defer c.lock.Unlock()
 	if _, exists := c.sigToID[sig]; exists {
-		return uuid.UUID{}, errors.New("signature already exists")
+		return errors.New("signature already exists")
+	}
+	if _, exists := c.broadcastedTxs[tx.id]; exists {
+		return errors.New("tx id already exists")
```
```suggestion
		return errors.New("tx id already exists in broadcasted map")
```

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/gagliardetto/solana-go"
-	"github.com/google/uuid"
 	"golang.org/x/exp/maps"
 )
 
+var ErrAlreadyInExpectedState = errors.New("transaction already in expected state")
```
silaslenihan: Maybe add more granular erroring?
```suggestion
func ErrAlreadyInExpectedState(state TxState) error {
    return errors.New(fmt.Sprintf("transaction already in expected state: %s", state.String()))
}
```

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +196,270 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
...
+		tx = tempTx
+	}
+	if tempTx, confirmedExists = c.confirmedTxs[id]; confirmedExists {
+		tx = tempTx
+	}
+	if !broadcastedExists && !confirmedExists {
+		return id, errors.New("id does not exist in broadcasted or confirmed map")
+	}
+	// call cancel func + remove from map to stop the retry/bumping cycle for this transaction
+	// cancel is expected to be called and removed when tx is confirmed but checked here too in case state is skipped
```
silaslenihan: In what case would that happen?

pkg/solana/txm/txm.go
```diff
@@ -554,6 +603,27 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	return nil
 }
 
+// GetTransactionStatus translates internal TXM transaction statuses to chainlink common statuses
+func (txm *Txm) GetTransactionStatus(ctx context.Context, transactionID string) (commontypes.TransactionStatus, error) {
+	state, err := txm.txs.GetTxState(transactionID)
+	if err != nil {
+		return commontypes.Unknown, fmt.Errorf("failed to find transaction with id: %s", transactionID)
```
silaslenihan:
```suggestion
		return commontypes.Unknown, fmt.Errorf("failed to find transaction with id %s: %w", transactionID, err)
```

pkg/solana/txm/pendingtx.go
amit-momin: Think it's an optimization to do validations with just a read lock since state changes could be frequent depending on transaction throughput. But we have to recheck to be safe when switching to the write lock. I also wanted to maintain the flow from the existing in-memory storage methods.

pkg/solana/txm/pendingtx.go
amit-momin: I added this just in case there's some transient issue with RPCs and/or maybe a slow confirmation poll period config. There could be a scenario where we read the state of a tx as `Processed` in one run loop but for one reason or another the next time we check it's already `Finalized`.

pkg/solana/txm/pendingtx.go
silaslenihan: Ok, yeah that makes sense

pkg/solana/txm/pendingtx.go
amit-momin: This was a good suggestion but think this is causing some problems with `errors.Is` for some reason. I think it's ok if we skip appending the state here. We use this predefined error to skip logging actually so the message itself probably isn't as important.

pkg/solana/config/config.go
```diff
@@ -17,7 +17,8 @@ var defaultConfigSet = Chain{
 	OCR2CacheTTL:        config.MustNewDuration(time.Minute),            // stale cache deadline
 	TxTimeout:           config.MustNewDuration(time.Minute),            // timeout for send tx method in client
 	TxRetryTimeout:      config.MustNewDuration(10 * time.Second),       // duration for tx rebroadcasting to RPC node
-	TxConfirmTimeout:    config.MustNewDuration(30 * time.Second),       // duration before discarding tx as unconfirmed
+	TxConfirmTimeout:    config.MustNewDuration(30 * time.Second),       // duration before discarding tx as unconfirmed. Set to 0 to disable discarding tx.
```
Farber98, line 20: - What would be the use case of disabling discarding a tx? 
- Could we have a case where we stack confirmed transactions that don't go through and we are unable to drop them? Could this cause a memory leak and should be worry about it?

Farber98, line 20: - What would be the use case of disabling discarding a tx?
- Could we have a case where we stack confirmed transactions that don't go through and we are unable to drop them? 
- Could this cause a memory leak and should we worry about it?

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
 
-func (c *pendingTxContext) New(sig solana.Signature, cancel context.CancelFunc) (uuid.UUID, error) {
-	// validate signature does not exist
+func (c *pendingTxContext) New(tx pendingTx, sig solana.Signature, cancel context.CancelFunc) error {
 	c.lock.RLock()
+	// validate signature does not exist
 	if _, exists := c.sigToID[sig]; exists {
 		c.lock.RUnlock()
-		return uuid.UUID{}, errors.New("signature already exists")
+		return errors.New("signature already exists")
```
Farber98: nit: could define a constant for errors that are used multiple times across the file:
- signature already exists
- tx id already exists in broadcasted map
- id does not exist in broadcasted map
- id does not exist in broadcasted or confirmed map

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
+	tx.state = Broadcasted
+	// save to the broadcasted map since transaction was just broadcasted
+	c.broadcastedTxs[tx.id] = tx
+	return nil
 }
 
-func (c *pendingTxContext) Add(id uuid.UUID, sig solana.Signature) error {
-	// already exists
+func (c *pendingTxContext) AddSignature(id string, sig solana.Signature) error {
 	c.lock.RLock()
```
Farber98: nit: would make sense to encapsulate locking logic into functions? It may reduce a couple of lines and make the locked portion more explicit: 

Something like: 

```
func (c *pendingTxContext) withWriteLock(fn func()) {
    c.lock.Lock()
    defer c.lock.Unlock()
    fn()
}

func (c *pendingTxContext) withReadLock(fn func()) {
    c.lock.RLock()
    defer c.lock.RUnlock()
    fn()
}

```
that could be used as:

```
c.withReadLock(func() {
    // Perform read operations
})

c.withWriteLock(func() {
    // Perform write operations
})
```

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +196,270 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
...
+	if !broadcastedExists && !confirmedExists {
+		c.lock.RUnlock()
+		return id, errors.New("id does not exist in broadcasted or confirmed map")
+	}
+	c.lock.RUnlock()
+
+	// upgrade to write lock if id exists
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if _, exists := c.sigToID[sig]; !exists {
```
Farber98: nit: would make sense to abstract these checks that are used across multiple places?

```
func (c *pendingTxContext) signatureExists(sig solana.Signature) bool {
    _, exists := c.sigToID[sig]
    return exists
}

func (c *pendingTxContext) idExists(id string) bool {
    _, exists := c.broadcastedTxs[id]
    return exists
}
```

and then use them as:
```
if c.signatureExists(sig) {
    return ErrSignatureExists
}
if !c.idExists(tx.id) {
    return ErrIDNotExists
}
```

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
+	// Expired returns whether or not confirmation timeout amount of time has passed since creation
+	Expired(sig solana.Signature, confirmationTimeout time.Duration) bool
+	// OnProcessed marks transactions as Processed
+	OnProcessed(sig solana.Signature) (string, error)
+	// OnConfirmed marks transaction as Confirmed and moves it from broadcast map to confirmed map
+	OnConfirmed(sig solana.Signature) (string, error)
+	// OnFinalized marks transaction as Finalized, moves it from confirmed map to finalized map, removes signatures from signature map to stop confirmation checks
+	OnFinalized(sig solana.Signature, retentionTimeout time.Duration) (string, error)
+	// OnError marks transaction as errored, matches err type using enum, moves it from confirmed map to finalized map, removes signatures from signature map to stop confirmation checks
+	OnError(sig solana.Signature, retentionTimeout time.Duration, errType int) string
```
Farber98: why not returning an error here?

pkg/solana/txm/pendingtx.go
```diff
@@ -3,78 +3,122 @@ package txm
...
 type PendingTxContext interface {
-	New(sig solana.Signature, cancel context.CancelFunc) (uuid.UUID, error)
-	Add(id uuid.UUID, sig solana.Signature) error
-	Remove(sig solana.Signature) uuid.UUID
+	// New adds a new tranasction in Broadcasted state to the storage
+	New(msg pendingTx, sig solana.Signature, cancel context.CancelFunc) error
+	// AddSignature adds a new signature for an existing transaction ID
+	AddSignature(id string, sig solana.Signature) error
+	// Remove removes transaction and related signatures from storage if not in finalized or errored state
+	Remove(sig solana.Signature) string
```
Farber98: why not returning an error here?

Farber98 (commented):
Looks really good! Like the way you structured the solution. Leaving some nits/questions

pkg/solana/txm/pendingtx.go
Farber98: The optimization of handling a read lock and then a write lock is considerable? Or could we just use a write lock and make the things simpler

pkg/solana/config/config.go
amit-momin: - I made this disable-able for the new fee bumping strategy actually. Dropping transactions actually only really works for Data Feeds where there's a follow up update anyways but for CCIP and other products we would want to get transactions through to finalization
- What do you mean by stacking confirmed transactions? I should mention that we probably only expect to disable this if we're using the new fee bumping strategy where we wait for expiration before sending a new tx. So we would have another mechanism to drop transactions before sending a new one so txs wouldn't stack.
- Glad you're thinking about that. I think if we can reliably pick up on transaction expiration/re-orgs and handle those cases then we wouldn't have to worry about memory leaks

pkg/solana/txm/pendingtx.go
Farber98: Was curious about this. It seems the problem is the following:

- Every call to `ErrAlreadyInExpectedState(state)` creates a new error instance
- The `errors.Is` function checks if any error in the chain matches the target error instance, not the message. 
- Since each call generates a distinct error instance, `errors.Is(err, ErrAlreadyInExpectedState(state))` will always return false because they are different instances in memory.

pkg/solana/config/config.go
Farber98: Thanks for the clarification!

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +196,270 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
...
+		if time.Now().After(tx.retentionTs) {
+			expiredIDs = append(expiredIDs, id)
+		}
+	}
+	c.lock.RUnlock()
+
+	// upgrade to write lock
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	for _, id := range expiredIDs {
```
Farber98, line 475: - Should we worry about this being unbounded if the number of txes is too large? 
- Should we add an upper bound? 
- Should we add a note to the TxReapInterval config to make this visible?

pkg/solana/txm/pendingtx.go
amit-momin: This was an optimization that was built into the existing TXM's in-memory layer so I didn't want to cause any regressions if it was decided it was helpful before.

pkg/solana/txm/pendingtx.go
amit-momin: Right, I was thinking it was just a string match under the hood until running into this issue

silaslenihan: We could potentially do a string match for that check but it probably isn't necessary

amit-momin: Right, we could but think it's better to skip since we don't even log this err on the TXM side. I actually added this predefined error to avoid noise in the logs.

pkg/solana/txm/pendingtx.go
amit-momin: Good catch, I initialize the slice now. Also gave this a second thought from your other comment and changed this method to only use the write lock. Realized we probably shouldn't have a gap between read and write here. But what kind of note did you think we needed for TxReapInterval?

pkg/solana/txm/pendingtx.go
amit-momin: Tbh don't think this method is being used currently but left it from the existing code since it seemed useful when we start detecting transaction expiration. I've updated to return an error though to keep things consistent

pkg/solana/txm/pendingtx.go
amit-momin: Good point, added an error for consistency

pkg/solana/txm/pendingtx.go
amit-momin: Good idea, I added some more predefined errors in the latest commit. I ended up consolidating the `id does not exist...` ones into a single error. Think the stack trace with the error can be used to determine which map(s) it was missing from to keep things simpler.

pkg/solana/txm/pendingtx.go
amit-momin: This could help in a few places but in others we also need the `id` from this check too where this wouldn't work. I don't think it's worth abstracting since we just replace the map access with a method call so we aren't really saving any lines either.

pkg/solana/txm/pendingtx.go
amit-momin: I like this suggestion. I updated the code to use this in the latest commit.

pkg/solana/txm/pendingtx.go
Farber98: Great! My concern was that if the number of transactions we need to delete is large and frequent, we may be locking a lot and will affect the actual processing of the transactions

But thinking it again, better not to over engineer too soon and check later if this is a thing. In that case we could:
- Limit number of deletions on each pass as a simple approach
- Analyze if having different locks makes the processing independent of the reaping

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +210,283 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
 }
 
 // Expired returns if the timeout for trying to confirm a signature has been reached
-func (c *pendingTxContext) Expired(sig solana.Signature, lifespan time.Duration) bool {
+func (c *pendingTxContext) Expired(sig solana.Signature, confirmationTimeout time.Duration) bool {
```
Farber98, line 213: should we use the new lock methods for consistency here?

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +210,283 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
...
+		// update tx state to Errored
+		tx.state = Errored
+		// move transaction from broadcasted to error map
+		c.finalizedErroredTxs[id] = tx
+		return id, nil
+	})
+}
+
+func (c *pendingTxContext) GetTxState(id string) (TxState, error) {
+	c.lock.RLock()
```
Farber98, line 451: should we use the new lock methods for consistency here?

pkg/solana/txm/pendingtx.go
```diff
@@ -138,28 +210,283 @@ func (c *pendingTxContext) ListAll() []solana.Signature {
...
+		return tx.state, nil
+	}
+	return NotFound, fmt.Errorf("failed to find transaction for id: %s", id)
 }
 
-func (c *pendingTxContext) OnError(sig solana.Signature, _ int) uuid.UUID {
-	return c.Remove(sig)
+// TrimFinalizedErroredTxs deletes transactions from the finalized/errored map and the allTxs map after the retention period has passed
+func (c *pendingTxContext) TrimFinalizedErroredTxs() {
+	c.lock.Lock()
```
Farber98, line 467: should we use the new lock methods for consistency here?

pkg/solana/txm/pendingtx_test.go
```diff
@@ -32,51 +30,818 @@ func TestPendingTxContext(t *testing.T) {
...
+	tx, exists = txs.confirmedTxs[msg.id]
+	require.False(t, exists)
+
+	// Check it does not exist in finalized map
+	tx, exists = txs.finalizedErroredTxs[msg.id]
+	require.False(t, exists)
+}
+
+func TestPendingTxContext_add_signature(t *testing.T) {
+	t.Parallel()
```
Farber98, line 105: Should we include checks on:
- Duplicate signature?
- Including a signature after confirmation?

pkg/solana/txm/pendingtx_test.go
```diff
@@ -32,51 +30,818 @@ func TestPendingTxContext(t *testing.T) {
...
+	tx, exists = txs.confirmedTxs[msg.id]
+	require.False(t, exists)
+
+	// Check finalized map
+	tx, exists = txs.finalizedErroredTxs[msg.id]
+	require.False(t, exists)
+}
+
+func TestPendingTxContext_on_broadcasted_processed(t *testing.T) {
+	t.Parallel()
```
Farber98, line 195: Should we include checks on:
- Processed to Finalized transitions as mentioned in [this comment](https://github.com/smartcontractkit/chainlink-solana/pull/909#discussion_r1829875519)?

pkg/solana/txm/pendingtx_test.go
```diff
@@ -15,15 +15,13 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
 )
```
Farber98, line 17: Should we add tests for `GetTxState`?

pkg/solana/txm/pendingtx.go
amit-momin: I had considered this but decided to only use the new methods for just the state transition methods. They have similar returns like id and error which some of these other methods don't. I could rework it to set fields outside of the `wthReadLock` method but thought it was easier to leave the methods that followed the normal locking standard.
```
c.lock.Lock()
defer c.lock.Unlock()
```

pkg/solana/txm/pendingtx_test.go
amit-momin: I actually have that test under the tests for the finalized map [here](https://github.com/smartcontractkit/chainlink-solana/pull/909/files#diff-ef0d3e62fe23dcbd48fbea8fd84e018d0bc97e17163e852f6e42ecda168caa4cR396)

Farber98 (approved):

silaslenihan (approved):

augustbleeds (approved):

PR #911, opened 2024-10-31
event loader with event watch and backfill

EasterTheBunny:
### Description

This PR introduces a component that monitors solana slot production and loads program log messages looking for emitted events. This component also provides a backfill feature to look for events in past blocks by program address.

Included is an integration test to a local solana validator as a smoke test and a benchmark test for watching slot production.

```
goos: darwin
goarch: arm64
pkg: github.com/smartcontractkit/chainlink-solana/pkg/solana/logpoller
cpu: Apple M1 Pro
BenchmarkEncodedLogCollector-10          1770250              6545 ns/op              6910 events/sec            2.762 rcp_calls/sec         662 B/op          3 allocs/op
PASS
ok      github.com/smartcontractkit/chainlink-solana/pkg/solana/logpoller       18.798s
```

[NONEVM-742](https://smartcontract-it.atlassian.net/browse/NONEVM-742)

[NONEVM-742]: https://smartcontract-it.atlassian.net/browse/NONEVM-742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/logpoller/parser.go
```diff
@@ -0,0 +1,194 @@
...
+					Style:  MessageStyleSuccess,
+					Text:   "Program returned success",
+				})
+
+				depth--
+			} else if strings.Contains(log, "failed") {
+				instLogs[len(instLogs)-1].Failed = true
+
+				idx := strings.Index(log, ": ") + 2
+				currText := fmt.Sprintf(`Program returned error: "%s"`, log[idx:])
```
reductionista: Interesting... this looks useful, but I didn't think it was in scope for what we were trying to support (reading success or failure of contract calls, as opposed to just indexing events).  Is this something CCIP mentioned they want, or is it just included here in case it comes in handy at some point?

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,265 @@
...
+	// ProcessEvent should be thread safe.
+	ProcessEvent(ProgramEvent) error
+}
+
+type RPCClient interface {
+	GetLatestBlockhash(ctx context.Context, commitment rpc.CommitmentType) (out *rpc.GetLatestBlockhashResult, err error)
+	GetBlocks(ctx context.Context, startSlot uint64, endSlot *uint64, commitment rpc.CommitmentType) (out rpc.BlocksResult, err error)
+	GetBlockWithOpts(context.Context, uint64, *rpc.GetBlockOpts) (*rpc.GetBlockResult, error)
+	GetSignaturesForAddressWithOpts(context.Context, solana.PublicKey, *rpc.GetSignaturesForAddressOpts) ([]*rpc.TransactionSignature, error)
+	GetTransaction(context.Context, solana.Signature, *rpc.GetTransactionOpts) (*rpc.GetTransactionResult, error)
```
reductionista: Did we end up not needing `GetTransaction`?  I was thinking we would fetch each of the transactions for a particular contract individually, but I guess the way you're doing it here might be more efficient in most cases.

Seems like a good place to start I guess.

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,265 @@
...
+			return []services.Service{c.workers}
+		},
+		Start: c.start,
+		Close: c.close,
+	}.NewServiceEngine(lggr)
+
+	return c
+}
+
+func (c *EncodedLogCollector) BackfillForAddress(ctx context.Context, address string, fromSlot uint64) error {
```
reductionista, line 87: I don't see anything calling this, have you tried testing it yet?

pkg/solana/logpoller/job.go
```diff
@@ -0,0 +1,142 @@
...
+			// exclude rewards
+			Rewards: &excludeRewards,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	detail := eventDetail{
+		blockNumber: *block.BlockHeight,
```
reductionista: Why is the block number a pointer, but the block hash is a value?

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,265 @@
...
+	"time"
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+type Parser interface {
```
reductionista: There's an unfortunate naming conflict here we'll have to resolve at some point. One of the files in the evm log poller is parser.go and that component is referred to as the parser in code and docs. It parses DSL into SQL queries.

Not a big deal, but one or both we'll need to rename slightly to disambiguate.

reductionista:
Code looking good so far!  Nice division of work into different jobs and workers, looks like it will be pretty efficient as long as the cpu usage isn't too high on the chainlink node or the solana validator.  I haven't run the tests myself yet, but looking forward to playing around with it.

pkg/solana/logpoller/parser.go
EasterTheBunny: This was just in case it is needed. I carried it over from my initial POC and didn't feel like losing it as a feature.

pkg/solana/logpoller/loader.go
EasterTheBunny: I started by reading all transactions but found that we SHOULD be able to get the transaction detail from the blocks themselves. This method, if it works for all RPCs, will _drastically_ reduce the number of RPC calls.

pkg/solana/logpoller/loader.go
EasterTheBunny: I don't have a direct test for it yet. It uses roughly the same internal mechanisms, but I still need a smoke test at minimum.

pkg/solana/logpoller/job.go
EasterTheBunny: I am not sure why this is. So far, the smoke test on a local test validator seems to work. I would be interested to see why the block height is a pointer. Maybe a nil check would be safer here anyway.

pkg/solana/logpoller/loader.go
EasterTheBunny: That makes sense to me. I can change it to whatever. Maybe `LogDataParser`?

pkg/solana/logpoller/loader.go
reductionista: Yeah, I think it should work for all RPC's. I'm not sure how much it will reduce the number of rpc calls though, because usually each contract wouldn't emit more than 1 event per block (that would be a very high rate of event emission!)... so as long as there are only a handful of contracts involved you'd only rarely get a collision where we needed more than one transaction from a block.  If there are a large number of contracts the node is handling, I could see it helping a lot... but if it's only a small number it would probably be about the same number of rpc calls... but you'd be requesting a lot of extra data, so the network bandwidth might be wasted (potentially even slowing down other requests?)

reductionista: I think what I'd lean toward is using one method for "backfills" and a different one for normal polling operation (where the node is already caught up, and there's only been one or a couple new blocks since the last polling tick).

One type of backfill we need is for a single filter when it gets initially registered.  The filter will include a `startingBlock` field to tell it how far to go back. If that is very far back, it could be a large number of events... but we only need them for this one contract. In that case, I think it's pretty clear that we should request them one transaction at a time. It would mean the difference between requesting say, a thousand transactions, and requesting a thousand blocks each with hundreds of transactions. I'd rather we just request 1000 tx's than 100,000 tx's even if the number of rpc calls is the same or slightly more for the 1000. What do you think?

For the other type of backfill, where the node is actually behind and it's trying to catch up on everything... it seems less clear, but I'd probably still lean toward doing it 1 tx at a time. We could also make an arbitrary decision based on how many filters are registered and how far back it's going.

pkg/solana/logpoller/loader_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (p *testBlockProducer) GetTransaction(_ context.Context, sig solana.Signature, _ *rpc.GetTransactionOpts) (*rpc.GetTransactionResult, error) {
+	p.b.Helper()
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	var msgs []string
+
+	p.count++
```
reductionista, line 333: I'm not sure I understand what `p.count` is supposed to represent.  Looks like it is incremented when some methods of `testBlockProducer` are called (eg. `GetTransaction`, `GetBlocks`) but not when other methods are called (eg. `GetSignatureForAddressWithOpts`)

reductionista: I guess `GetSignaturesForAddressWithOpts` is the only one that doesn't call it, so maybe just unimplemented. Still, does it make sense for us to be adding together counts from many different methods? Seems like it would be more useful to track them each individually.

pkg/solana/logpoller/loader_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+BenchLoop:
+	for i := 0; i < b.N; i++ {
+		select {
+		case <-ticker.C:
+			blockProducer.incrementSlot()
+		case <-ctx.Done():
+			continue BenchLoop
+		default:
+			blockProducer.makeEvent()
```
reductionista, line 199: This will just emit events rapid-fire one after another with no delay in between. Is that what was intended? Would be better I think to have more control over how many events are emitted in each slot instead of leaving it up to random chance.

pkg/solana/logpoller/loader_test.go
EasterTheBunny: `GetSignatureForAddressWithOpts` is for backfilling and that currently isn't part of the benchmark test. `p.count` increments the simulated rpc call count and is used to measure the call rate in the benchmark test.

pkg/solana/logpoller/loader_test.go
EasterTheBunny: The benchmark test isn't benchmarking end to end. It is only benchmarking what this component is actually doing, which is calling an RPC and parsing program logs as fast as possible with as little load on the RPC as possible. Maybe another good measurement would be the average delay between slot discovery and event 'production'. Since we are currently using polling to watch for slots, there will naturally be a delay between when a new slot is found in polling and when the log event is passed to the `ProgramEventProcessor`.

This benchmark test also does not take into account `ProgramEventProcessor` complexity/delay so some wide variance is expected, which is why it is passed around as an interface.

pkg/solana/logpoller/job.go
```diff
@@ -0,0 +1,145 @@
...
+		blockHash: block.Blockhash,
+	}
+
+	if block.BlockHeight != nil {
+		detail.blockNumber = *block.BlockHeight
+	}
+
+	for idx, trx := range block.Transactions {
+		detail.trxIdx = idx
+		if len(blockSigsOnly.Signatures)-1 <= idx {
```
dhaidashenko, line 119: Is it possible that a transaction does not have a corresponding signature?

pkg/solana/logpoller/job.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func messagesToEvents(messages []string, parser ProgramEventProcessor, detail eventDetail, chJobs chan Job) {
+	var logIdx uint
+	for _, outputs := range parseProgramLogs(messages) {
+		for _, event := range outputs.Events {
+			logIdx++
+
+			event.BlockNumber = detail.blockNumber
+			event.BlockHash = detail.blockHash.String()
```
dhaidashenko: Do we need to convert to string here? We plan to store hashes and addresses as bytes. Ideally, we should avoid redundant encoding/decoding

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+			timer.Stop()
+
+			return
+		case <-timer.C:
+			ctxB, cancel := context.WithTimeout(ctx, c.rpcTimeLimit)
+
+			// not to be run as a job, but as a blocking call
+			result, err := c.client.GetLatestBlockhash(ctxB, rpc.CommitmentFinalized)
+			if err != nil {
+				c.lggr.Info("failed to get latest blockhash", "err", err)
```
dhaidashenko: Should not this be an error/warn level?

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+
+			cancel()
+
+			// if the slot is not higher than the highest slot, skip it
+			if c.lastSentSlot.Load() >= result.Context.Slot {
+				continue
+			}
+
+			c.lastSentSlot.Store(result.Context.Slot)
+			c.chSlot <- result.Context.Slot
```
dhaidashenko: Should not we use "select context" here to ensure that we can stop gracefully if `runSlotProcessing` is slightly slower than `runSlotPolling`?

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+}
+
+func (c *EncodedLogCollector) runSlotProcessing(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case slot := <-c.chSlot:
+			if slot > c.highestSlot.Load() {
```
dhaidashenko: nit: it would be nice to invert if statements to reduce nesting

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+				c.highestSlot.Store(slot)
+
+				if !c.loadingBlocks.Load() {
+					c.loadingBlocks.Store(true)
+
+					// run routine to load blocks in slot range
+					go func(start, end uint64) {
```
dhaidashenko: nit: this may cause `runSlotProcessing` to wait for an additional `DefaultNextSlotPollingInterval` if the loading is done right after reading from `<-c.chSlot.` `mailbox.Mailbox` addresses that and slightly simplifies logic.

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+						defer c.loadingBlocks.Store(false)
+
+						if err := c.loadSlotBlocksRange(ctx, start, end); err != nil {
+							// TODO: probably log something here
+							// a retry will happen anyway on the next round of slots
+							// so the error is handled by doing nothing
+							c.lggr.Info("failed to load slot blocks range", "start", start, "end", end, "err", err)
```
dhaidashenko: Should not this be an error level to make it easier for us to identify cases when we constantly fail to load blocks range?

pkg/solana/logpoller/loader.go
dhaidashenko: If we decide to stick to spawning a goroutine we should use `c.engine.Go()` to prevent flaky tests panic when we log msg after test is done

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,258 @@
...
+		result rpc.BlocksResult
+		err    error
+	)
+
+	if result, err = c.client.GetBlocks(ctx, start, &end, rpc.CommitmentFinalized); err != nil {
+		return err
+	}
+
+	for _, block := range result {
+		c.chBlock <- block
```
dhaidashenko: We should abort write into the channel if `EncodedLogCollector` was closed

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+	if len(q.values) == 0 {
+		return getZero[T](), fmt.Errorf("no values to return")
+	}
+
+	val := q.values[0]
+
+	if len(q.values) > 1 {
+		q.values = q.values[1:]
+	} else {
+		q.values = []T{}
```
dhaidashenko, line 331: If queue processing is not fast enough, it will grow indefinitely. Shouldn't we use a linked list to avoid this?

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+				g.lggr.Infof("retrying job in %dms", wait/time.Millisecond)
+
+				retry.when = time.Now().Add(wait)
+			default:
+				wait := calculateExponentialBackoff(0)
+
+				g.lggr.Infof("retrying job %s in %dms", job, wait/time.Millisecond)
+
+				retry = retryableJob{
+					name: createRandomString(12),
```
dhaidashenko, line 216: Can't we use uint64 counter as a name generator?

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+		case <-g.chInputNotify:
+			g.processQueue(ctx)
+		case <-ctx.Done():
+			break Loop
+		}
+	}
+
+	// run the job queue one more time just in case some
+	// new work items snuck in
+	g.processQueue(ctx)
```
dhaidashenko: Why do we need to process new work if we've received a signal to stop?

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func (g *WorkerGroup) processQueue(ctx context.Context) {
+	for {
+		if g.queue.Len() == 0 {
+			break
+		}
+
+		if g.queue.Len() >= DefaultNotifyQueueDepth {
+			g.lggr.Infof("queue depth: %d", g.queue.Len())
```
dhaidashenko: nit: shouldn't this be a warning?

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+		if err := job.Run(ctx); err != nil {
+			w.Lggr.Infof("job %s failed with error; retrying: %s", job, err)
+			w.Retry <- job
+		}
+	}
+
+	// put itself back on the queue when done
+	select {
+	case w.Queue <- w:
+	default:
```
dhaidashenko, line 50: nit: on default we should log a critical error

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+		case job := <-g.chRetry:
+			var retry retryableJob
+
+			switch typedJob := job.(type) {
+			case retryableJob:
+				retry = typedJob
+				retry.count++
+
+				if retry.count > g.maxRetryCount {
+					g.lggr.Infof("job %s dropped after max retries", job)
```
dhaidashenko: nit: should we make it a warn error level?

pkg/solana/logpoller/worker.go
```diff
@@ -0,0 +1,372 @@
...
+
+					continue
+				}
+
+				wait := calculateExponentialBackoff(retry.count)
+				g.lggr.Infof("retrying job in %dms", wait/time.Millisecond)
```
dhaidashenko: nit: why do we manually convert to ms? wait.String() will pretty print it. Is it to create chart with loki? (in that case shouldn't it be logged as a separate field?)

pkg/solana/logpoller/worker.go
dhaidashenko: Do we need name? Can't we use slice instead of map and when we try to process retry just swap current and last slice element and trim at the end.
It should save us multiple allocations and the generation of a random string

pkg/solana/logpoller/worker_test.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+
+			output[idx] = 1
+
+			return nil
+		}})
+	}
+
+	// wait for the first 9 to finish and close the group
+	time.Sleep(100 * time.Millisecond)
```
dhaidashenko: nit: should we use waitGroup instead of time.Sleep & lock?  It can be flaky especially in CI as there is no guarantee that job will be processed before we close the group.

pkg/solana/logpoller/parser.go
```diff
@@ -0,0 +1,194 @@
...
+
+			if len(logDataMatches) <= 1 {
+				continue
+			}
+
+			instructionMatches := instructionMatcher.FindStringSubmatch(logDataMatches[1])
+
+			if len(instructionMatches) > 1 {
+				// is an event which should be followed by: Program data: (.*)
+				instLogs[len(instLogs)-1].Events = append(instLogs[len(instLogs)-1].Events, ProgramEvent{
```
dhaidashenko: Can't `len(instLogs)-1` be negative?

pkg/solana/logpoller/parser.go
```diff
@@ -0,0 +1,194 @@
...
+			instLogs[len(instLogs)-1].Truncated = true
+		} else {
+			matches := invokeMatcher.FindStringSubmatch(log)
+
+			if len(matches) > 0 {
+				if depth == 0 {
+					instLogs = append(instLogs, ProgramOutput{
+						ComputeUnits: 0,
+						Failed:       false,
+						Program:      matches[1],
```
dhaidashenko: Can't an external program produce a log that we'll treat as a log produced by our contract?

pkg/solana/logpoller/parser.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/*
+Program J1zQwrBNBngz26jRPNWsUSZMHJwBwpkoDitXRV95LdK4 invoke [1]
+Program log: Instruction: CreateLog
+Program data: HDQnaQjSWwkNAAAASGVsbG8sIFdvcmxkISoAAAAAAAAA // base64 encoded; borsh encoded with identifier
+Program J1zQwrBNBngz26jRPNWsUSZMHJwBwpkoDitXRV95LdK4 consumed 1477 of 200000 compute units
+Program J1zQwrBNBngz26jRPNWsUSZMHJwBwpkoDitXRV95LdK4 success
+*/
+func parseProgramLogs(logs []string) []ProgramOutput {
```
dhaidashenko: Should we have test coverage for this func? It seems to be very complex and can be broken easily.

dhaidashenko (commented):
Impressive work!

pkg/solana/logpoller/job.go
EasterTheBunny: The integration test using a test validator node implies that each transaction should have a signature. At this point I can only assume that live nodes should function the same way.

pkg/solana/logpoller/job.go
EasterTheBunny: I can change that. I was still in the mindset of chain agnostic types which isn't needed here.

pkg/solana/logpoller/loader.go
EasterTheBunny: I really don't get the whole levels thing. If something is an error it should either be bubbled up or handled. Logging is purely informational anyway for debug purposes. In this case, the function will run again on the polling interval so even if it's an error, it's handled. ¯\_(ツ)_/¯

pkg/solana/logpoller/loader.go
EasterTheBunny: That makes sense to me.

pkg/solana/logpoller/loader.go
EasterTheBunny: I think instead of assuming that one may or may not run before the other, I can structure the channel listener and sender such that if the listener isn't listening (because its processing), the sender drops the slot. This will allow polling to continue and the range listener will still pull an effective range regardless of whether it misses a slot or not.

pkg/solana/logpoller/loader.go
EasterTheBunny: I can change the log level, but conflating the purpose of logging. Getting granular on log levels is just giving people a reason to turn them off.

pkg/solana/logpoller/loader.go
EasterTheBunny: I'm not sure this is relevant since that channel doesn't ever get closed. It is always open within the life-cycle of `EncodedLogCollector`

pkg/solana/logpoller/worker.go
EasterTheBunny: I thought about limiting the queue length, but backfilling will increase it to an unknown amount. I did add log notifications about the queue depth which should give some indication of intense memory load.

pkg/solana/logpoller/worker.go
EasterTheBunny: I understand the uint64 counter which could save some memory footprint, but I don't understand using a slice to save allocations. The retry still needs the job and the time in which to retry. This could possibly be in a `map[time.Time]Job` where the key is the time when the job should be retried. At that point the random name isn't relevant.

pkg/solana/logpoller/worker.go
EasterTheBunny: This was leftover from something else. It probably isn't useful here.

pkg/solana/logpoller/worker.go
EasterTheBunny: Not sure this is necessary. The worker queue is always the size of the number of workers. The default in this case would only run if the queue of workers was full, which it can't be unless we have accidentally created too many workers.

pkg/solana/logpoller/worker.go
EasterTheBunny: This is just the way I tend to print durations. I can use pretty print just as well.

pkg/solana/logpoller/worker_test.go
EasterTheBunny: I could make it signal based where once a single job runs, the worker group is closed and the test continues.

pkg/solana/logpoller/parser.go
EasterTheBunny: It can if the program log output for solana changes. Program output always starts with an invoke message which is handled. The other way to do this would be to track a `lastLogIndex` value. It would be safer.

pkg/solana/logpoller/parser.go
EasterTheBunny: I'm not sure what case you're referencing. Can you expand on that?

pkg/solana/logpoller/parser.go
EasterTheBunny: Some real-world solana logs would be very helpful for this. I am simply basing the parser off the logs that I was able to create with the test contract and the research I did.

pkg/solana/logpoller/job.go
dhaidashenko: nit: should we throw an error `if len(block.Transactions) != len(blockSigsOnly.Signatures)`

EasterTheBunny: Seems fair

pkg/solana/logpoller/loader.go
dhaidashenko: Yes, closing is not the problem. My concern is that if EncodedLogCollector gets closed none is reading from `c.chBlock`, so this goroutine will block forever.

pkg/solana/logpoller/worker.go
dhaidashenko: In case when `len(q.values)` fluctuates in the range [1, 99] we will observe infinite growth of the underlying array, as we'll never reach len = 0.

pkg/solana/logpoller/worker.go
dhaidashenko: I was thinking that we can do something like
```go
var retries []Job
for i := 0; i < len(g.retrySlice); {
  retry := g.retrySlice[i]
  if time.Now().After(retry.when) {
    retries = append(retries, retry)
     g.retrySlice[i] =  g.retrySlice[len(g.retrySlice)-1] 
     g.retrySlice  = g.retrySlice[:len(g.retrySlice)-1] 
   } else {
     i++
   }
}
```

But now realized that it could cause dead lock.

pkg/solana/logpoller/parser.go
dhaidashenko: Let me rephrase it. Is there a guarantee that the Program's output is always base58 encoded?

pkg/solana/logpoller/worker.go
EasterTheBunny: I still don't understand. This is just taking the first element off the slice. If the length is greater than 1, we reset the slice using `[1:]` which returns all elements except the first (which is being returned). If the length is 0, the slice should be empty after the function returns. So we reach a length of 0 when the last value is popped from the queue.

It only pops one value at a time, but adds multiple. The array can grow indefinitely, yes, but that will depend on the job load when backfilling.

pkg/solana/logpoller/worker.go
reductionista: > unless we have accidentally created too many workers.

Seems better to log an error just in case, so we know what's happening

pkg/solana/logpoller/loader.go
```diff
@@ -249,7 +251,10 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
 	}
 
 	for _, block := range result {
-		c.chBlock <- block
+		select {
+		case <-ctx.Done():
```
reductionista, line 274: Shouldn't there be a `break` statement here to break out of the for loop if the context is cancelled?

reductionista: Oh, also in this line at the beginning of the `loadSlotBlockRange` method where you add an rpc timeout... I think we should rename the child context to something else and use that only for the call to `GetBlocks`... making use of the parent context here.

```
ctx, cancel := context.WithTimeout(ctx, c.rpcTimeLimit)
```

pkg/solana/logpoller/loader.go
```diff
@@ -214,7 +200,7 @@ func (c *EncodedLogCollector) runBlockProcessing(ctx context.Context) {
 				parser:     c.parser,
 				chJobs:     c.chJobs,
 			}); err != nil {
-				c.lggr.Infof("failed to add job to queue: %s", err)
+				c.lggr.Errorf("failed to add job to queue: %s", err)
```
reductionista, line 205: These %s's could be changed to %w's I think that's the preferred way to log errors

pkg/solana/logpoller/loader.go
```diff
@@ -227,12 +213,24 @@ func (c *EncodedLogCollector) runJobProcessing(ctx context.Context) {
...
 			}
 		}
 	}
 }
 
+func (c *EncodedLogCollector) loadRange(ctx context.Context, start, end uint64) {
+	if err := c.loadSlotBlocksRange(ctx, start, end); err != nil {
+		// a retry will happen anyway on the next round of slots
+		// so the error is handled by doing nothing
+		c.lggr.Error("failed to load slot blocks range", "start", start, "end", end, "err", err)
```
reductionista: I think this needs to be `c.lggr.Errorw` to work

pkg/solana/logpoller/worker.go
dhaidashenko: When we use slice, memory is only released when we pop the last element in the queue.
If we use a linked list memory is released right on every pop.
But I guess we can leave it as is.

pkg/solana/logpoller/loader.go
EasterTheBunny: That can provide assurance that the function exits immediately. I'll add it.

pkg/solana/logpoller/worker.go
dhaidashenko: @EasterTheBunny, what are your thoughts? Should not we at least drop string generation?

pkg/solana/logpoller/parser.go
dhaidashenko: Can't we have a test that uses logs that you've found? Just to ensure we lock implementation and it's more difficult to break something that was previously working.

pkg/solana/logpoller/loader.go
EasterTheBunny: The linter doesn't seem to care about variable name shadowing here, probably because it the first and only reset of `ctx` within the function and the scope doesn't change (making the meaning of `ctx` consistent for all the function). I don't think in this ... `context` ... variable name shadowing applies.

reductionista: I'm not worried about shadowing, I'm saying I don't agree with the functional behavior of the code--I'm suggesting that we use the child context only for the rpc call, and the parent context for the post processing.

The rpc timeout should not apply to the post processing after the rpc server has returned all the data.  There's no reason to abort the for loop unless the whole node is shutting down, since it doesn't depend on a network connection which might hang forever.

reductionista:
Presently the logs will just come back in random order which doesn't work. We need to make sure that the logs come back from the ingester in order of ascending order of `(block_number, tx_index, [index of log within tx])`

One way we could guarantee this would be to add a job type that listens to a channel of incoming logs from all of the jobs doing the parsing/matching and saves them to a priority queue (aka "heap") with the priority set as the order key above. Then either the same job or another job which pops them off the heap in order and sends them on their way to the destination channel.

We can split off the Borsch-decoding and saving of the logs to the db to another PR, but I think the ordering job(s) would fit in better with this one. What do you think?

EasterTheBunny:
> Presently the logs will just come back in random order which doesn't work. We need to make sure that the logs come back from the ingester in order of ascending order of `(block_number, tx_index, [index of log within tx])`
> 
> One way we could guarantee this would be to add a job type that listens to a channel of incoming logs from all of the jobs doing the parsing/matching and saves them to a priority queue (aka "heap") with the priority set as the order key above. Then either the same job or another job which pops them off the heap in order and sends them on their way to the destination channel.
> 
> We can split off the Borsch-decoding and saving of the logs to the db to another PR, but I think the ordering job(s) would fit in better with this one. What do you think?

I think the next layer that handles the decoding of the event data is the proper place for ordering. The reason is that many of the events will end up being discarded anyway. It doesn't seem valuable to order logs that will be discarded before even being decoded.

pkg/solana/logpoller/job.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	blockSigsOnly, err := j.client.GetBlockWithOpts(
+		ctx,
+		j.slotNumber,
+		// You can specify more options here:
```
reductionista: Is this comment referring to the possibility of us adding some more things to this code later, like a TODO? Or something else?

pkg/solana/logpoller/job.go
```diff
@@ -0,0 +1,149 @@
...
+	return fmt.Sprintf("getTransactionsFromBlockJob for block: %d", j.slotNumber)
+}
+
+func (j *getTransactionsFromBlockJob) Run(ctx context.Context) error {
+	var excludeRewards bool
+
+	block, err := j.client.GetBlockWithOpts(
+		ctx,
+		j.slotNumber,
+		// You can specify more options here:
```
reductionista: Same question here (see below)

pkg/solana/logpoller/loader_test.go
```diff
@@ -0,0 +1,340 @@
...
+	b.ReportAllocs()
+	b.ResetTimer()
+
+BenchLoop:
+	for i := 0; i < b.N; i++ {
+		select {
+		case <-ticker.C:
+			blockProducer.incrementSlot()
+		case <-ctx.Done():
+			continue BenchLoop
```
reductionista: Should this be break rather than continue? No point in continuing to loop over 0... b.N after the context is closed is there?

pkg/solana/logpoller/job.go
EasterTheBunny: Not a useful comment. I'll remove it

pkg/solana/logpoller/job.go
EasterTheBunny: Same answer. lol

pkg/solana/logpoller/loader_test.go
EasterTheBunny: I don't really think it matters because the context comes from the time the function is expected to run, but so is `b.N`. I'll make it a break statement instead just for better clarity.

reductionista:
Looks like the CI failed to install `postgresql-client-16`? Hopefully someone knows what's going on there and can unblock us.

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,262 @@
...
+	return c
+}
+
+func (c *EncodedLogCollector) BackfillForAddress(ctx context.Context, address string, fromSlot uint64) error {
+	pubKey, err := solana.PublicKeyFromBase58(address)
+	if err != nil {
+		return err
+	}
+
+	sigs, err := c.client.GetSignaturesForAddressWithOpts(ctx, pubKey, &rpc.GetSignaturesForAddressOpts{
```
dhaidashenko: Isn't this method only return 1,000 results at max? What if we need to backfill more transactions

EasterTheBunny: Made an update to this PR to address this concern

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,287 @@
...
+		if lowestSlotRead > 0 {
+			opts.Before = lowestSlotSig
+		}
+
+		sigs, err := c.client.GetSignaturesForAddressWithOpts(ctx, pubKey, &opts)
+		if err != nil {
+			return err
+		}
+
+		slots := make(map[uint64]struct{})
```
dhaidashenko: Should not slots be declared on higher level to avoid scheduling same slot processing twice, if transactions from the same slot appear on two different pages? 
nit: as transactions are ordered by slot number it should be enough to track lowestSlotRead to avoid scheduling work for the same slot twice

pkg/solana/logpoller/loader.go
```diff
@@ -0,0 +1,287 @@
...
+
+	var (
+		lowestSlotRead uint64
+		lowestSlotSig  solana.Signature
+	)
+
+	for lowestSlotRead > fromSlot || lowestSlotRead == 0 {
+		opts := rpc.GetSignaturesForAddressOpts{
+			Commitment:     rpc.CommitmentFinalized,
+			MinContextSlot: &fromSlot,
```
dhaidashenko, line 101:
```suggestion
			MinContextSlot: &toSlot,
```
We might need to introduce toSlot to guarantee that backfill and forward processing do not have gaps for a new filter

pkg/solana/logpoller/loader.go
```diff
@@ -90,29 +90,54 @@ func (c *EncodedLogCollector) BackfillForAddress(ctx context.Context, address st
...
+
+		for _, sig := range sigs {
+			if sig.Slot < lowestSlotRead {
+				lowestSlotRead = sig.Slot
+				lowestSlotSig = sig.Signature
 			}
 
-			slots[sig.Slot] = struct{}{}
+			_, ok := slots[sig.Slot]
+			if !ok {
```
reductionista: nit: could reduce indentation complexity
```go
if ok {
    continue
}
```

pkg/solana/logpoller/loader.go
EasterTheBunny: Are you meaning a range filter with both from and to? The RPC call is weird because the from endpoint is a slot number and the to endpoint is a transaction signature.

pkg/solana/logpoller/loader.go
```diff
@@ -90,29 +90,54 @@ func (c *EncodedLogCollector) BackfillForAddress(ctx context.Context, address st
...
-	if err != nil {
-		return err
-	}
+	var (
+		lowestSlotRead uint64
+		lowestSlotSig  solana.Signature
+	)
 
-	slots := make(map[uint64]struct{})
+	for lowestSlotRead > fromSlot {
```
reductionista: I think it's important we start from `fromSot` and read upward from there rather than starting at the current block and reading down to fromSlot. Doing it backwards will maximize latency because LogPoller won't be able to start saving the first slot (fromSlot) until all of them are read.  I know we're not guaranteeing the order at this level, but I think it would be a shame to have the typical case be maximally inefficient and correct order being maximally unlikely.

reductionista: I left a similar comment in the JIRA about this last week, although that was before you convinced me we didn't need to guarantee the order here.

pkg/solana/logpoller/loader.go
reductionista: I didn't realize this until after leaving mine, but I think he might be saying the same thing or similar to my comment above.  LogPoller must always process backfills in ascending order of block number, otherwise gaps can be introduced by node restarts which can result in logs getting permanently missed.

pkg/solana/logpoller/parser.go
```diff
@@ -0,0 +1,225 @@
+package logpoller
```
reductionista: This filename is going to have a merge conflict with @dhaidashenko 's PR which adds some of `logpoller/parser.go` from evm logpoller in the same location. Since this file already means something else in evm, I'd prefer to keep it that way in Solana and rename this to something else. (I think you mentioned you were renaming Parser to LogDataParser, maybe this can be `log_data_parser.go` ?

reductionista: Ah, is `ProgramEventProcessor` what used to be `Parser`? If so, maybe this should be `event_processor.go`? Actually I'd be fine with `event_parser.go` too... maybe that makes more sense not sure.

pkg/solana/logpoller/loader.go
reductionista: Bumping this in case you didn't see it earlier ^^ I was hoping you'd change this... or give an argument for why it's okay as is?

reductionista: As is, it's going to fetch data sometimes and then throw it away for no reason, which seems like the wrong behavior.

pkg/solana/logpoller/loader.go
EasterTheBunny: I'll have to think about how we need to do that. The api uses a slot as the lower limit and a transaction signature as an upper limit.

pkg/solana/logpoller/parser.go
EasterTheBunny: I can do that.

pkg/solana/logpoller/loader.go
reductionista: Hmm... yeah I guess that does pose a slight problem.

I don't know of any case where we'd care about having an upper limit, so that part is fine.  But the lower limit might be tricky if you can get back some signatures from a slot but not others. (For example, if 1000 is the maximum # it can return... then what if txs 998 - 1002 are all in a block together? I guess if it only returns 998, 999 and 1000 then we'd have to set the next lower limit as the same block, and get back a few that we already know about.  I think that's probably fine, since it's unlikely a single block would have more than 1 transaction we care about anyway.

pkg/solana/logpoller/loader.go
dhaidashenko: AFAIK the `GetSignaturesForAddressOpts` does not allow to go forward (from oldest to newest), so the only way to avoid gaps is to sync backfill and forward processing to ensure backfill starts going backwards from a block that was already processed by forward processing.

pkg/solana/logpoller/loader.go
EasterTheBunny: Ahh I understand now. I agree and I'll make the change.

pkg/solana/logpoller/loader.go
reductionista: @dhaidashenko I left a suggestion here for how we could address this:
https://github.com/smartcontractkit/chainlink-solana/pull/911#discussion_r1847593321

I guess it is not using the api for what it was designed for tho.  If we're only going to have 24 hours of history, then maybe it's not worth worrying about, we can just reverse it in the next layer. I'd mostly be worried if we had to go back a long ways like we do on evm.

EasterTheBunny: But we can't specify a `to` block/slot. We can only specify a transaction signature that acts as the upper limit. Maybe we will need this, but maybe not. I think we can add it later if we end up needing it though.

reductionista: Even if we could have a `to_slot` I don't follow @dhaidashenko 's reasoning about how/why that could help avoid gaps.

If we eventually implement a replay feature, it could be helpful for skipping over blocks that we've already pulled (to get to earlier ones), but it can't ever give us extra results that would be missing without it.

For now, the only use case we'll have is fetching new results up through and including the current block--where fromBlock will be passed in by `RegisterFilter` the first time, and then set to the most recent block in the db each additional time it's called.

That reminds me, is there an easy way to convert a `fromBlock` to a `fromSlot`? I guess we might have to call an extra rpc call to that.

reductionista: > If we eventually implement a replay feature, it could be helpful for skipping over blocks that we've already pulled (to get to earlier ones)

Also: when we do get around to implementing Replay, I think looking up the earliest log in the db and passing that transaction hash as the upper limiting signature will if anything be easier than it would be to use the block from that log to compute a `toSlot` to pass.  Either way we have to run the same query on the logs table to get the oldest log matching the filter--but in the second case we also have an extra conversion to do.

reductionista (dismissed):

reductionista:
Any idea what's causing this CI failure? Looks like it's trying to untar a compressed copy of the `chainlink-solana` repo:
```
/usr/bin/tar -xf /home/runner/work/_temp/95929ca9-9774-4949-a6bc-504001ec3a7d/cache.tzst -P -C /home/runner/work/chainlink-solana/chainlink-solana --use-compress-program unzstd
```
But the files already exist, so it just errors out?

EasterTheBunny:
> Any idea what's causing this CI failure? Looks like it's trying to untar a compressed copy of the `chainlink-solana` repo:
> 
> ```
> /usr/bin/tar -xf /home/runner/work/_temp/95929ca9-9774-4949-a6bc-504001ec3a7d/cache.tzst -P -C /home/runner/work/chainlink-solana/chainlink-solana --use-compress-program unzstd
> ```
> 
> But the files already exist, so it just errors out?

I don't know what's happening. It appears there is a conflict with the go version since this repo uses `1.23` in the toolchain but `1.22` as the primary. Maybe the github action is set up to use `1.23` already and it's trying to download the toolchain?? However to fix that we would need to go to another repo and fix the github action itself since I don't think we can upgrade this repo yet to `1.23`.

dhaidashenko (dismissed):

reductionista:
> I don't know what's happening. It appears there is a conflict with the go version since this repo uses 1.23 in the toolchain but 1.22 as the primary.

All I saw was an error about tar decompressing the files into a directory where they'd already been extracted. I didn't see anything about go versions in the logs, although I just fixed an issue with my PR related to that by updating to the latest version of `chainlink-testing-framework` where they've bumped the go version to 1.23 now.  Now I'm not seeing either of those errors, so I guess you must have fixed it also?  Looks like, just an ordinary test failure... hopefully a flake?

aalu1418 (dismissed):
the error was caused by an issue with the upgrade test having some hard coded values - it should be fixed here: 
- https://github.com/smartcontractkit/chainlink-solana/pull/940

other changes look good!

aalu1418 (dismissed):

aalu1418 (dismissed):

aalu1418 (approved):

PR #916, opened 2024-11-06
Add DataSource to ChainOpts and pass to NewChain

reductionista:
### Requires Dependencies
- https://github.com/smartcontractkit/chainlink-common/pull/919

### Resolves Dependencies
  - https://github.com/smartcontractkit/chainlink/pull/15127

core ref: 7b0c8369f2ae3a9a7fdf0eb490a6ae877b5d9540


reductionista:
Everything is green on all 3 linked PR's except for this odd golint error:

```
level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package nodeclient: could not load export data: no export data for \"github.com/smartcontractkit/chainlink/deployment/environment/nodeclient\""
level=error msg="Running error: can't run linter goanalysis_metalinter\ninspect: failed to load package nodeclient: could not load export data: no export data for \"github.com/smartcontractkit/chainlink/deployment/environment/nodeclient\""
make: *** [Makefile:117: lint-go-integration-tests] Error 3
```

This seems more likely to be an issue with golint than with the PR, but not sure. Any ideas?

integration-tests/go.mod
```diff
@@ -517,5 +527,14 @@ require (
...
+replace (
+	// geth wants v2.3.4 but that is incompatible with github.com/cometbft/cometbft v0.37.5 which when bumped is incompatible with github.com/cosmos/cosmos-sdk
+	// This line can be removed after these imports are bumped or removed.
+	github.com/btcsuite/btcd/btcec/v2 => github.com/btcsuite/btcd/btcec/v2 v2.3.2
+
+	// replicating the replace directive on cosmos SDK
+	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+
+	// replicating replace directive in chainlink
+	github.com/sourcegraph/sourcegraph/conc => github.com/sourcegraph/sourcegraph-public-snapshot/conc v0.0.0-20240822153003-c864f15af264
```
jmank88: This doesn't exist. Where did it come from?
```suggestion
```

reductionista: Ah, was trying to keep the version of that module in sync with `chainlink` because I often get errors about that module in GoLand and thought the version mismatch might be the cause. Didn't realize they'd removed the `conc` package from this version.

integration-tests/common/test_common.go
```diff
@@ -21,7 +21,7 @@ import (
 	test_env_ctf "github.com/smartcontractkit/chainlink-testing-framework/lib/docker/test_env"
 	"github.com/smartcontractkit/chainlink-testing-framework/lib/utils/testcontext"
 
-	"github.com/smartcontractkit/chainlink/integration-tests/client"
+	client "github.com/smartcontractkit/chainlink/deployment/environment/nodeclient"
```
dhaidashenko, line 24: Is this change intentional?

reductionista: Yes, it was necessary in order to fix errors resulting from updating the ref to `chainlink/integration-tests` in `chainlink-solana/integration-tests/go.mod`. The `client` package was moved into a new `deployment` package in recent versions of `chainlink`, but nobody had updated the Solana repo with that change for some reason.

integration-tests/common/test_common.go
```diff
@@ -15,13 +15,12 @@ import (
 	"github.com/google/uuid"
 	"github.com/lib/pq"
 	"github.com/rs/zerolog/log"
-	"github.com/stretchr/testify/require"
-	"gopkg.in/guregu/null.v4"
-
 	test_env_ctf "github.com/smartcontractkit/chainlink-testing-framework/lib/docker/test_env"
 	"github.com/smartcontractkit/chainlink-testing-framework/lib/utils/testcontext"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/guregu/null.v2"
```
jmank88: Why downgrade this from v4?

jmank88: Maybe consider updating the linter file like we do in core:
https://github.com/smartcontractkit/chainlink/blob/50c1b3dbf31bd145b312739b08967600a5c67f30/.golangci.yml#L139-L144

pkg/solana/chain.go
```diff
@@ -222,7 +226,7 @@ func (v *verifiedCachedClient) GetAccountInfoWithOpts(ctx context.Context, addr
 	return v.ReaderWriter.GetAccountInfoWithOpts(ctx, addr, opts)
 }
 
-func newChain(id string, cfg *config.TOMLConfig, ks loop.Keystore, lggr logger.Logger) (*chain, error) {
+func newChain(id string, cfg *config.TOMLConfig, ks core.Keystore, lggr logger.Logger, ds sqlutil.DataSource) (*chain, error) {
```
jmank88, line 229: When will this be used?

integration-tests/common/test_common.go
reductionista: Weird, I don't know why that keeps happening. Each time I notice it changes to `null.v2` I put it back to `null.v4` but... something I'm running really wants it to be v2.  Must be either goland or vim that's changing it?

pkg/solana/chain.go
reductionista: It will be used by LogPoller to initialize its ORM. We have the db schemas & some of the orm written now, but haven't added the constructor for LogPoller yet. I could add it as a part of this PR, but it seems more related to the JIRA we have for the main LogPoller runloop than this one. Originally it looked like tx mgr would also need it, but they've since confirmed they won't need to persist anything to db.

integration-tests/common/test_common.go
reductionista: Hmm... it looks like we do have those linter rules already in `chainlink-solana/.golangci.yml`. Maybe we need to create a separate `chainlink-solana/integration-tests/.golangci.yml` or do the existing linter rules apply to both modules?

reductionista: Looks like the existing yml file does apply to both, because the `make lint-go-integration-tests` command is run from `chainlink-solana` root dir

reductionista: Oh, I see why this null.v2 problem came back... I resolved some merge conflicts while rebasing `go.mod` by running `git rebase --skip`, apparently I undid the fix for that and some related things.

pkg/solana/chain_test.go
```diff
@@ -17,14 +17,16 @@ import (
 	"github.com/gagliardetto/solana-go/programs/system"
 	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/google/uuid"
-	"github.com/smartcontractkit/chainlink-common/pkg/config"
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
-	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+	"github.com/smartcontractkit/chainlink-common/pkg/sqlutil/sqltest"
```
jmank88: I expect the linter to complain about the sorting here :thinking:

jmank88:
I think you will need to update the setup-go action here to get core to pass.
https://github.com/smartcontractkit/chainlink-solana/blob/3ff8c1bf1fe2218be4ff9c54b8a062d7463b1553/.github/workflows/relay.yml#L28-L31
I think the correct thing to do would be to set `GOTOOLCHAIN=auto`? Because setting explicit 1.23 would require manual updates, and bumping the go.mod version just to force this indirectly would be hacky.

.github/workflows/relay.yml
```diff
@@ -26,6 +26,8 @@ jobs:
         uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
       - name: Setup go
         uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+        env:
+          GOTOOLCHAIN=auto
```
chainchad:
```suggestion
          GOTOOLCHAIN: auto
```

jmank88 (dismissed):

jmank88 (dismissed):

aalu1418 (dismissed):

dhaidashenko (dismissed):

reductionista:
Hmm... says I still need 1 approving review, despite having 3 already.

It must be due to one of these two changes, since all the rest is under bix-build for CODEOWNERS:
```
[MB-TQH59VKW66:~/workspace/chainlink-solana> git diff origin/develop   --  .github/
diff --git a/.github/workflows/relay.yml b/.github/workflows/relay.yml
index 67921b87..164754d6 100644
--- a/.github/workflows/relay.yml
+++ b/.github/workflows/relay.yml
@@ -26,6 +26,8 @@ jobs:
         uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
       - name: Setup go
         uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+        env:
+          GOTOOLCHAIN: auto
         with:
           go-version-file: "go.mod"
           check-latest: true
[MB-TQH59VKW66:~/workspace/chainlink-solana> git diff origin/develop   --  ops/monitoring
diff --git a/ops/monitoring/Dockerfile b/ops/monitoring/Dockerfile
index 825fb5e4..2b281f41 100644
--- a/ops/monitoring/Dockerfile
+++ b/ops/monitoring/Dockerfile
@@ -1,6 +1,6 @@
 # Build image

-FROM golang:1.22.6 AS build
+FROM golang:1.23.3 AS build

 # OS dependencies
 RUN apt-get update && apt-get install -y wget gcc
```

`.github/*` is rel-eng while `ops/monitoring/*` is realtime. I'm not sure how I've satisfied either of those requirements given the 3 approvals I've got, so I guess I'll ask both of them to look at it 🤔

(If these changes don't make sense, happy to undo them I just figured since we're setting it to 1.23.3 and GOTOOLCHAIN everywhere else we may as well here.)

jmank88:
> (If these changes don't make sense, happy to undo them I just figured since we're setting it to 1.23.3 and GOTOOLCHAIN everywhere else we may as well here.)

`auto` is the default, so it would be cleaner to only set it when necessary.

hendoxc (dismissed):

aalu1418 (approved):

jmank88 (approved):

PR #919, opened 2024-11-08
Update simulation to set max compute unit limit and enable sig verification

amit-momin:
- If CU limit is not set prior to simulation, the default 200k limit could cause simulation to return an avoidable insufficient CU error. This error can be avoided by setting the max compute unit limit (1.4M) for a transaction before simulating.
- Signature verification can have a big impact on the compute units used during simulation so the option has been enabled.


aalu1418 (dismissed):

silaslenihan (approved):
LGTM

aalu1418 (approved):

PR #924, opened 2024-11-11
[NONEVM-706][solana] - Support custom fee bumping strategies through RPC call 

Farber98:
### Description
Current Solana TXM implementation bumps fees at at a specified interval of time without waiting for previous transaction expiration. CCIP should not automatically bump fees without ensuring expiration to prevent excessive bumping

### Acceptance Criteria
- The TXM should be able to `detect when a transaction has expired`
- The TXM should be configurable to `enable/disable bumping`
- The TXM should be configurable to wait for either `interval of time` or `tx expiry` before bumping, if enabled

### This PR
**New `FeeBumpCriteria` Options:**
- `fixed-intervals`: Bumps the transaction fee at regular, fixed intervals.
- `expiration`: Bumps the fee based on blockhash expiration, ensuring transactions remain valid.
- `none`: Disables fee bumping

**sendWithRetry refactorization:**
- Organized the retry logic into clear, manageable steps using helper functions for better readability and maintenance.
- Integrate Fee Bumping based on `FeeBumpCriteria`
	- `Fixed Intervals` bump every `FeeBumpPeriod`. Added blockhash expiration handling here too.
	- `Expiration` checks every `FeeBumpPeriod` if we need a bump based on blockhash expiration.
	- Added `lastValidBlockHeight` to the `pendingTx` struct so we can monitor blockhash expiration. This is stored when we request a new blockhash for a tx.
	- Added `IsBlockhashExpired` and `UpdateBlockhash` to handle changes in our in-memory layer.

### DIFF
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -580,6 +580,7 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 		txm.SetComputeUnitPriceMax(0),
 		txm.SetComputeUnitPriceMin(0),
 		txm.SetBaseComputeUnitPrice(0),
+		txm.SetFeeBumpCriteria("fixed-interval"),
 		txm.SetFeeBumpPeriod(0),
 	)
 	if err != nil {
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -574,7 +574,7 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 	}
 
 	chainTxm := c.TxManager()
-	err = chainTxm.Enqueue(ctx, "", tx, nil,
+	err = chainTxm.Enqueue(ctx, "", tx, nil, blockhash.Value.LastValidBlockHeight,
 		txm.SetComputeUnitLimit(500), // reduce from default 200K limit - should only take 450 compute units
 		// no fee bumping and no additional fee - makes validating balance accurate
 		txm.SetComputeUnitPriceMax(0),
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -580,7 +580,7 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 		txm.SetComputeUnitPriceMax(0),
 		txm.SetComputeUnitPriceMin(0),
 		txm.SetBaseComputeUnitPrice(0),
-		txm.SetFeeBumpCriteria("fixed-interval"),
+		txm.SetFeeBumpCriteria("fixed-intervals"),
 		txm.SetFeeBumpPeriod(0),
 	)
 	if err != nil {
--- a/pkg/solana/txm/pendingtx.go
+++ b/pkg/solana/txm/pendingtx.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
 	"golang.org/x/exp/maps"
 )
 
@@ -30,6 +31,10 @@ type PendingTxContext interface {
 	ListAll() []solana.Signature
 	// Expired returns whether or not confirmation timeout amount of time has passed since creation
 	Expired(sig solana.Signature, confirmationTimeout time.Duration) bool
+	// IsBlockhashExpired returns whether or not the blockhash of a transaction has expired when compared to currHeight. Useful to check if should be bumped.
+	IsBlockhashExpired(id string, currHeight uint64) bool
+	// UpdateBlockhash updates the last valid block height for a transaction when expired and needs bumping in the context of FeeBumpCriteria='expiration'.
+	UpdateBlockhash(id string, blockhashResult *rpc.GetLatestBlockhashResult) error
 	// OnProcessed marks transactions as Processed
 	OnProcessed(sig solana.Signature) (string, error)
 	// OnConfirmed marks transaction as Confirmed and moves it from broadcast map to confirmed map
@@ -45,13 +50,14 @@ type PendingTxContext interface {
 }
 
 type pendingTx struct {
-	tx          solana.Transaction
-	cfg         TxConfig
-	signatures  []solana.Signature
-	id          string
-	createTs    time.Time
-	retentionTs time.Time
-	state       TxState
+	tx                   solana.Transaction
+	cfg                  TxConfig
+	signatures           []solana.Signature
+	id                   string
+	createTs             time.Time
+	retentionTs          time.Time
+	state                TxState
+	lastValidBlockHeight uint64
 }
 
 var _ PendingTxContext = &pendingTxContext{}
@@ -269,6 +275,30 @@ func (c *pendingTxContext) OnProcessed(sig solana.Signature) (string, error) {
 		c.broadcastedTxs[id] = tx
 		return id, nil
 	})
+}
+
+// IsBlockhashExpired returns whether or not the blockhash of a transaction has expired when compared to currHeight. Useful to check if should be bumped.
+func (c *pendingTxContext) IsBlockhashExpired(id string, currHeight uint64) bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	tx, exists := c.broadcastedTxs[id]
+	if !exists {
+		return false
+	}
+	return tx.lastValidBlockHeight < currHeight
+}
+
+// UpdateBlockhash updates the blockhash and last valid block height for a transaction when expired and needs bumping in the context of FeeBumpCriteria='expiration'.
+func (c *pendingTxContext) UpdateBlockhash(id string, blockhashResult *rpc.GetLatestBlockhashResult) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if tx, exists := c.broadcastedTxs[id]; exists {
+		tx.tx.Message.RecentBlockhash = blockhashResult.Value.Blockhash
+		tx.lastValidBlockHeight = blockhashResult.Value.LastValidBlockHeight
+		c.broadcastedTxs[id] = tx
+		return nil
+	}
+	return fmt.Errorf("transaction not found for id: %s", id)
 }
 
 func (c *pendingTxContext) OnConfirmed(sig solana.Signature) (string, error) {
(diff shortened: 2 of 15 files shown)


pkg/solana/txm/pendingtx.go
```diff
@@ -271,6 +277,30 @@ func (c *pendingTxContext) OnProcessed(sig solana.Signature) (string, error) {
...
+	}
+	return tx.lastValidBlockHeight < currHeight
+}
+
+// UpdateBlockhash updates the blockhash and last valid height for a transaction when expired and needs bumping.
+func (c *pendingTxContext) UpdateBlockhash(id string, blockhashResult *rpc.GetLatestBlockhashResult) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if tx, exists := c.broadcastedTxs[id]; exists {
+		tx.tx.Message.RecentBlockhash = blockhashResult.Value.Blockhash
```
Farber98, line 296: It's okay to handle the update like this? Or this breaks some signing assumptions and we should replace the entire tx

pkg/solana/txm/txm.go
```diff
@@ -303,65 +295,224 @@ func (txm *Txm) sendWithRetry(ctx context.Context, msg pendingTx) (solanaGo.Tran
...
+	}
+
+	// We need to update blockhash in retryTx and pendingTx context if it has expired
+	if needBlockhashUpdate {
+		// Updates Blockhash and LastValidBlockHeight in the retry transaction
+		blockhashResult, err := getLatestBlockhash(ctx)
+		if err != nil {
+			return solanaGo.Transaction{}, fmt.Errorf("failed to get latest blockhash: %w", err)
+		}
+		baseTx.Message.RecentBlockhash = blockhashResult.Value.Blockhash
```
Farber98, line 436: same as [this other comment](https://github.com/smartcontractkit/chainlink-solana/pull/924/files#r1840462992)

pkg/solana/txm/pendingtx.go
```diff
@@ -271,6 +277,30 @@ func (c *pendingTxContext) OnProcessed(sig solana.Signature) (string, error) {
...
 
+// IsBlockhashExpired returns whether or not the blockhash of a transaction has expired when compared to currHeight.
+func (c *pendingTxContext) IsBlockhashExpired(id string, currHeight uint64) bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	tx, exists := c.broadcastedTxs[id]
+	if !exists {
+		return false
+	}
+	return tx.lastValidBlockHeight < currHeight
```
amit-momin, line 288: I think we would want to also double check the `tx.state` to make sure it's `Broadcasted`. I probably could have done a better job naming this map but transactions in here could also be `Processed` in which case maybe expiration doesn't matter

pkg/solana/chain.go
```diff
@@ -574,12 +574,13 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 	}
 
 	chainTxm := c.TxManager()
-	err = chainTxm.Enqueue(ctx, "", tx, nil,
+	err = chainTxm.Enqueue(ctx, "", tx, nil, blockhash.Value.LastValidBlockHeight,
```
amit-momin, line 577: I think we can avoid changing the `Enqueue` method signature here. Not an issue if we want to though. I think we can add the `c.reader.LatestBlockhash(ctx)` logic in `sendWithRetry` and just overwrite `tx.Message.RecentBlockhash`. That way the TXM can own assigning the block hashes to transactions.

pkg/solana/txm/txm.go
```diff
@@ -303,65 +295,224 @@ func (txm *Txm) sendWithRetry(ctx context.Context, msg pendingTx) (solanaGo.Tran
...
+	// base compute unit price should only be calculated once
+	// prevent underlying base changing when bumping (could occur with RPC based estimation)
+	fee := fees.CalculateFee(
+		txm.cfg.ComputeUnitPriceDefault(),
+		txm.cfg.ComputeUnitPriceMax(),
+		txm.cfg.ComputeUnitPriceMin(),
+		uint(count), //nolint:gosec // reasonable number of bumps should never cause overflow
+	)
+	return fees.ComputeUnitPrice(fee)
+}
```
amit-momin, line 355: I think this behavior is different than what was extracted from above. Looks like you're using the txm configs here where above we're using the msg configs. Even though msg configs are derived from the txm configs, these are still separate since txm level configs can be overridden at the msg level. Think you just need to pass in the `TxConfig` as a parameter and just use those. Then update passing `msg.cfg` into this method
```suggestion
func (txm *Txm) getFee(count int, cfg TxConfig) fees.ComputeUnitPrice {
	// base compute unit price should only be calculated once
	// prevent underlying base changing when bumping (could occur with RPC based estimation)
	fee := fees.CalculateFee(
		cfg.ComputeUnitPriceDefault(),
		cfg.ComputeUnitPriceMax(),
		cfg.ComputeUnitPriceMin(),
		uint(count), //nolint:gosec // reasonable number of bumps should never cause overflow
	)
	return fees.ComputeUnitPrice(fee)
}
```

pkg/solana/txm/pendingtx.go
amit-momin: I think this is ok since this is essentially just the base tx. We should also clear the `msg.signatures` though since those correlate to the expires transaction which we don't need to track anymore for confirmation

pkg/solana/txm/txm.go
```diff
@@ -303,65 +295,224 @@ func (txm *Txm) sendWithRetry(ctx context.Context, msg pendingTx) (solanaGo.Tran
...
+
+// bumpFeeAndUpdateBlockhash determines whether the transaction fee should be bumped
+// and whether the blockhash needs to be updated based on the configured criteria.
+func (txm *Txm) bumpFeeAndUpdateBlockhash(ctx context.Context, txID string, bumpTime time.Time) (shouldBump bool, needBlockhashUpdate bool, newBumpTime time.Time, err error) {
```
amit-momin, line 359: I noticed we call this in the retry routine for each transaction. I think that could lead to more RPC calls than we need to make for this. Otherwise, we'd be fetching `SlotHeight` for each transaction that's reached its retry interval. I feel like we could use a TXM level go routine like the `reap` or `confirm` loops to handle expiration. That way we only call `SlotHeight` once every time we reach some "check expiration" interval then evaluate all transactions in `Broadcasted` state against it

pkg/solana/txm/txm.go
```diff
@@ -303,65 +295,224 @@ func (txm *Txm) sendWithRetry(ctx context.Context, msg pendingTx) (solanaGo.Tran
...
-			if deltaT > MaxRetryTimeMs {
-				deltaT = MaxRetryTimeMs
+			// in case blockhash has expired, we need to update it.
+			if txm.txs.IsBlockhashExpired(txID, currHeight) {
+				needBlockhashUpdate = true
 			}
-			tick = time.After(time.Duration(deltaT) * time.Millisecond)
+		case FeeBumpCriteriaExpiration:
+			// Bump the fee only if the blockhash has expired
+			if txm.txs.IsBlockhashExpired(txID, currHeight) {
```
amit-momin, line 397: I noticed you added expiration detection for both bumping strategies. I was thinking about it after we last talking and think this is a good idea. This might also help simplify this a bit. If we move the the expiration detection logic to a TXM level loop instead, we could actually keep that separate from gas bumping. It would be it's own separate feature that can be toggled on and off by some `TxExpirationTimeout` config (tentative name). So regardless if we have bumping or not, we could detect if txs expire and replace them with new ones with new block hashes. In that case, we could maybe leave bumping the way it is. Then we also wouldn't need an explicit `FeeBumpCriteria` config since we could implicitly set these "strategies" by either toggling fee bumping on or off and expiration detection on or off.

pkg/solana/txm/txm.go
amit-momin: I think this is ok since buildTx would take care of the signing below. The only thing like the previous comment is that we'd want to clear out `msg.signatures` too along with the entries for those in the signature map. But if we move the code to a TXM level loop, it might be easier to just use the `Remove` method to drop the expired tx from our storage and call `sendWithRetry` again on the baseTx. If we also move the `getLatestBlockhash` logic into `sendWithRetry` we wouldn't have to worry about that either

pkg/solana/txm/txm.go
Farber98: I like this approach! It feels like it will simplify things.

Let’s confirm we’re on the same page before moving forward:

#### Bumping as it currently works:
- **Data Feeds** use case aiming to get transactions through as quickly as possible, even with the risk of overpaying.
- We’ll bump the existing transaction using exponential backoff.

#### Expiration feature:
- **CCIP** use case aiming to wait longer to avoid overpaying for transactions.
- We’ll wait until expiration to trigger `sendWithRetry` again with `baseTx` and a new `blockhash`. Although there’s no explicit bump, it may occur implicitly if `recalculated fees` are higher.
- We will use a separate loop to detect `expiration` and call `sendWithRetry` again.
- For this, we will need to move blockhash assignment to `sendWithRetry`

#### Interaction between them:
- If `expiration` and `bumping` are both on:
	- We’ll bump the existing transaction using exponential backoff. If expiration is reached, we’ll trigger `sendWithRetry` again with `baseTx` and a different blockhash, restarting exponential backoff bumping.
	- Reaching expiration (150 blocks) when bumping is on should be rare.

- If `expiration` is on and `bumping` is off:
	- We’ll wait until expiration to trigger `sendWithRetry` again with `baseTx` and a new `blockhash`. This won’t explicitly bump but may do so implicitly if recalculated fees are higher.
	
- If `expiration` is off and `bumping` is on:
	- Current behavior: We’ll bump the existing transaction using exponential backoff. If expiration is reached, tx won't go through.
	 
- If `expiration` and `bumping` are both off:
	- The transaction may not go through if it expires and the fees are insufficient.

Farber98: We should: 
- be careful with `TxConfirmTimeout` though as it may overlap with `expiration`. 
- be clear with config docs so we don't throw the complexity within interaction between the variables

Farber98: Notes for meeting: 

Expiration functionality changes:
- Create a loop that, every `TxExpirationTimeout`, retrieves the `slotHeight` and checks if txes in our `broadcastedMap` that are in a `broadcasted` state have expired.
- We do this by adding `lastValidBlockHeight` to the `pendingTx` when we create it. This allows us to compare the `slotHeight` to see if it has exceeded that value and decide if we need to rebuild and rebroadcast the transaction.
- To achieve this, we will need to move the blockhash assignment logic to the `sendWithRetry` function.
- If we need to rebroadcast a transaction, we should:
	- remove previous one by calling the `Remove()`
	- build the new transaction and call `sendWithRetry` with it

PR #926, opened 2024-11-13
Batch Contract Read

EasterTheBunny:
Batch contract reads are added and multiple address support was removed. All tests now pass.


pkg/solana/chainreader/chain_reader.go
```diff
@@ -114,123 +111,73 @@ func (s *SolanaChainReaderService) GetLatestValue(ctx context.Context, readIdent
...
-	value, err := values.Wrap(contractType)
-	if err != nil {
-		return err
+	if results[0].err != nil {
+		return fmt.Errorf("%w: %s", types.ErrInternal, results[0].err)
 	}
 
-	*ptrToValue = value
-
 	return nil
```
patrick-dowell, line 136: Don't we still need to set `returnVal.(*values.Value)`?

patrick-dowell (commented):
I'm curious - what are we testing via these tests? Do we have any sort of E2E testing ability to test against a simulated backend? What are the existing tests actually testing?

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: The type casting creates the pointer to the value. Setting the value on line 157 is correct. You can see the implementation in evm is the same.

pkg/solana/chainreader/batch.go
```diff
@@ -0,0 +1,107 @@
...
+
+			continue
+		}
+
+		results[idx].err = errors.Join(
+			results[idx].err,
+			binding.Decode(ctx, data[idx], contractType),
+		)
+
+		value, err := values.Wrap(contractType)
```
ilija42, line 96: In what format does the rpc return data? Do we have to da a borsh decode?

ilija42 (commented):
Don't have enough context on batch, but everything else looks fine

pkg/solana/chainreader/batch.go
EasterTheBunny: Since the start, we have been assuming binary encoded data (borsh). The other option is JSON but so far expectations did not include json.

patrick-dowell (approved):
LGTM

ilija42 (approved):

aalu1418 (approved):

PR #928, opened 2024-11-15
[NONEVM-706][Solana][deprecated] - Refactor TXM + Rebroadcast Expired Tx functionality

Farber98:
### Description
Refactor `confirm` and `sendWithRetry` funcs

Adding a new `TxExpirationRebroadcast` toggable routine:
- Inside `confirm` loop, after checking confirmations, we will determine if we need to rebroadcast expired transactions that where not processed yet. Rebroadcasting a transaction involves:
	- Removing previous tx and creating a new one with updated blockhash 
		- **Consideration:** txid will be maintained, in case it was set by caller
	- calling `sendWithRetry` directly without enqueuing
		- If error, we should use [prebroadcast error functionality](https://github.com/smartcontractkit/chainlink-solana/pull/936). Client will need to retry the tx if this happens
	- **Consideration:** Implicitly it may be "bumping" despite config being off. We may get a new higher price for the rebroadcasted tx when we build and get a new compute unit price.

https://smartcontract-it.atlassian.net/browse/NONEVM-706


Farber98:
until this point, only refactor of existing logic: 
- moved blockhash assignment to txm
- reduce sendWithRetry complexity making it simpler/clearer

Farber98:
everything is refactor until this point

Farber98:
Until this point, everything was refactor of the existing logic. From now on, new functionality is being added:

pkg/solana/chain.go
```diff
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gagliardetto/solana-go"
 	solanago "github.com/gagliardetto/solana-go"
```
amit-momin: I think the import already existed under the `solanago` alias. Think this was aliased to avoid naming collision with the package

pkg/solana/chain.go
```diff
@@ -557,21 +549,25 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 				toKey,
 			).Build(),
 		},
-		blockhash.Value.Blockhash,
```
amit-momin: I think it's ok if the caller to `Enqueue` sets a block hash here. We'll just overwrite it when needed.

pkg/solana/chain.go
```diff
@@ -557,21 +549,25 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
...
+		// To perform balanceCheck we need a blockhash.
+		// Storing values to perform balanceCheck within sendWithRetry txm function before sending tx.
+		BalanceCheck: balanceCheck,
+		From:         fromKey,
+		Amount:       amountI,
 	}
 
 	chainTxm := c.TxManager()
-	err = chainTxm.Enqueue(ctx, "", tx, nil,
+	err = chainTxm.Enqueue(ctx, msg,
```
amit-momin: I don't think we should move building the `msg` outside of `Enqueue`. `pendingTx` should just be an internal struct to the TXM for how we want to store/track things. So a caller to `Enqueue` should only worry about the solana tx and pass that in. Then we can build this on our side.

pkg/solana/txm/pendingtx.go
```diff
@@ -44,14 +46,21 @@ type PendingTxContext interface {
...
-	id          string
-	createTs    time.Time
-	retentionTs time.Time
-	state       TxState
+type PendingTx struct {
+	Tx                   solana.Transaction
+	AccountID            string
+	cfg                  TxConfig
+	signatures           []solana.Signature
+	UUID                 string
```
amit-momin: I think this name might be misleading. This won't necessarily be a uuid. We allow the caller to provide a unique identifier for a tx so they can track it's status externally. Some of them like CCIP might make it some deterministic string that they can build from their metadata so they don't have to store a uuid.

pkg/solana/txm/txm.go
```diff
@@ -378,135 +453,203 @@ func (txm *Txm) confirm() {
...
 				break
 			}
 
-			// get client
-			client, err := txm.client.Get()
 			if err != nil {
-				txm.lggr.Errorw("failed to get client in soltxm.confirm", "error", err)
-				break // exit switch
+				txm.lggr.Errorw("failed to get client in txm.confirm", "error", err)
+				return
```
amit-momin: Why did we change this to return? Wouldn't this end the confirm loop entirely?

pkg/solana/txm/txm.go
```diff
@@ -378,135 +453,203 @@ func (txm *Txm) confirm() {
...
 			}
-
-			// batch sigs no more than MaxSigsToConfirm each
-			sigsBatch, err := utils.BatchSplit(sigs, MaxSigsToConfirm)
-			if err != nil { // this should never happen
-				txm.lggr.Fatalw("failed to batch signatures", "error", err)
-				break // exit switch
+			if txm.cfg.TxExpirationRebroadcast() {
+				txm.rebroadcastExpiredTxs(ctx, client)
 			}
```
amit-momin: We may want to do this after we process confirmation in case a transaction gets processed between the last check and this one close to its expiration slot. We could potentially rebroadcast a transaction we think is expired even though it was included. But this does remind me that we probably need to be careful with `BlockhashNotFound` errors in the confirm loop so we're not marking expired tx as errored if `TxExpirationRebroadcast` is enabled

pkg/solana/txm/txm.go
```diff
@@ -378,135 +453,203 @@ func (txm *Txm) confirm() {
...
+			continue
+		}
+	}
+}
+
+func (txm *Txm) handleNotFoundSignatureStatus(sig solanaGo.Signature) {
+	txm.lggr.Debugw("tx state: not found", "signature", sig)
+
+	// check confirm timeout exceeded
+	if txm.txs.Expired(sig, txm.cfg.TxConfirmTimeout()) {
```
amit-momin: Just noticed comparing this logic against the processed method, we should also add the check for TxConfirmTimeout=0 here too
```if txm.cfg.TxConfirmTimeout() != 0*time.Second && txm.txs.Expired(sig, txm.cfg.TxConfirmTimeout()) {```

File: pkg/solana/txm/txm.go [hunk 8]
```diff
@@ -175,193 +175,236 @@ func (txm *Txm) run() {
...
-		if signErr != nil {
-			return solanaGo.Transaction{}, fmt.Errorf("error in soltxm.SendWithRetry.Sign: %w", signErr)
+// prepareTransaction sets blockhash and lastValidBlockHeight which will be used to track expiration.
+// If required, it also performs balanceCheck and sets compute unit limit.
+func (txm *Txm) prepareTransaction(ctx context.Context, msg *pendingTx) error {
+	client, err := txm.client.Get()
+	if err != nil {
+		return fmt.Errorf("failed to get client in sendWithRetry: %w", err)
+	}
+	blockhash, err := client.LatestBlockhash(ctx)
```
Farber98: by moving things around we may be able to execute this rpc call only when explicitly needed to overwrite blockhash (in case of expiration). This may require to add lastValidBlockheight in Enqueue signature

Farber98:
Discussion with Amit: 

- Should we [enqueue or call sendWithRetry](https://github.com/smartcontractkit/chainlink-solana/blob/47d8d64eb46a813fec136b20dbe76e6a78358baa/pkg/solana/txm/txm.go#L545) directly? 
	- As the tx was already broadcasted before, it may make sense to "prioritize it" with a direct call to `sendWithRetry`. If we do so, and a lot of txes are being rebroadcasted directly with `sendWithRetry`, could this cause problems? 
		- Ans: call sendWithRetry directly. We won't have problems by doing so.
- Should we handle the case where [enqueueing the rebroadcast fails](https://github.com/smartcontractkit/chainlink-solana/blob/47d8d64eb46a813fec136b20dbe76e6a78358baa/pkg/solana/txm/txm.go#L545)?
	- We have already removed our tx so we may lose it. Should we first `enqueue/sendWithRetry` and then remove if succesful?
		- Ans: If fails, mark tx as errored. Client will retry
- Should we continue trying with other txes if [enqueuing the rebroadcast fails](https://github.com/smartcontractkit/chainlink-solana/blob/47d8d64eb46a813fec136b20dbe76e6a78358baa/pkg/solana/txm/txm.go#L546)?
	- Mainly thinking if we decide to enqueue and the queue is full.
		- Ans: We will call sendWithRetry, so we don't need to worry about this.
- Should we limit the number of [rebroadcasted txes per cycle](https://github.com/smartcontractkit/chainlink-solana/blob/nonevm-706-support-custom-bumping-strategy-rpc-expiration-within-confirmation/pkg/solana/txm/txm.go#L513) in terms of performance?
	- Ans: not for now with the load we currently have.
- Should we generate [new ids for rebroadcasted txes](https://github.com/smartcontractkit/chainlink-solana/blob/nonevm-706-support-custom-bumping-strategy-rpc-expiration-within-confirmation/pkg/solana/txm/txm.go#L539)? 
- Should we change the signature?
	- Not doing it will prevent bumping from working with expiration. signature will already exist and when trying to bump we'll get [`error in adding retry transaction`](https://github.com/smartcontractkit/chainlink-solana/blob/nonevm-706-support-custom-bumping-strategy-rpc-expiration-within-confirmation/pkg/solana/txm/txm.go#L414)	
	- If we change the signature, it may require to rethink how signatures are assigned when bumping is also enabled as it may produce races
		- Ans: we should try to use the same id. Was thinking on using the same family but with a slight difference and see how it works.

pkg/solana/txm/txm.go
Farber98: done

pkg/solana/txm/txm.go
```diff
@@ -377,135 +381,219 @@ func (txm *Txm) confirm() {
...
+			txm.processConfirmations(ctx, client)
 
-			// batch sigs no more than MaxSigsToConfirm each
-			sigsBatch, err := utils.BatchSplit(sigs, MaxSigsToConfirm)
-			if err != nil { // this should never happen
-				txm.lggr.Fatalw("failed to batch signatures", "error", err)
-				break // exit switch
+			// In case all txes where confirmed and there's nothing to rebroadcast.
+			// This check saves making 2 RPC calls (slot height + blockhash) when there's nothing to process.
+			if len(txm.txs.ListAllBroadcastedTxs()) == 0 {
```
Farber98: I could call ListAllExpiredBroadcastedTxs(val) with val being really high (max.Int maybe) and it will return all broadcasted txes. Despite being dirty, it would let us remove ListAllBroadcastedTxs() implementation and prevent doing the RPC call to get the slotHeight

pkg/solana/chain.go
```diff
@@ -557,7 +557,7 @@ func (c *chain) sendTx(ctx context.Context, from, to string, amount *big.Int, ba
 				toKey,
 			).Build(),
 		},
-		blockhash.Value.Blockhash,
+		blockhash.Value.Blockhash, // Will be override if needed within sendWithRetry txm function.
```
amit-momin: nit: just a grammar thing but we also probably don't need to add the internal TXM details here since this is outside of the TXM
```suggestion
		blockhash.Value.Blockhash, // Will be overridden by the TXM if needed
```

pkg/solana/transmitter.go
```diff
@@ -78,7 +78,7 @@ func (c *Transmitter) Transmit(
 		[]solana.Instruction{
 			solana.NewInstruction(c.programID, accounts, data.Bytes()),
 		},
-		blockhash.Value.Blockhash,
+		blockhash.Value.Blockhash, // Will be override if needed within sendWithRetry txm function.
```
amit-momin: Just to track, same as the other comment in `pkg/solana/chain.go`

pkg/solana/txm/pendingtx.go
```diff
@@ -60,9 +68,9 @@ type pendingTxContext struct {
 	cancelBy map[string]context.CancelFunc
 	sigToID  map[solana.Signature]string
 
-	broadcastedTxs      map[string]pendingTx // transactions that require retry and bumping i.e broadcasted, processed
-	confirmedTxs        map[string]pendingTx // transactions that require monitoring for re-org
-	finalizedErroredTxs map[string]pendingTx // finalized and errored transactions held onto for status
+	broadcastedProcessedTxs map[string]pendingTx // broadcasted and processed transactions that may require retry and bumping
```
amit-momin: Thanks for the change! Kind of bothered me I didn't do this with the initial change but also didn't want to open a separate PR for it

pkg/solana/chain.go
Farber98: thanks, I'll remove it

pkg/solana/txm/txm.go
Farber98: done

pkg/solana/txm/txm.go
```diff
@@ -175,197 +180,197 @@ func (txm *Txm) run() {
...
-		txm.txs.OnError(sig, txm.cfg.TxRetentionTimeout(), TxFailReject) //nolint // no need to check error since only incrementing metric here
-		return solanaGo.Transaction{}, "", solanaGo.Signature{}, fmt.Errorf("tx failed initial transmit: %w", initSendErr)
+// buildTx builds and signs the transaction with the appropriate compute unit price.
+func (txm *Txm) buildTx(ctx context.Context, msg pendingTx, retryCount int) (solanaGo.Transaction, error) {
+	// Set compute unit limit if specified
+	if msg.cfg.ComputeUnitLimit != 0 {
+		if err := fees.SetComputeUnitLimit(&msg.tx, fees.ComputeUnitLimit(msg.cfg.ComputeUnitLimit)); err != nil {
+			return solanaGo.Transaction{}, fmt.Errorf("failed to add compute unit limit instruction: %w", err)
+		}
 	}
```
amit-momin: I think previously we were adding the limit outside of `buildTx` so think this might introduce a behavior change. Think we would add the limit to `msg.tx` every time we call this method including in the retry method. Not sure if `SetComputeUnitLimit` is idempotent unless your testing confirmed that. Otherwise, think we need to move this back to `sendWithRetry` before we build the tx for the first time. Or we move this below and only add it to the tx copy.

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
+				break
 			}
+			txm.processConfirmations(ctx, client)
 
-			// batch sigs no more than MaxSigsToConfirm each
-			sigsBatch, err := utils.BatchSplit(sigs, MaxSigsToConfirm)
-			if err != nil { // this should never happen
-				txm.lggr.Fatalw("failed to batch signatures", "error", err)
-				break // exit switch
+			// In case all txes where confirmed and there's nothing to rebroadcast.
```
amit-momin: nit: 
```suggestion
			// In case all txes were confirmed and there's nothing to rebroadcast.
```

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
+		return
+	}
+	blockhash, err := client.LatestBlockhash(ctx)
+	if err != nil {
+		txm.lggr.Errorw("failed to get blockhash", "error", err)
+		return
+	}
+	// Rebroadcast all expired txes
+	for _, tx := range txm.txs.ListAllExpiredBroadcastedTxs(currHeight) {
+		txm.lggr.Infow("transaction expired, rebroadcasting", "id", tx.id, "signature", tx.signatures)
```
amit-momin: I think we can maybe make this a Debug log instead to avoid this filling up the logs. Even some of the confirmation logs I updated in my other PR to be Debug unless it's an error

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
+			continue
+		}
+		// Overwrite blockhash and lastValidBlockHeight with latest values so that the transaction can be rebroadcasted an accepted.
+		tx.tx.Message.RecentBlockhash = blockhash.Value.Blockhash
+		rebroadcastTx := pendingTx{
+			tx:                   tx.tx,
+			id:                   tx.id,
+			rebroadcastCount:     tx.rebroadcastCount + 1,
+			lastValidBlockHeight: blockhash.Value.LastValidBlockHeight,
+		}
```
amit-momin: I think `sendWithRetry` also expects `cfg` to be set in the message
```suggestion
		rebroadcastTx := pendingTx{
			tx:                   tx.tx,
			id:                   tx.id,
			cfg:                  tx.cfg,
			rebroadcastCount:     tx.rebroadcastCount + 1,
			lastValidBlockHeight: blockhash.Value.LastValidBlockHeight,
		}
```

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
+		// call sendWithRetry directly to avoid enqueuing
+		// using same id in case it was set by caller and we need to maintain it.
+		_, _, _, err = txm.sendWithRetry(ctx, rebroadcastTx)
+		if err != nil {
+			// TODO: add prebroadcast error handling when merged https://github.com/smartcontractkit/chainlink-solana/pull/936
+			txm.lggr.Errorw("failed to rebroadcast transaction", "id", tx.id, "error", err)
+			continue
+		}
+
+		txm.lggr.Infow("rebroadcast transaction sent", "id", tx.id)
```
amit-momin: Same with this one. Think we can mark this as Debug too

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
+			txm.lggr.Errorw("no signatures found for expired transaction", "id", tx.id)
+			continue
+		}
+		_, err := txm.txs.Remove(tx.signatures[0]) // only picking signature[0]. Remove func removes all related remaining signatures.
+		if err != nil {
+			txm.lggr.Errorw("failed to remove expired transaction", "id", tx.id, "error", err)
+			continue
+		}
+		// Overwrite blockhash and lastValidBlockHeight with latest values so that the transaction can be rebroadcasted an accepted.
+		tx.tx.Message.RecentBlockhash = blockhash.Value.Blockhash
```
amit-momin: Didn't we want to set `RecentBlockhash` and `lastValidBlockHeight ` in `sendWithRetry`? Otherwise, we wouldn't have `lastValidBlockHeight` set in any `msg` to pick up for expiration. In that case we can remove the blockhash logic from here and just rely on the `sendWithRetry` call later

pkg/solana/txm/txm.go
```diff
@@ -377,135 +382,220 @@ func (txm *Txm) confirm() {
...
 
-			// batch sigs no more than MaxSigsToConfirm each
-			sigsBatch, err := utils.BatchSplit(sigs, MaxSigsToConfirm)
-			if err != nil { // this should never happen
-				txm.lggr.Fatalw("failed to batch signatures", "error", err)
-				break // exit switch
+			// In case all txes where confirmed and there's nothing to rebroadcast.
+			// This check saves making 2 RPC calls (slot height + blockhash) when there's nothing to process.
+			// Passing MaxUint64 as currHeight to ListAllExpiredBroadcastedTxs will return all broadcasted txs.
+			if len(txm.txs.ListAllExpiredBroadcastedTxs(math.MaxUint64)) == 0 {
```
amit-momin: If we move the blockhash logic to `sendWithRetry`, this would only help avoid the 1 RPC call for slot height. I think it's good we're conscious about RPC calls but think it's probably extremely rare we don't have something in broadcast state at any given time. And if we have quite a few, this might be doing more work than it's saving since it'd parse every broadcasted tx and return it. I think we can nix this check. If you think it's helping more than I'm thinking, we should at least move it into `rebroadcastExpiredTxs` so we're not doing the extra validation if the feature is turned off.

pkg/solana/txm/txm.go
```diff
@@ -602,19 +692,20 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 		}
 	}
 
-	// Use transaction ID provided by caller if set
-	id := uuid.New().String()
-	if txID != nil && *txID != "" {
-		id = *txID
-	}
-	msg := pendingTx{
+	msg := &pendingTx{
```
amit-momin: Is there a benefit to using a pointer to `pendingTx` here if we're just dereferencing it before sending it over the channel?

pkg/solana/txm/txm.go
Farber98: Yes, maybe was trying to overengineer this. Let's keep it simple right now and remove it

pkg/solana/txm/txm.go
Farber98: thanks for this!

pkg/solana/txm/txm.go
Farber98: My bad! Fixing this.

I think we have some tradeoffs to make here as the first blockhash is set within Enqueue callers.
- If we want to reduce an RPC call per tx, we have the following alternatives:
	- **pass lastValidBlockheight to Enqueue:** will let us use it within `sendWithRetry`
	- **move blockhash owning to txm:** involves moving the balance check that needs a blockhash from callers to txm

- If we don't mind making an extra RPC call per tx:
	- We can just request an extra blockhash every time we call sendWithRetry and use that value 


I took the **pass lastValidBlockheight to Enqueue** approach in the changes pushed. In my head saving 1 RPC call per tx seems a good idea.

If you think I'm overengineering this or adding unnecessary complexity, we can stick with the duplicate RPC call and revisit this later when we really need to improve performance

pkg/solana/txm/txm.go
Farber98: added to the copy. Thanks for this one!

pkg/solana/txm/txm.go
Farber98: quick call notes: 
- We will make the additional RPC call per tx. Given the fast nature of solana, it doesn't hurt to have the blockhash assignment close to where the tx is actually sent

pkg/solana/txm/txm.go
```diff
@@ -377,135 +399,205 @@ func (txm *Txm) confirm() {
...
+}
 
-			// waitgroup for processing
-			var wg sync.WaitGroup
+// handleErrorSignatureStatus handles the case where a transaction signature has an error on-chain.
+// If the error is BlockhashNotFound and expiration rebroadcast is enabled, it skips error handling to allow rebroadcasting.
+// Otherwise, it marks the transaction as errored.
+func (txm *Txm) handleErrorSignatureStatus(sig solanaGo.Signature, status *rpc.SignatureStatusesResult) {
+	// We want to rebroadcast rather than drop tx if expiration rebroadcast is enabled when blockhash was not found.
+	if status.Err != nil && status.Err == client.ErrBlockhashNotFound && txm.cfg.TxExpirationRebroadcast() {
```
amit-momin: Huh I didn't realize we had these error patterns already defined in the client package. But don't think we can just do an `==` here for this match. The `status.Err` interface doesn't return like a normal RPC string. Think it looks something like [this](https://github.com/smartcontractkit/chainlink-solana/blob/c9ef4b36eb7b806cb3fed4a6e418f48db0305b85/pkg/solana/txm/txm_internal_test.go#L334) so don't think we can use these patterns. We probably have to do something like [this](https://github.com/smartcontractkit/chainlink-solana/blob/c9ef4b36eb7b806cb3fed4a6e418f48db0305b85/pkg/solana/txm/txm.go#L726) then a `strings.Contains`

pkg/solana/txm/txm_internal_test.go
```diff
@@ -163,6 +163,12 @@ func TestTxm(t *testing.T) {
 
 			// happy path (send => simulate success => tx: nil => tx: processed => tx: confirmed => finalized => done)
 			t.Run("happyPath", func(t *testing.T) {
+				mc.On("LatestBlockhash", mock.Anything).Return(&rpc.GetLatestBlockhashResult{
+					Value: &rpc.LatestBlockhashResult{
+						LastValidBlockHeight: 100,
+						Blockhash:            solana.Hash{},
+					},
+				}, nil).Once()
```
amit-momin: Looks like we mock the `LatestBlockhash` the same way for all of the tests. Could we move this one layer higher so we only have to mock it once? We can ditch the `.Once()` unless some tests don't get that far then I'm ok with replacing it with `.Maybe()`

pkg/solana/txm/txm_internal_test.go
```diff
@@ -908,6 +983,12 @@ func TestTxm_compute_unit_limit_estimation(t *testing.T) {
 	)
 
 	t.Run("simulation_succeeds", func(t *testing.T) {
+		mc.On("LatestBlockhash", mock.Anything).Return(&rpc.GetLatestBlockhashResult{
```
amit-momin: Same for this test

pkg/solana/txm/txm_internal_test.go
```diff
@@ -1090,3 +1177,390 @@ func addSigAndLimitToTx(t *testing.T, keystore SimpleKeystore, pubkey solana.Pub
...
 }
+
+func TestTxm_ExpirationRebroadcast(t *testing.T) {
+	t.Parallel()
+	estimator := "fixed"
+	id := "mocknet-" + estimator + "-" + uuid.NewString()
+	cfg := config.NewDefault()
+	cfg.Chain.FeeEstimatorMode = &estimator
+	cfg.Chain.TxConfirmTimeout = relayconfig.MustNewDuration(5 * time.Second)
+	cfg.Chain.TxRetentionTimeout = relayconfig.MustNewDuration(10 * time.Second) // Enable retention to keep transactions after finality and be able to check.
```
amit-momin: nit: think this comment is maybe incomplete?
```suggestion
	cfg.Chain.TxRetentionTimeout = relayconfig.MustNewDuration(10 * time.Second) // Enable retention to keep transactions after finality and be able to check their statuses.
```

pkg/solana/txm/txm_internal_test.go
```diff
@@ -1090,3 +1177,390 @@ func addSigAndLimitToTx(t *testing.T, keystore SimpleKeystore, pubkey solana.Pub
...
+			}
+		}
+
+		txm, _, mkey := setupTxmTest(txExpirationRebroadcast, latestBlockhashFunc, slotHeightFunc, sendTxFunc, statuses)
+
+		tx, _ := getTx(t, 0, mkey)
+		txID := "test-rebroadcast"
+		assert.NoError(t, txm.Enqueue(ctx, t.Name(), tx, &txID))
+		wg.Wait()
+		time.Sleep(2 * time.Second) // Sleep to allow for rebroadcasting
```
amit-momin: You don't have to use this if 2 second sleep is deterministic so rebroadcast always happen in that time. If not, maybe it's safer to use something like `waitFor(t, waitDuration, txm, prom, empty)` which we use in other tests above to wait till all tx have gone through.

pkg/solana/txm/txm_internal_test.go
```diff
@@ -1090,3 +1177,390 @@ func addSigAndLimitToTx(t *testing.T, keystore SimpleKeystore, pubkey solana.Pub
...
+	})
+
+	t.Run("WithMultipleRebroadcast", func(t *testing.T) {
+		txExpirationRebroadcast := true
+		statuses := map[solana.Signature]func() *rpc.SignatureStatusesResult{}
+
+		slotHeightFunc := func() (uint64, error) {
+			return uint64(1500), nil
+		}
+		// Mock LatestBlockhash to return a invalid blockhash in the first 3 attempts (initial + 2 rebroadcasts)
```
amit-momin: nit: typo
```suggestion
		// Mock LatestBlockhash to return an invalid blockhash in the first 3 attempts (initial + 2 rebroadcasts)
```

pkg/solana/txm/txm_race_test.go
```diff
@@ -8,7 +8,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gagliardetto/solana-go"
 	solanaGo "github.com/gagliardetto/solana-go"
```
amit-momin: Looks like we have duplicate imports here

pkg/solana/txm/txm_race_test.go
```diff
@@ -84,7 +85,12 @@ func TestTxm_SendWithRetry_Race(t *testing.T) {
 
 	t.Run("delay in rebroadcasting tx", func(t *testing.T) {
 		client := clientmocks.NewReaderWriter(t)
-		// client mock
+		client.On("LatestBlockhash", mock.Anything).Return(&rpc.GetLatestBlockhashResult{
```
amit-momin: Like in the other tests, I think we can move this mock to the top level since we return the same thing for all of them

pkg/solana/txm/txm_internal_test.go
Farber98: ohh yes removing the once makes that generalization work 👍

PR #930, opened 2024-11-18
Disable MultiNode

DylanTinianov:
### Description

Disable MultiNode in Smoke Tests by default since heads are not fully produced in testing causing integration tests to be flaky.


aalu1418 (commented):
why are we disabling a feature?

the core side tests run the same exact tests as the solana side

if there is flakiness, let's find/fix the root cause please

dhaidashenko (approved):

PR #932, opened 2024-11-19
Move mockery configs to single config file

EasterTheBunny:
- Move mockery configurations to config file
- Update go version in build script


aalu1418 (dismissed):

Makefile
```diff
@@ -103,6 +103,7 @@ gomodtidy: gomods
 .PHONY: mockery
 mockery: $(mockery) ## Install mockery.
 	go install github.com/vektra/mockery/v2@v2.43.2
+	mockery
```
jmank88: This might be surprising. Should at least update the comment to be consistent.

On the other hand, I think it would be expected for this to be included in `generate` down below.

EasterTheBunny: generate should be running the make mockery command which should install mockery and run it

EasterTheBunny: both are named the same so it's a bit confusing

jmank88: But the comment is currently inconsistent, and it may be better to align with core anyways:
https://github.com/smartcontractkit/chainlink/blob/develop/GNUmakefile#L148-L149
https://github.com/smartcontractkit/chainlink/blob/develop/GNUmakefile#L98-L100

EasterTheBunny: Done

aalu1418 (approved):

jmank88 (approved):

skippaDaBitFlippa (approved):

PR #933, opened 2024-11-19
Update Solana Client for LogPoller

DylanTinianov:
### Description
Added the following methods required for LogPoller
```
GetSignaturesForAddressWithOpts(ctx context.Context, address solana.Publickey, opts *GetSignaturesForAddressOpts) ([]*solana.TransactionSignature, error)
GetTransaction(ctx context.Context, txHash solana.Signature, opts *solana.GetTransactionOpts) (*GetTransactionResult, err error)
GetBlocks(ctx context.Context, startSlot uint64, endSlot *uint64) (out rpc.BlocksResult, err error) 
```

Ticket: https://smartcontract-it.atlassian.net/browse/NONEVM-743


reductionista:
I notice there is a `GetBlocksWithLimit()` method here, but we don't have `GetBlocks()`.  @EasterTheBunny 's PR uses the latter:
https://github.com/smartcontractkit/chainlink-solana/pull/911/files#diff-899d462b4e5fef3021d2bfebde4de451f3e9294185c3e17c14322d5595dbac89R268

This makes me realize though, maybe we should be using `GetBlocksWithLimit`? At least one of these PR's needs to change, I'm not sure off the top of my head which one but I'll leave an update when we get it figured out.

reductionista:
> I notice there is a `GetBlocksWithLimit()` method here, but we don't have `GetBlocks()`. @EasterTheBunny 's PR uses the latter: [#911 (files)](https://github.com/smartcontractkit/chainlink-solana/pull/911/files#diff-899d462b4e5fef3021d2bfebde4de451f3e9294185c3e17c14322d5595dbac89R268)
> 
> This makes me realize though, maybe we should be using `GetBlocksWithLimit`? At least one of these PR's needs to be updated, I'm not sure off the top of my head which one but I'll leave an update when we get it figured out.

Ok yeah, actually we do need `GetBlocks()`. `GetBlocksWithLimit` doesn't allow you to specify an upper slot limit only lower. @DylanTinianov can you add that one to this PR?

pkg/solana/client/client.go
```diff
@@ -131,6 +165,18 @@ func (c *Client) GetAccountInfoWithOpts(ctx context.Context, addr solana.PublicK
...
+	defer done()
+
+	ctx, cancel := context.WithTimeout(ctx, c.contextDuration)
+	defer cancel()
+	if commitment == nil {
+		commitment = &c.commitment
+	}
+	return c.rpc.GetBlocks(ctx, startSlot, endSlot, *commitment)
+}
+
```
reductionista, line 186: I'm guessing this is still a WIP, but just in case: we should probably add a test for `GetBlocks` (assuming the others are already all there) for completeness.

reductionista (approved):

aalu1418 (approved):

PR #935, opened 2024-11-20
Solana ChainWriter

silaslenihan:
### Description

 
Related Tickets: 
[NONEVM-753](https://smartcontract-it.atlassian.net/browse/NONEVM-753)
[NONEVM-963](https://smartcontract-it.atlassian.net/browse/NONEVM-963)
[NONEVM-964](https://smartcontract-it.atlassian.net/browse/NONEVM-964)

### Resolves Dependencies

[NONEVM-753]: https://smartcontract-it.atlassian.net/browse/NONEVM-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NONEVM-963]: https://smartcontract-it.atlassian.net/browse/NONEVM-963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NONEVM-964]: https://smartcontract-it.atlassian.net/browse/NONEVM-964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


integration-tests/relayinterface/lookups_test.go
```diff
@@ -0,0 +1,431 @@
...
+			Name:      "TestPDA",
+			PublicKey: chainwriter.AccountConstant{Name: "ProgramID", Address: programID.String()},
+			Seeds: []chainwriter.Lookup{
+				chainwriter.AccountConstant{Name: "seed", Address: seed.String()},
+			},
+			IsSigner:   false,
+			IsWritable: true,
+		}
+
+		ctx := context.Background()
```
amit-momin: nit: think we should use a test context for all of these sub-tests `ctx := tests.Context(t)`.

pkg/solana/chainwriter/ccip_example_config.go
```diff
@@ -0,0 +1,340 @@
...
+
+	executeConfig := MethodConfig{
+		FromAddress: userAddress,
+		InputModifications: commoncodec.ModifiersConfig{
+			// remove merkle root since it isn't a part of the on-chain type
+			&commoncodec.DropModifierConfig{
+				Fields: []string{"Message.ExtraArgs.MerkleRoot"},
+			},
```
amit-momin: Do we still want the drop modifier in the config here? Think I remember you mentioning these are dropped automatically which is why we didn't have to define it in the chainwriter unit tests

silaslenihan: Good point, removed

integration-tests/relayinterface/lookups_test.go
silaslenihan: fixed

pkg/solana/chainwriter/ccip_example_config.go
```diff
@@ -0,0 +1,333 @@
...
+				IsSigner:   true,
+				IsWritable: false,
+			},
+		},
+		DebugIDLocation: "",
+	}
+
+	chainWriterConfig := ChainWriterConfig{
+		Programs: map[string]ProgramConfig{
+			"ccip-router": {
```
amit-momin, line 323: To make this example accurate, I think this program name needs to be a public key

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,329 @@
...
+		return errorWithDebugID(fmt.Errorf("error enqueuing transaction: %w", err), debugID)
+	}
+
+	return nil
+}
+
+var (
+	_ services.Service     = &SolanaChainWriterService{}
+	_ types.ContractWriter = &SolanaChainWriterService{}
+)
```
amit-momin: I think this is duplicated so we can remove this.

pkg/solana/chainwriter/helpers.go
```diff
@@ -0,0 +1,218 @@
...
+		return "", err
+	}
+
+	// there should only be one debug ID, others will be ignored.
+	debugID := string(debugIDList[0])
+
+	return debugID, nil
+}
+
+func GetValueAtLocation(args any, location string) ([][]byte, error) {
```
amit-momin: Noticed this is essentially the same as `GetValuesAtLocation` and only really used in `GetDebugIDAtLocation` as far as I can tell. Couldn't we just use `GetValuesAtLocation` for that method as well?

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (alt AccountsFromLookupTable) Resolve(_ context.Context, _ any, derivedTableMap map[string]map[string][]*solana.AccountMeta, _ client.Reader) ([]*solana.AccountMeta, error) {
+	// Fetch the inner map for the specified lookup table name
+	innerMap, ok := derivedTableMap[alt.LookupTableName]
+	if !ok {
+		return nil, fmt.Errorf("lookup table not found: %s", alt.LookupTableName)
+	}
+
+	var result []*solana.AccountMeta
```
amit-momin, line 116: nit: think we can initialize this with `len(innerMap)` to avoid reallocations
```suggestion
	result := make([]*solana.AccountMeta, 0, len(innerMap))
```

amit-momin: Actually never mind, realized this isn't one to one

pkg/solana/chainwriter/ccip_example_config.go
silaslenihan: The Program Config's key can be any string, it doesn't have to be a PublicKey. The program address is passed through the `toAddress` param in `SubmitTransaction`

pkg/solana/chainwriter/helpers.go
silaslenihan: good call, fixed

pkg/solana/chainwriter/chain_writer.go
silaslenihan: good catch! fixed

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,356 @@
...
+			}
+		}
+	}
+
+	return seedBytes, nil
+}
+
+// generatePDAs generates program-derived addresses (PDAs) from public keys and seeds.
+func generatePDAs(publicKeys []*solana.AccountMeta, seeds [][]byte, lookup PDALookups) ([]*solana.AccountMeta, error) {
+	if len(seeds) > 1 && len(publicKeys) > 1 {
```
amit-momin: Not really feedback but curious about this validation and this might just be my lack of knowledge on PDAs. So it's ok to process multiple publicKeys as long as we only have a single seed? And it's ok to process multiple seeds as long as we only have a single publicKey? Also on a separate note, are no seeds a valid scenario too?

silaslenihan: Honestly, I'm having trouble remembering why I added that validation in... I think it is okay to have multiple public keys and multiple seeds, each program ID would be used to generate a PDA with all the seeds and then all the addresses returned from the multiple PDAs would be returned by ChainWriter. I'm going to remove that validation. 

> Also on a separate note, are no seeds a valid scenario too?

I think technically no seeds is a valid scenario as long as a publicKey is supplied. The publicKey is also a seed but it just comes first and always corresponds to the overarching program / account address

pkg/solana/chainwriter/ccip_example_config.go
amit-momin: Ah ok then I think there's a change needed in `SubmitTransaction`. I'll add a comment at the spot. I ran into this issue when writing the unit tests which is why I changed the program config key to an address.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+	filteredLookupTableMap := s.FilterLookupTableAddresses(accounts, derivedTableMap, staticTableMap)
+
+	// Fetch latest blockhash
+	blockhash, err := s.reader.LatestBlockhash(ctx)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error fetching latest blockhash: %w", err), debugID)
+	}
+
+	// Prepare transaction
+	programID, err := solana.PublicKeyFromBase58(contractName)
```
amit-momin: I think this should be `toAddress` based on our conversation in the other thread.
```suggestion
	programID, err := solana.PublicKeyFromBase58(toAddress)
```

silaslenihan: Yep, good catch

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -0,0 +1,691 @@
...
+			return true
+		}), &txID).Return(nil).Once()
+
+		args := map[string]interface{}{
+			"lookupTable":  chainwriter.GetRandomPubKey(t).Bytes(),
+			"lookup_table": account2.Bytes(),
+			"seed1":        seed1,
+			"seed2":        seed2,
+		}
+		submitErr := cw.SubmitTransaction(ctx, "6AfuXF6HapDUhQfE4nQG9C1SGtA1YjP3icaJyRfU4RyE", "initializeLookupTable", args, txID, programID.String(), nil, nil)
```
amit-momin: Also based on our other conversation, I think we can change this program config key back to a human readable name. And think you'll also need to update this to pass a `toAddress`

silaslenihan: fixed!

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -0,0 +1,690 @@
...
+		txID := uuid.NewString()
+		args := map[string]interface{}{}
+		submitErr := cw.SubmitTransaction(ctx, "contractReaderInterface", "initializeLookupTable", args, txID, programID.String(), nil, nil)
+		require.Error(t, submitErr)
+	})
+
+	t.Run("fails if invalid contract name provided", func(t *testing.T) {
+		txID := uuid.NewString()
+		args := map[string]interface{}{}
+		submitErr := cw.SubmitTransaction(ctx, "contractReaderInterface", "initializeLookupTable", args, txID, programID.String(), nil, nil)
```
amit-momin: Huh seems like tests are passing but this might be picking up on a different error. I think we'll need to pass in a bad contract name here for this invalid case.
```suggestion
		submitErr := cw.SubmitTransaction(ctx, "badContract", "initializeLookupTable", args, txID, programID.String(), nil, nil)
```

contracts/artifacts/localnet/write_test-keypair.json
```diff
@@ -0,0 +1,6 @@
+[
```
Farber98: Where is this being used?

contracts/programs/contract-reader-interface/src/lib.rs
```diff
@@ -37,6 +50,34 @@ pub struct Initialize<'info> {
     pub system_program: Program<'info, System>,
 }
 
+#[derive(Accounts)]
+pub struct InitializeLookupTableData<'info> {
+    /// PDA for LookupTableDataAccount, derived from seeds and created by the System Program
+    #[account(
+        init,
+        payer = admin,
+        space = size_of::<LookupTableDataAccount>() + 8,
```
Farber98, line 59: just curious, what is +8 here?

pkg/solana/chainwriter/ccip_example_config.go
```diff
@@ -0,0 +1,333 @@
...
+				Methods: map[string]MethodConfig{
+					"execute": executeConfig,
+					"commit":  commitConfig,
+				},
+				IDL: executionReportSingleChainIDL,
+			},
+		},
+	}
+	fmt.Println(chainWriterConfig)
```
Farber98: keeping this print?

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,356 @@
...
+
+	ag_binary "github.com/gagliardetto/binary"
+	"github.com/gagliardetto/solana-go"
+	addresslookuptable "github.com/gagliardetto/solana-go/programs/address-lookup-table"
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+)
+
+type Lookup interface {
```
Farber98, line 16: love how you abstracted the lookup functionality 🔥

Farber98 (commented):
Excellent job Silas! The structure looks really good and the complexity is well hidden behind the interface 🙌 . I feel I'm missing some context to review some details, will ask you some questions on Slack

pkg/solana/chainwriter/ccip_example_config.go
silaslenihan: Just did this so the compiler wouldn't complain about the unused variable

contracts/programs/contract-reader-interface/src/lib.rs
silaslenihan: That accounts for the 8-byte discriminator I believe

contracts/artifacts/localnet/write_test-keypair.json
silaslenihan: Good catch, this is an old relic of some deleted code, I'll get rid of it.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+		codecs: codecs,
+	}, nil
+}
+
+func parseIDLCodecs(config ChainWriterConfig) (map[string]types.Codec, error) {
+	codecs := make(map[string]types.Codec)
+	for program, programConfig := range config.Programs {
+		var idl codec.IDL
+		if err := json.Unmarshal([]byte(programConfig.IDL), &idl); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal IDL: %w", err)
```
prashantkumar1982: nit: Also log the program name with the error message, for better debuggability.
Also applies to other places with same pattern.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+	accounts, err := GetAddresses(ctx, args, methodConfig.Accounts, derivedTableMap, s.reader)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error resolving account addresses: %w", err), debugID)
+	}
+
+	// Filter the lookup table addresses based on which accounts are actually used
+	filteredLookupTableMap := s.FilterLookupTableAddresses(accounts, derivedTableMap, staticTableMap)
+
+	// Fetch latest blockhash
+	blockhash, err := s.reader.LatestBlockhash(ctx)
```
prashantkumar1982, line 292: In my view, fetching and setting the latest blockHash should be entirely encapsulated inside the TXM.
No point in letting a caller of TXM setting it.

For one, the caller has no idea when will the TXM actually process this Tx. It will go into a queue, and get picked up when TXM is ready, which may be pretty late, causing this blockHash to get stale.
Secondly, I think the tXM anyways sets recent blockhashes on all txs that it processes. So this is duplicate.


@Farber98 / @amit-momin  what do you think?

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,356 @@
...
+		addresses = append(addresses, &solana.AccountMeta{
+			PublicKey:  address,
+			IsSigner:   lookup.IsSigner,
+			IsWritable: lookup.IsWritable,
+		})
+	}
+	return addresses, nil
+}
+
+func (s *SolanaChainWriterService) ResolveLookupTables(ctx context.Context, args any, lookupTables LookupTables) (map[string]map[string][]*solana.AccountMeta, map[solana.PublicKey]solana.PublicKeySlice, error) {
```
prashantkumar1982: Out of curiosity, why is this and the LoadTable methods in this file, and not in the chain_writer file?
They are members of SolanaChainWriterService, so that seems to be a better place.
Also, are both of these functions internal to this struct? If yes, can we not export them? Otherwise it appears to the reader that they are used from outside too.

pkg/solana/chainwriter/lookups.go
amit-momin: Noticed this validation is still here unless there was a reason for this. I also wanted to mention that we might want to validate the seeds here as well. According to the [docs](https://solana.com/developers/courses/native-onchain-development/program-derived-addresses) on PDAs:
```A PDA allows a maximum of 16 seeds, with each seed limited to 32 bytes in length```

Comment (silaslenihan) [reply pkg/solana/chainwriter/lookups.go]: Good call out! Thought I removed that but guess I never got around to it. Also, I added validation for seed maximum [here](https://github.com/smartcontractkit/chainlink-solana/blob/135e6b9ecfaf7d055ea933685e8a3d274a448902/pkg/solana/chainwriter/lookups.go#L250-L252) and seed length [here](https://github.com/smartcontractkit/chainlink-solana/blob/135e6b9ecfaf7d055ea933685e8a3d274a448902/pkg/solana/chainwriter/lookups.go#L225-L230)

File: pkg/solana/chainwriter/lookups.go [hunk 18]
```diff
@@ -0,0 +1,362 @@
...
+			}
+		}
+	}
+
+	return seedBytes, nil
+}
+
+// generatePDAs generates program-derived addresses (PDAs) from public keys and seeds.
+func generatePDAs(publicKeys []*solana.AccountMeta, seeds [][]byte, lookup PDALookups) ([]*solana.AccountMeta, error) {
+	if len(seeds) >= 16 {
```
amit-momin: Since 16 seeds is allowed, think this check should just be >
```suggestion
	if len(seeds) > 16 {
```

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+		solana.TransactionAddressTables(filteredLookupTableMap),
+	)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error constructing transaction: %w", err), debugID)
+	}
+
+	// Enqueue transaction
+	if err = s.txm.Enqueue(ctx, accounts[0].PublicKey.String(), tx, &transactionID); err != nil {
+		return errorWithDebugID(fmt.Errorf("error enqueuing transaction: %w", err), debugID)
+	}
```
Farber98: Enqueue now needs `lastValidBlockHeight` param too. You can access it through `blockhash.Value.LastValidBlockHeight`

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,361 @@
...
+	for _, seed := range lookup.Seeds {
+		if lookupSeed, ok := seed.(AccountLookup); ok {
+			// Get value from a location (This doens't have to be an address, it can be any value)
+			bytes, err := GetValuesAtLocation(args, lookupSeed.Location)
+			if err != nil {
+				return nil, fmt.Errorf("error getting address seed: %w", err)
+			}
+			// validate seed length
+			for _, b := range bytes {
+				if len(b) > maxSeedLength {
```
amit-momin: nit: noticed this while looking into PDAs for ChainReader but think the solana go package has public values we can use for this instead
```suggestion
				if len(b) > solana.MaxSeedLength {
```

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,361 @@
...
+			}
+		}
+	}
+
+	return seedBytes, nil
+}
+
+// generatePDAs generates program-derived addresses (PDAs) from public keys and seeds.
+func generatePDAs(publicKeys []*solana.AccountMeta, seeds [][]byte, lookup PDALookups) ([]*solana.AccountMeta, error) {
+	if len(seeds) > 16 {
```
amit-momin: nit: same for here
```suggestion
	if len(seeds) > solana.MaxSeeds {
```

pkg/solana/chainwriter/lookups.go
silaslenihan: ah good find! fixed

ilija42 (commented):
Can we rename chain writer to contract writer everywhere before this is merged? Should be a lot easier to do now than later. Common already has it as contract writer

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,360 @@
...
+
+type InternalField struct {
+	Type     reflect.Type
+	Location string
+}
+
+// LookupTables represents a list of lookup tables that are used to derive addresses for a program.
+type LookupTables struct {
+	DerivedLookupTables []DerivedLookupTable
+	StaticLookupTables  []string
```
ilija42: Can't you have an actual solana.PublicKey type here? I think it has json interface methods defined, you get more instant validation if you use the actual type

silaslenihan: Good point, fixed.

pkg/solana/chainwriter/ccip_example_config.go
```diff
@@ -0,0 +1,335 @@
+package chainwriter
+
+import (
+	"fmt"
+
+	"github.com/gagliardetto/solana-go"
+)
+
```
ilija42, line 6: Nice! Love this whole file

pkg/solana/chainwriter/ccip_example_config.go
ilija42: I think this is the idiomatic way to do this 
```
executeConfig := MethodConfig{
_ = executeConfig

```

silaslenihan: Ah true, fixed

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+type ChainWriterConfig struct {
+	Programs map[string]ProgramConfig
+}
+
+type ProgramConfig struct {
+	Methods map[string]MethodConfig
+	IDL     string
+}
+
+type MethodConfig struct {
```
ilija42, line 54: Wdyt about storing configs separately from the service similar to EVM Chain Components impl.?

silaslenihan: What's the advantage of doing it that way?

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,354 @@
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("error fetching static lookup table address: %w", err)
+		}
+		staticTableMap[staticTable] = addressses
+	}
+
+	return derivedTableMap, staticTableMap, nil
+}
+
+func (s *SolanaChainWriterService) LoadTable(ctx context.Context, args any, rlt DerivedLookupTable, reader client.Reader, derivedTableMap map[string]map[string][]*solana.AccountMeta) (map[string]map[string][]*solana.AccountMeta, []*solana.AccountMeta, error) {
```
prashantkumar1982: nit: The param "reader" is member of SolanaChainWriterService already, so no need to pass it as a parameter.

prashantkumar1982: Also, the param "derivedTableMap" is unused in this function.

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,354 @@
...
+		for _, addr := range addresses {
+			resultMap[rlt.Name][addressMeta.PublicKey.String()] = append(resultMap[rlt.Name][addressMeta.PublicKey.String()], &solana.AccountMeta{
+				PublicKey:  addr,
+				IsSigner:   addressMeta.IsSigner,
+				IsWritable: addressMeta.IsWritable,
+			})
+		}
+
+		// Add the current lookup table address to the list of metas
+		lookupTableMetas = append(lookupTableMetas, addressMeta)
```
prashantkumar1982: Why do we have this?
I see there's just 1 caller, which ignores this return value.

pkg/solana/chainwriter/chain_writer.go
ilija42: Its best practice, we have CR and multinode configs in pkg/solana. Configs should have unit tests and validations, so keeping them in a separate package is handy to have all of that in the same place. You can see [some examples here](https://github.com/smartcontractkit/chainlink-solana/tree/NONEVM-702-cleanup-codec-usage/pkg/solana/config).

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+		blockhash.Value.Blockhash,
+		solana.TransactionPayer(feePayer),
+		solana.TransactionAddressTables(filteredLookupTableMap),
+	)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error constructing transaction: %w", err), debugID)
+	}
+
+	// Enqueue transaction
+	if err = s.txm.Enqueue(ctx, accounts[0].PublicKey.String(), tx, &transactionID, blockhash.Value.LastValidBlockHeight); err != nil {
```
ilija42: Config doesn't have any validation for accounts  lookup having any entries, so wouldn't this panic if one of the methodConfigs has no account lookups listed? You should check for accounts size

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+		// Add to the filtered map if there are any used addresses
+		if len(usedAddresses) > 0 {
+			filteredLookupTables[tableKey] = usedAddresses
+		}
+	}
+
+	return filteredLookupTables
+}
+
+func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *types.TxMeta, value *big.Int) error {
```
ilija42: lint nit
```suggestion
func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, _ *types.TxMeta, _ *big.Int) error {
```

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// getSeedBytes extracts the seeds for the PDALookups.
+// It handles both AddressSeeds (which are public keys) and ValueSeeds (which are byte arrays from input args).
+func getSeedBytes(ctx context.Context, lookup PDALookups, args any, derivedTableMap map[string]map[string][]*solana.AccountMeta, reader client.Reader) ([][]byte, error) {
+	var seedBytes [][]byte
+
+	for _, seed := range lookup.Seeds {
+		if lookupSeed, ok := seed.(AccountLookup); ok {
+			// Get value from a location (This doens't have to be an address, it can be any value)
```
ilija42: lint
```suggestion
			// Get value from a location (This  doens't have to be an address, it can be any value)
```

pkg/solana/chainwriter/helpers.go
```diff
@@ -0,0 +1,198 @@
...
+	PendingAdministrator solana.PublicKey
+	LookupTable          solana.PublicKey
+}
+
+// GetValuesAtLocation parses through nested types and arrays to find all locations of values
+func GetValuesAtLocation(args any, location string) ([][]byte, error) {
+	var vals [][]byte
+	path := strings.Split(location, ".")
+
+	addressList, err := traversePath(args, path)
```
ilija42, line 40: I see

pkg/solana/chainwriter/helpers.go
```diff
@@ -0,0 +1,198 @@
...
+type DataAccount struct {
+	Discriminator        [8]byte
+	Version              uint8
+	Administrator        solana.PublicKey
+	PendingAdministrator solana.PublicKey
+	LookupTable          solana.PublicKey
+}
+
+// GetValuesAtLocation parses through nested types and arrays to find all locations of values
+func GetValuesAtLocation(args any, location string) ([][]byte, error) {
```
ilija42, line 37: Callers in multiple locations are trying to access response slice elem by index without checking the size which is panic prone

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,354 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error resolving addresses for lookup table: %w", err)
+	}
+
+	// Nested map in case the lookup table resolves to multiple addresses
+	resultMap := make(map[string]map[string][]*solana.AccountMeta)
+	var lookupTableMetas []*solana.AccountMeta
+
+	// Iterate over each address of the lookup table
+	for _, addressMeta := range lookupTableAddresses {
```
ilija42: Is it possible to get empty slice of lookupTableAddresses, if it is is that okay? If not maybe throw an error here

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,354 @@
...
+		// Read the full list of addresses from the lookup table
+		addresses, err := getLookupTableAddresses(ctx, reader, addressMeta.PublicKey)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error fetching lookup table address: %w", err)
```
ilija42: does it make sense to also log the table address here?

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+
+3. **PDALookups**:
+  - Generates Program Derived Addresses (PDA) by combining a derived public key with one or more seeds.
+  - Seeds can be `AddressSeeds` (public keys from the input args) or `ValueSeeds` (byte arrays).
+  - Ensures there is only one public key if multiple seeds are provided.
+
+### Error Handling:
+- Errors are wrapped with the `debugID` for easier tracing.
+*/
+// GetAddresses resolves account addresses from various `Lookup` configurations to build the required `solana.AccountMeta` list
```
Farber98: Info already included in the above comment 
```suggestion

```

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+// for Solana transactions.
```
Farber98:
```suggestion

```

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+	for _, accountConfig := range accounts {
+		meta, err := accountConfig.Resolve(ctx, args, derivedTableMap, reader)
+		if err != nil {
+			return nil, err
+		}
+		addresses = append(addresses, meta...)
+	}
+	return addresses, nil
+}
+
```
Farber98, line 157: I think it's worth adding a comment here, something like: 

```suggestion
// FilterLookupTableAddresses takes a list of accounts and two lookup table maps
// (one for derived tables, one for static tables) and filters out any addresses that are
// not used by the accounts. It returns a map of only those lookup table
// addresses that match entries in `accounts`.
```

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codec"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/fees"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/txm"
+)
+
+const ServiceName = "SolanaChainWriter"
+
+type SolanaChainWriterService struct {
```
Farber98, line 26: why are we [doing this](https://github.com/smartcontractkit/chainlink/blob/498b0b8579ad52a8c394fe3cbf55d3a86a8e29a0/core/services/relay/evm/chain_writer.go#L27-L33) differently than EVM?

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
+package chainwriter
```
Farber98, line 1: Are we planning to rename it later to [ContractWriter](https://github.com/smartcontractkit/chainlink-common/blob/db7919d60550c76b37a0d7cc5e694a272cc54bdd/pkg/types/contract_writer.go#L14)? Maybe it's easier to do it now once and for all for Solana? 

I know it's a very annoying nit 🙏

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,324 @@
...
+
+		// Add to the filtered map if there are any used addresses
+		if len(usedAddresses) > 0 {
+			filteredLookupTables[tableKey] = usedAddresses
+		}
+	}
+
+	return filteredLookupTables
+}
+
```
Farber98, line 215: Thinking in terms of users, we should have a comment here so they don't have to dive into the implementation to understand it. Something like: 

```suggestion
// SubmitTransaction builds, encodes, and enqueues a transaction using the provided program
// configuration and method details. It relies on the configured IDL, account lookups, and
// lookup tables to gather the necessary accounts and data. The function retrieves the latest
// blockhash and assigns it to the transaction, so callers do not need to provide one.
//
// Submissions and retries are handled by the underlying transaction manager. If a “debug ID”
// location is configured, SubmitTransaction extracts it from the provided `args` and attaches
// it to errors for easier troubleshooting. Only the first debug ID it encounters will be used.
//
// Parameters:
//   - ctx: The context for cancellation and timeouts.
//   - contractName: Identifies which Solana program config to use from `s.config.Programs`.
//   - method: Specifies which method config to invoke within the chosen program config.
//   - args: Arbitrary arguments that are encoded into the transaction payload and/or used for dynamic address lookups.
//   - transactionID: A unique identifier for the transaction, used for tracking within the transaction manager.
//   - toAddress: The on-chain address (program ID) to which the transaction is directed.
//   - meta: Currently unused; included for interface compatibility.
//   - value: Currently unused; included for interface compatibility.
//
// Returns:
//   An error if any stage of the transaction preparation or enqueueing fails. A nil return
//   indicates that the transaction was successfully submitted to the transaction manager.
```

Farber98 (commented):
some nits and questions

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Actually, I was missing the fact that the accounts list always needs to contain the `feePayer` first, and the system program at the end. I've fixed this

pkg/solana/chainwriter/lookups.go
silaslenihan: ? I see that I misspelled doesn't but is the additional space you suggested intentional?

pkg/solana/chainwriter/helpers.go
silaslenihan: I believe the only panic prone one was the debugID check, which I've fixed now... I think all the ones in `traversePath` are safe. Let me know if you think otherwise

pkg/solana/chainwriter/lookups.go
silaslenihan: I think these checks should be sufficient, no? https://github.com/smartcontractkit/chainlink-solana/blob/9ea5350470c7f93bbcf2ff47af600f4c201faa89/pkg/solana/chainwriter/lookups.go#L346-L352

pkg/solana/chainwriter/lookups.go
silaslenihan: Good call, fixed.

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Awesome! adding this in

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Added!

pkg/solana/chainwriter/chain_writer.go
silaslenihan: To be honest, I don't really understand the distinction between ContractWriter and ChainWriter, maybe @ilija42 can fill in here. My understanding was that the chain-specific version would still be called ChainWriter and the generalized interface is the ContractWriter, but I'm not sure if that's correct nor do I have strong opinions on the semantics either way.

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Is there an advantage of doing it the same way as EVM? The current way I'm doing it allows for Solana ChainWriter specific methods (such as `ResolveLookupTables`, etc. to be tested outside of the package.

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -0,0 +1,692 @@
...
...ce","instructions":[{"name":"initialize","accounts":[{"name":"data","isMut":true,"isSigner":false},{"name":"signer","isMut":true,"isSigner":true},{"name":"systemProgram","isMut":false,"isSigner":false}],"args":[{"name":"testIdx","type":"u64"},{"name":"value","type":"u64"}]},{"name":"initializeLookupTable","accounts":[{"name":"writeDataAccount","isMut":true,"isSigner":false,"docs":["PDA for LookupTableDataAccount, derived from seeds and created by the System Program"]},{"name":"admin","isMut":true,"isSigner":true,"docs":["Admin account that pays for PDA creation and signs the transaction"]},{"name":"systemProgram","isMut":false,"isSigner":false,"docs":["System Program required for PDA creation"]}],"args":[{"name":"lookupTable","type":"publicKey"}]}],"accounts":[{"name":"LookupTableDataAccount","type":{"kind":"struct","fields":[{"name":"version","type":"u8"},{"name":"administrator","type":"publicKey"},{"name":"pendingAdministrator","type":"publicKey"},{"name":"lookupTable","type":"publicKey"}]}},{"name":"DataAccount","type":{"kind":"struct","fields":[{"name":"idx","type":"u64"},{"name":"bump","type":"u8"},{"name":"u64Value","type":"u64"},{"name":"u64Slice","type":{"vec":"u64"}}]}}]}`
```
prashantkumar1982: nit/optional: move into a new json file with better formatting, so easier to read this.

For reference, see this example: https://github.com/smartcontractkit/chainlink-internal-integrations/blob/bfb5305929ea0a6f81924d3181a898b5e8e97a7c/solana/pkg/solana/codec/testutils/testIDL.json#L13

But I understand, this PR is huge, and ok to let it be like this right now. Maybe we can do this work later.

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Completely agree

pkg/solana/chainwriter/chain_writer.go
silaslenihan: fixed.

pkg/solana/chainwriter/lookups.go
silaslenihan: Yeah, I think it makes sense to include these in `chain_writer.go` as opposed to `lookups.go`. I moved them over.

silaslenihan: > Also, are both of these functions internal to this struct? If yes, can we not export them? Otherwise it appears to the reader that they are used from outside too.

I made `loadTable` unexported, however, ResolveLookupTables is used outside of the package in [lookups_test.go](https://github.com/smartcontractkit/chainlink-solana/blob/54e3ae30a247dfc24d3413d1c6ce8a3265558109/integration-tests/relayinterface/lookups_test.go#L307). I think it makes sense to have that test be external since it is an integration test which uses the Solana test validator

pkg/solana/chainwriter/lookups.go
silaslenihan: good catches, fixed

pkg/solana/chainwriter/lookups.go
silaslenihan: You're right, that was unnecessary. I've removed it now.

pkg/solana/chainwriter/chain_writer_test.go
silaslenihan: Added a TODO note with a ticket linked for this https://github.com/smartcontractkit/chainlink-solana/blob/0e212288f68b2b5fe17afa47867645f9f9eeee02/pkg/solana/chainwriter/chain_writer_test.go#L28

prashantkumar1982: I didn't mean to say move it to the different repo. Just to move it to a local json file, so it helps in readability.
Also fixed the ticket

pkg/solana/chainwriter/helpers.go
```diff
@@ -41,12 +42,15 @@ func GetValuesAtLocation(args any, location string) ([][]byte, error) {
...
 	}
-
 	for _, value := range addressList {
 		if byteArray, ok := value.([]byte); ok {
 			vals = append(vals, byteArray)
 		} else if address, ok := value.(solana.PublicKey); ok {
 			vals = append(vals, address.Bytes())
+		} else if num, ok := value.(uint64); ok {
+			buf := make([]byte, 8)
+			binary.LittleEndian.PutUint64(buf, num)
```
prashantkumar1982, line 52: btw, out of curiosity, why are we using Little Endian here?
I saw some onchain code in solana used big endian while encoding data.

pkg/solana/chainwriter/chain_writer_test.go
silaslenihan: Ah I see. I just went ahead and fixed that

pkg/solana/chainwriter/helpers.go
silaslenihan: Borsh encoding uses little endian, not sure where you saw the Big Endian usage?

pkg/solana/chainwriter/chain_writer.go
Farber98: Makes sense!

pkg/solana/chainwriter/chain_writer.go
Farber98: > In my view, fetching and setting the latest blockHash should be entirely encapsulated inside the TXM.
> No point in letting a caller of TXM setting it.

We discussed this with Amit but chose not to move forward yet (implementation track) because there are a few places in the codebase where the caller needs the blockhash to do some operations before passing the transaction to the TXM. 

An example is [this code](https://github.com/smartcontractkit/chainlink-solana/blob/e2ff830d9bc57ada79ba9b2a9e8304c4f6f6b816/pkg/solana/chain.go#L552-L588)

> For one, the caller has no idea when will the TXM actually process this Tx. It will go into a queue, and get picked up when TXM is ready, which may be pretty late, causing this blockHash to get stale.

Agreed. Having the TXM itself fetch and assign the latest blockhash would reduce the risk of staleness and hide complexity from the caller. I'm okay refactoring these other call sites so that the TXM becomes the single source of truth for fetching the current blockhash. That refactor isn’t difficult, but it does require some care to ensure no existing logic depends on a blockhash before calling the TXM. 

@amit-momin if you are okay with this, I can create a ticket inside the [optimization track](https://smartcontract-it.atlassian.net/browse/NONEVM-1023)

> Secondly, I think the tXM anyways sets recent blockhashes on all txs that it processes. So this is duplicate.

Currently, the TXM only fetches and updates the blockhash when it needs to rebroadcast expired transactions ([see code](https://github.com/smartcontractkit/chainlink-solana/blob/e2ff830d9bc57ada79ba9b2a9e8304c4f6f6b816/pkg/solana/txm/txm.go#L569-L577))

silaslenihan: An in-between solution could be making the blockhash an optional parameter/override of the TXM's value? Though I'm not sure that makes that much sense since this value is refreshed constantly

pkg/solana/chainwriter/lookups.go
```diff
@@ -0,0 +1,271 @@
...
+	"fmt"
+	"reflect"
+
+	ag_binary "github.com/gagliardetto/binary"
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+)
+
```
Farber98, line 14: nit for readability 
```suggestion
// Lookup represents any mechanisms for resolving Solana accounts into `solana.AccountMeta`
```

Farber98, line 14:
```suggestion
// Lookup represents any mechanisms for resolving Solana accounts into `solana.AccountMeta`
```

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -0,0 +1,702 @@
...
+	staticLookupTablePubkey := chainwriter.GetRandomPubKey(t)
+	staticLookupKeys := chainwriter.CreateTestPubKeys(t, 2)
+	mockFetchLookupTableAddresses(t, rw, staticLookupTablePubkey, staticLookupKeys)
+
+	jsonFile, err := os.Open("testContractIDL.json")
+	require.NoError(t, err)
+
+	defer jsonFile.Close()
+
+	data, err := ioutil.ReadAll(jsonFile)
```
Farber98: nit: [ioutil](https://pkg.go.dev/io/ioutil) package is deprecated. We should prefer [io package](https://pkg.go.dev/io) instead. This will fix the linter issue

quoting from ioutil docs:
> Deprecated: As of Go 1.16, the same functionality is now provided by package [io](https://pkg.go.dev/io) or package [os](https://pkg.go.dev/os), and those implementations should be preferred in new code. See the specific function documentation for details.

quoting from [ioutil.readAll](https://pkg.go.dev/io/ioutil#ReadAll) docs:
> Deprecated: As of Go 1.16, this function simply calls [io.ReadAll](https://pkg.go.dev/io#ReadAll).

```suggestion
	data, err := io.ReadAll(jsonFile)
```

Farber98 (commented):
Logic looks fine to me. Leaving 2 nits

pkg/solana/chainwriter/chain_writer.go
Farber98: [this](https://github.com/smartcontractkit/chainlink-solana/pull/935#pullrequestreview-2520580600) answers our question about what to do 😅

pkg/solana/chainwriter/chain_writer.go
Farber98: I don’t think this issue blocks the current `ChainWriter` PR and it should be merged as-is with the existing blockhash call. 

If we decide to optimize later, I’d suggest going straight to a long term solution where the `TXM` manages blockhash fetching and assignment entirely, rather than an in-between approach if possible. 

In doing so, we’d refactor the callers so they no longer fetch the blockhash themselves and potentially call the `ChainWriter` rather than the TXM directly 💪 . In the case a caller really needs the blockhash, the `TXM` could simply overwrite it later. This would still ensure the correctness of the flow, avoiding the special-case complexity. The tradeoff is that we would be making a duplicate blockhash call for that particular case if needed.

pkg/solana/chainwriter/chain_writer.go
ilija42: > My understanding was that the chain-specific version would still be called ChainWriter and the generalized interface is the ContractWriter, but I'm not sure if that's correct nor do I have strong opinions on the semantics either way.

The Relayer interface should have ChainWriter and ChainReader, which consist of ContractWriter and ContractReader. Contract Reader/Writer are components that need to be configured and initialised, while the Chain Reader/Writer interfaces also contain functionalities that don't. Some examples are `LatestHead`, `GetChainStatus` for Reading and `Transact` for Writing. It would be nice to be consistent with naming across all Relayer implementations.

pkg/solana/chainwriter/lookups.go
silaslenihan: I went with something a little different 

`// Lookup is an interface that defines a method to resolve an address (or multiple addresses) from a given definition.
`

Farber98 (dismissed):

Farber98 (dismissed):

pkg/solana/chainwriter/ccip_example_config.go
```diff
@@ -0,0 +1,333 @@
...
+			// 	a. The user can configure any type of look up to get a list of lookupTables to read from.
+			// 	b. The ChainWriter reads from this lookup table and store the internal addresses in memory
+			//	c. Later, in the []Accounts the user can specify which accounts to include in the TX with an AccountsFromLookupTable lookup.
+			// 	d. Lastly, the lookup table is used to compress the size of the transaction.
+			DerivedLookupTables: []DerivedLookupTable{
+				{
+					Name: "RegistryTokenState",
+					// In this case, the user configured the lookup table accounts to use a PDALookup, which
+					// generates a list of one of more PDA accounts based on the input parameters. Specifically,
+					// there will be multple PDA accounts if there are multiple addresses in the message, otherwise,
```
amit-momin: nit:
```suggestion
					// there will be multiple PDA accounts if there are multiple addresses in the message, otherwise,
```

pkg/solana/chainwriter/chain_writer.go
amit-momin: Just to add my thoughts here too, I think it's ok to keep this as is. I was originally thinking the same thing that this would be the TXM's responsibility but like Juan mentioned there's places that need it before broadcast. We could refactor but we handle staleness in the TXM anyways so setting it here isn't much of a risk. Also the Solana TXM is built to spam the network unlike EVM to account for harsh network conditions. So transactions aren't stuck in the queue for as long and don't have much of a risk to go stale. Most are passed straight through if they pass validations. Think we'd only get backed up if the channel is full which is set at a 1000 capacity.

amit-momin (dismissed):

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -403,6 +409,7 @@ func TestChainWriter_SubmitTransaction(t *testing.T) {
 	seed2 := []byte("seed2")
 	programID := solana.MustPublicKeyFromBase58("6AfuXF6HapDUhQfE4nQG9C1SGtA1YjP3icaJyRfU4RyE")
 	derivedTablePda := mustFindPdaProgramAddress(t, [][]byte{seed2}, programID)
+	fmt.Println("pda:", derivedTablePda)
```
amit-momin: nit: think this got left behind

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,452 @@
...
+		if err := json.Unmarshal([]byte(programConfig.IDL), &idl); err != nil {
+			return fmt.Errorf("failed to unmarshal IDL for program: %s, error: %w", program, err)
+		}
+		for method, methodConfig := range programConfig.Methods {
+			idlDef, err := codec.FindDefinitionFromIDL(codec.ChainConfigTypeInstructionDef, methodConfig.ChainSpecificName, idl)
+			if err != nil {
+				return err
+			}
+
+			inputMod, err := methodConfig.InputModifications.ToModifier(codec.DecoderHooks...)
```
amit-momin, line 98: Noticed you made some changes to this method and we're not checking if `methodConfig.InputModifications` is nil or not anymore. Was there a change to the interface or validation that we don't need to do that here anymore?

silaslenihan: That's a good catch, I didn't notice it before because the tests run fine with `nil` `InputModifications`. I believe the reason for this is that a `nil` slice is Go can still have operations run on it, so `.ToModifier` is called without a nil pointer exception and in the context of the codec it is completely valid to have no modifiers

pkg/solana/utils/utils.go
```diff
@@ -0,0 +1,220 @@
...
+		require.NotNil(t, statusRes.Value)
+
+		unconfirmedTxCount := 0
+		for _, res := range statusRes.Value {
+			if res == nil || res.ConfirmationStatus == rpc.ConfirmationStatusProcessed || res.ConfirmationStatus == rpc.ConfirmationStatusConfirmed {
+				unconfirmedTxCount++
+			}
+		}
+		remaining = unconfirmedTxCount
+		fmt.Printf("Waiting for finalized funding on %d addresses\n", remaining)
```
amit-momin: nit: Missed this one earlier too

prashantkumar1982 (approved):

amit-momin (approved):

aalu1418 (approved):

PR #936, opened 2024-11-20
Store transaction errors caught before broadcast

amit-momin:
[NONEVM-957](https://smartcontract-it.atlassian.net/browse/NONEVM-957)
- Introduced a new `OnPrebroadcastError` method to the in-memory storage to directly put transactions into the errored state. Previously a transaction had to transition from another state to errored.
- Separated error parsing and status transition methods so errors can be more easily categorized and handled in both simulation or confirmation
- Introduced a smaller struct for storing finalized or errored statuses to minimize memory usage in cases of long retention timeouts
- Introduced a new `FatallyErrored` status to classify specific Solana errors that should not be retried by plugin side code. Changes to parse and classify errors with this status will be introduced in a later update.

[NONEVM-957]: https://smartcontract-it.atlassian.net/browse/NONEVM-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/txm/pendingtx.go
```diff
@@ -371,17 +377,60 @@ func (c *pendingTxContext) OnFinalized(sig solana.Signature, retentionTimeout ti
...
+		if broadcastedExists || confirmedExists {
+			return ErrIDAlreadyExists
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// upgrade to write lock if sig exists
```
Farber98, line 411:
```suggestion
	// upgrade to write lock if id does not exists
```

pkg/solana/txm/pendingtx.go
```diff
@@ -554,44 +619,45 @@ func (c *pendingTxContextWithProm) OnFinalized(sig solana.Signature, retentionTi
...
-		promSolTxmErrorTxs.WithLabelValues(c.chainID).Add(1)
+		promSolTxmSimRevertTxs.WithLabelValues(chainID).Add(1)
 	case TxFailSimOther:
-		promSolTxmSimOtherTxs.WithLabelValues(c.chainID).Add(1)
-		promSolTxmErrorTxs.WithLabelValues(c.chainID).Add(1)
+		promSolTxmSimOtherTxs.WithLabelValues(chainID).Add(1)
 	}
-
-	return id, err
+	promSolTxmErrorTxs.WithLabelValues(chainID).Add(1)
```
Farber98:
```suggestion
func incrementErrorMetrics(errType TxErrType, chainID string) {
	switch errType {
	case NoFailure:
		// Return early if no failure identified
		return
	case TxFailReject:
		promSolTxmRejectTxs.WithLabelValues(chainID).Inc()
	case TxFailRevert:
		promSolTxmRevertTxs.WithLabelValues(chainID).Inc()
	case TxFailDrop:
		promSolTxmDropTxs.WithLabelValues(chainID).Inc()
	case TxFailSimRevert:
		promSolTxmSimRevertTxs.WithLabelValues(chainID).Inc()
	case TxFailSimOther:
		promSolTxmSimOtherTxs.WithLabelValues(chainID).Inc()
	}
	promSolTxmErrorTxs.WithLabelValues(chainID).Inc()
```

Farber98: nit: There's an Inc() function that increments by 1 with less overhead under the hood

pkg/solana/txm/pendingtx.go
Farber98:
```suggestion
	// upgrade to write lock if id does not exists
```

pkg/solana/txm/pendingtx.go
```diff
@@ -371,17 +377,60 @@ func (c *pendingTxContext) OnFinalized(sig solana.Signature, retentionTimeout ti
...
+}
+
+func (c *pendingTxContext) OnPrebroadcastError(id string, retentionTimeout time.Duration, txState TxState, _ TxErrType) error {
+	// nothing to do if retention timeout is 0 since transaction is not stored yet.
+	if retentionTimeout == 0 {
+		return nil
+	}
+	err := c.withReadLock(func() error {
+		if tx, exists := c.finalizedErroredTxs[id]; exists && tx.state == txState {
+			return ErrAlreadyInExpectedState
```
Farber98, line 398: Better to return an error than idempotency here?

pkg/solana/txm/pendingtx_test.go
```diff
@@ -473,9 +473,6 @@ func TestPendingTxContext_on_finalized(t *testing.T) {
 		// Check it exists in finalized map
 		tx, exists := txs.finalizedErroredTxs[msg.id]
 		require.True(t, exists)
-		require.Len(t, tx.signatures, 2)
```
Farber98, line 476: removing because these checks were useless?

pkg/solana/txm/txm.go
```diff
@@ -417,24 +417,26 @@ func (txm *Txm) confirm() {
...
 					}
 
 					// if signature has an error, end polling
 					if res[i].Err != nil {
-						id, err := txm.txs.OnError(s[i], txm.cfg.TxRetentionTimeout(), TxFailRevert)
-						if err != nil {
-							txm.lggr.Infow("failed to mark transaction as errored", "id", id, "signature", s[i], "error", err)
-						} else {
-							txm.lggr.Debugw("tx state: failed", "id", id, "signature", s[i], "error", res[i].Err, "status", res[i].ConfirmationStatus)
+						if txState, errType := txm.processError(s[i], res[i].Err, false); errType != NoFailure {
```
Farber98, line 435: in which case we will have an err of NoFailure type? what does it means?

Farber98: nvm, I found the explanation in processError func

Farber98 (dismissed):
LGTM. Nice job with the optimizations while developing this. Left some comments but nothing major

pkg/solana/txm/pendingtx.go
```diff
@@ -36,20 +36,27 @@ type PendingTxContext interface {
...
-	signatures  []solana.Signature
+	tx         solana.Transaction
+	cfg        TxConfig
+	signatures []solana.Signature
+	id         string
+	createTs   time.Time
+	state      TxState
+}
+
+type finishedTx struct {
```
Farber98, line 60: may be useful to add a comment here mentioning this is used for errored and finalized txes

pkg/solana/txm/pendingtx.go
```diff
@@ -36,20 +36,27 @@ type PendingTxContext interface {
 	OnConfirmed(sig solana.Signature) (string, error)
 	// OnFinalized marks transaction as Finalized, moves it from the broadcasted or confirmed map to finalized map, removes signatures from signature map to stop confirmation checks
 	OnFinalized(sig solana.Signature, retentionTimeout time.Duration) (string, error)
+	// OnPrebroadcastError adds transaction that have not yet been broadcasted to the finalized/errored map as errored, matches err type using enum
```
```suggestion
	// OnPrebroadcastError adds transaction that has not yet been broadcasted to the finalized/errored map as errored, matches err type using enum
```

pkg/solana/txm/txm.go
silaslenihan: maybe useful to add a comment here anyway

pkg/solana/txm/utils.go
```diff
@@ -26,6 +26,7 @@ const (
 	Processed
 	Confirmed
 	Finalized
+	FatallyErrored
```
silaslenihan, line 30: nit: Add comment for fatally errored above like other states

pkg/solana/txm/txm.go
```diff
@@ -718,41 +735,58 @@ func (txm *Txm) simulateTx(ctx context.Context, tx *solanaGo.Transaction) (res *
...
-			_, err := txm.txs.OnError(sig, txm.cfg.TxRetentionTimeout(), TxFailSimOther) // cancel retry
-			if err != nil {
-				logValues = append(logValues, "stateTransitionErr", err)
+			// if simulating, return TxFailSimOther if error unknown
+			if simulation {
+				errType = TxFailSimOther
 			}
-			txm.lggr.Errorw("simulate: unrecognized error", logValues...)
+			txm.lggr.Errorw("unrecognized error", logValues...)
+			return Errored, errType
```
silaslenihan, line 796: should this be fatally errored if it's simulation type?

silaslenihan (dismissed):

pkg/solana/txm/pendingtx.go
amit-momin: I think returning nil here and below where we have the same check wouldn't be idempotent. If we do, we would continue into the write portion and overwrite the existing finishedTx with a renewed retention time. Maybe you could make the argument that we should renew it because a transaction was retried with the same id. But if there's a stored status already, the tx probably shouldn't have been retried. So think it's better to avoid renewing the time in case there's some erroneous retry logic that keeps these errored transactions alive indefinitely. Although this does help me realize I should separate the id exists check and the txState check here.

pkg/solana/txm/pendingtx_test.go
amit-momin: Sort of. These fields weren't important to finalized/errored transactions so the checks were probably useless. But also removing because of the change to only store `finishedTx` now which just has id, retentionTs, and state.

pkg/solana/txm/txm.go
amit-momin: Good point, I'll check with CCIP to see what they think. I was sort of following the EVM TXM philosophy of keeping unknown errors retryable but surfacing it through logs. But we can tackle this with the later update for fatal errors

Farber98 (dismissed):

silaslenihan (dismissed):

Farber98 (dismissed):

silaslenihan (approved):

Farber98 (approved):

aalu1418 (approved):

PR #940, opened 2024-11-21
fix: upgrade test fallback when not mapped

aalu1418:
### Description

quick fix to fallback to keypair if mapping doesn't exist


reductionista:
Did you mean to raise this against @EasterTheBunny 's branch rather than `develop`?

aalu1418:
> Did you mean to raise this against @EasterTheBunny 's branch rather than `develop`?

yes i will switch it back - but the upgrade test only runs when it detects changes to the `contracts` folder, so i need his changes to have CI run :)

aalu1418:
fixed!
image

reductionista (approved):

PR #944, opened 2024-11-26
[TT-1806] add required chain.link labels to k8s soak test

Tofel:
**Why?**
To better track k8s expenses


Farber98 (approved):

integration-tests/common/common.go
```diff
@@ -384,10 +384,24 @@ func BuildNodeContractPairID(node *client.ChainlinkClient, ocr2Addr string) (str
 }
 
 func (c *Common) Default(t *testing.T, namespacePrefix string) (*Common, error) {
+	productName := "data-feedsv2.0"
```
krebernisak, line 387: Is this the right product name?

krebernisak (approved):

integration-tests/common/common.go
```diff
@@ -384,10 +384,24 @@ func BuildNodeContractPairID(node *client.ChainlinkClient, ocr2Addr string) (str
...
+	nsLabels, err := environment.GetRequiredChainLinkNamespaceLabels(productName, "soak")
+	if err != nil {
+		return nil, err
+	}
+
+	workloadPodLabels, err := environment.GetRequiredChainLinkWorkloadAndPodLabels(productName, "soak")
```
aalu1418, line 393: do we need to differentiate between `soak` vs `smoke`?

aalu1418 (approved):

integration-tests/common/common.go
Tofel: it's OCRv2 test => `"data-feedsv2.0"`, it doesn't matter whether it's running on Solana or Ethereum or some other chain.

integration-tests/common/common.go
Tofel: good callout, we actually should

Tofel: I think in this case these tests are always `soak`, no?

chainchad (approved):

PR #945, opened 2024-11-26
ContractReader Interface Tests

EasterTheBunny:
Create interface tests for basic ContractReader interface.


contracts/programs/contract-reader-interface/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+use anchor_lang::prelude::*;
+use std::mem::size_of;
+
+declare_id!("6AfuXF6HapDUhQfE4nQG9C1SGtA1YjP3icaJyRfU4RyE");
+
+#[program]
+pub mod contract_reader_interface {
```
silaslenihan, line 7: nit: Should we call this `chain_components_...` since this contract will presumably also be used for ChainWriter?

pkg/solana/chainreader/client_wrapper.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+)
+
+type RPCClientWrapper struct {
+	*rpc.Client
+}
+
+func (w *RPCClientWrapper) GetMultipleAccountData(ctx context.Context, keys ...solana.PublicKey) ([][]byte, error) {
```
ilija42, line 14: I don't see this being used, what test would need this?

ilija42, line 17: I don't see this being used, is the idea to use it for batching?

EasterTheBunny: The `RPCClientWrapper` is to satisfy the contract reader service expected interface. I had to use an interface here instead of a client directly because there was something in the response object that isn't directly exported from the solana rpc client, making it difficult to write unit tests for.

contracts/programs/contract-reader-interface/src/lib.rs
EasterTheBunny: Do you mind if we make that change when we add `ChainWriter` in a followup PR?

silaslenihan: sure

contracts/programs/contract-reader-interface/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "contract-reader-interface"
+version = "0.1.0"
+description = "Created with Anchor"
+edition = "2021"
```
silaslenihan, line 5:
```suggestion
edition = "2024"
```

pkg/solana/chainreader/client_wrapper.go
silaslenihan: can you add a comment here that explains that?

silaslenihan (dismissed):

ilija42 (dismissed):

.github/workflows/relay.yml
```diff
@@ -63,3 +63,90 @@ jobs:
...
+      id: cache-image
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: contracts/docker-build.tar
+        key: ${{ runner.os }}-docker-pnpm-build-${{ needs.get_projectserum_version.outputs.projectserum_version }}-${{ hashFiles('**/Cargo.lock') }}
+    - name: build & save image
+      if: steps.cache-image.outputs.cache-hit != 'true'
+      run: |
+        docker buildx build . -t chainlink-solana:build --build-arg ANCHOR_CLI=${{ needs.get_projectserum_version.outputs.projectserum_version }}
+        docker save chainlink-solana > docker-build.tar
```
aalu1418: this is a very expensive CI action - why are we duplicating it here? maybe it's better to run this test with the other contract tests?

contracts/Anchor.toml
aalu1418, line 1: please keep changes to this file to a minimum - this causes merge conflicts down stream in the private repo

contracts/programs/contract-reader-interface/src/lib.rs
aalu1418, line 1: how would you feel about just creating a single program (combining with the event program) rather than multiple programs?

aalu1418 (changes requested):

.github/workflows/relay.yml
EasterTheBunny: I tried pulling from the cache without building the image, but that didn't work since the cache wasn't in the workflow. I can move this test to the contract tests for now to reduce build time and resource usage.

contracts/Anchor.toml
EasterTheBunny: These changes were done automatically by Anchor. To keep the file the same, I have to modify it after running `anchor build`. I'll do that though.

contracts/programs/contract-reader-interface/src/lib.rs
EasterTheBunny: That will be the followup. Silas mentioned wanting to use this program as well for contract writer tests.

aalu1418 (commented):
looks like some of the e2e test work didn't get propagated correctly

can you add 
```
          CHAINLINK_USER_TEAM: ${{ github.event.inputs.team || 'BIX' }}
```
to here: https://github.com/smartcontractkit/chainlink-solana/blob/6fd1891d0fbc79b0f10be353f6bd97446588209a/.github/workflows/e2e_custom_cl.yml#L425

it should look like this:
https://github.com/smartcontractkit/chainlink-solana/blob/6fd1891d0fbc79b0f10be353f6bd97446588209a/.github/workflows/e2e_custom_cl.yml#L342C1-L343C1

aalu1418 (dismissed):

silaslenihan (dismissed):

.github/workflows/rust.yml
```diff
@@ -123,3 +123,57 @@ jobs:
...
+        with:
+          go-version-file: "./integration-tests/go.mod"
+          check-latest: true
+          cache-dependency-path: "./integration-tests/go.sum"
+
+      - name: Install gotestloghelper
+        run: go install github.com/smartcontractkit/chainlink-testing-framework/tools/gotestloghelper@latest
+
+      - name: Install Solana CLI
+        run: ../scripts/install-solana-ci.sh
```
chainchad, line 164:
```suggestion
        run: ./scripts/install-solana-ci.sh
```

Workflow steps by default run from the root.

chainchad (changes requested):

aalu1418 (approved):

chainchad (approved):

PR #946, opened 2024-11-26
[NONEVM-706][Solana] - Refactor TXM + Rebroadcast Expired Tx functionality

Farber98:
### Description
Refactor `confirm` and `sendWithRetry` funcs

Adding a new `TxExpirationRebroadcast` toggable routine:
- Inside `confirm` loop, after checking confirmations, we will determine if we need to rebroadcast expired transactions that where not processed yet. Rebroadcasting a transaction involves:
	- Removing previous tx and creating a new one with updated blockhash 
		- **Consideration:** txid will be maintained, in case it was set by caller
	- calling `sendWithRetry` directly without enqueuing
		- If error, we should use [prebroadcast error functionality](https://github.com/smartcontractkit/chainlink-solana/pull/936). Client will need to retry the tx if this happens
	- **Consideration:** Implicitly it may be "bumping" despite config being off. We may get a new higher price for the rebroadcasted tx when we build and get a new compute unit price.

https://smartcontract-it.atlassian.net/browse/NONEVM-706

prev branch: https://github.com/smartcontractkit/chainlink-solana/pull/928

### Merge together with: 
- https://github.com/smartcontractkit/chainlink/pull/15437

### Soak Tests: 
- https://github.com/smartcontractkit/chainlink-solana/pull/962


pkg/solana/txm/txm_load_test.go
```diff
@@ -1,6 +1,6 @@
 //go:build integration
 
-package txm_test
+package txm
```
amit-momin: Why did we change this package?

pkg/solana/txm/pendingtx.go
```diff
@@ -44,22 +47,27 @@ type PendingTxContext interface {
...
+	lastValidBlockHeight uint64 // to track expiration
 }
 
 // finishedTx is used to store minimal info specifically for finalized or errored transactions for external status checks
 type finishedTx struct {
-	retentionTs time.Time
-	state       TxState
+	retentionTs      time.Time
+	state            TxState
+	rebroadcastCount int
```
amit-momin: Do we need to store this for finished txs?

pkg/solana/txm/txm_load_test.go
Farber98: all the other test files I was working on where under `txm` package and not under `txm_test`. That's why I renamed it. Is there any advantage/reason to have it as `txm_test`?

pkg/solana/txm/pendingtx.go
Farber98: Right now, only using it to check that [an `erroredTx` has been rebroadcasted when `sendWithRetry` returns an error](https://github.com/smartcontractkit/chainlink-solana/blob/56a64dad5f035dc483e52a724cf563b85560bc3f/pkg/solana/txm/txm_internal_test.go#L1630). Maybe it's wasteful and we can simply verify that the transaction had an error. What do you think?

pkg/solana/txm/txm_load_test.go
amit-momin: Just curious. Usually the best practice is to use a `_test` package for tests. Think in scenarios we have to test/access private methods or fields, we'll keep it under the same package as the implementation. I guess these tests were probably the only ones that only needed the public methods.

pkg/solana/txm/pendingtx.go
amit-momin: In that case, I think we can remove it since it's only really being used in tests right now

amit-momin: This probably doesn't matter too much but I'm just being a stickler for the memory usage

Farber98: Great! Got a workaround to test this without adding the extra field. We can track the number of calls to `LatestBlockhash`. We know that every time we want to rebroadcast, we need a new blockhash 😄

pkg/solana/txm/txm_load_test.go
Farber98: fixed

docs/relay/README.md
```diff
@@ -43,7 +43,8 @@ chainlink nodes solana create --name=<node-name> --chain-id=<chain-id> --url=<ur
...
+| `TxConfirmTimeout`    | duration when confirming a tx signature before signature is discarded as unconfirmed                                                                                                                               | 30s         |  
+| `TxExpirationRebroadcast`       | enables or disables transaction rebroadcast if expired. Expiration check is performed every `ConfirmPollPeriod`. A transaction is considered expired if the blockhash it was sent with is 150 blocks older than the latest blockhash.                                                                                                                                                                 | `false`      | `true`, `false`                                      |
 | `SkipPreflight`       | enable or disable preflight checks when sending tx                                                                                                                                                                 | `true`      | `true`, `false`                       |
```
amit-momin: Just wanted to call this out for consistency. I think the `enables or disables` is probably the more standard way to describe configs in our code base but think we should also update `SkipPreflight`'s description here if we're updating that here.
```suggestion
| `TxExpirationRebroadcast`       | enables or disables transaction rebroadcast if expired. Expiration check is performed every `ConfirmPollPeriod`. A transaction is considered expired if the blockhash it was sent with is 150 blocks older than the latest blockhash.                                                                                                                                                                 | `false`      | `true`, `false`                                      |
| `SkipPreflight`       | enables or disables preflight checks when sending tx                                                                                                                                                                 | `true`      | `true`, `false`                       |
```

pkg/solana/txm/pendingtx_test.go
amit-momin, line 1: I know the general behavior is tested in txm_internal but could you also add unit tests here for `ListAllExpiredBroadcastedTxs`?

pkg/solana/txm/txm.go
```diff
@@ -377,139 +399,209 @@ func (txm *Txm) confirm() {
...
+	}
+	// Rebroadcast all expired txes
+	for _, tx := range txm.txs.ListAllExpiredBroadcastedTxs(currHeight) {
+		txm.lggr.Debugw("transaction expired, rebroadcasting", "id", tx.id, "signature", tx.signatures)
+		if len(tx.signatures) == 0 { // prevent panic, shouldn't happen.
+			txm.lggr.Errorw("no signatures found for expired transaction", "id", tx.id)
+			continue
+		}
+		// only picking signature[0]. Remove func removes all related remaining signatures and cancels tx context.
+		_, err := txm.txs.Remove(tx.signatures[0])
```
amit-momin: I think this is the only place that `Remove` is being used (other than tests). If this feels a roundabout way to delete txs, we could change the method to accept a tx ID instead.

pkg/solana/txm/txm.go
```diff
@@ -377,139 +399,209 @@ func (txm *Txm) confirm() {
...
+		txm.lggr.Errorw("failed to get current block height", "error", err)
+		return
+	}
+	// Rebroadcast all expired txes
+	for _, tx := range txm.txs.ListAllExpiredBroadcastedTxs(*currBlockHeight.BlockHeight) {
+		txm.lggr.Debugw("transaction expired, rebroadcasting", "id", tx.id, "signature", tx.signatures, "lastValidBlockHeight", tx.lastValidBlockHeight, "currentBlockHeight", *currBlockHeight.BlockHeight)
+		if len(tx.signatures) == 0 { // prevent panic, shouldn't happen.
+			txm.lggr.Errorw("no signatures found for expired transaction", "id", tx.id)
+			continue
+		}
```
amit-momin: Do we still need this check now that `Remove` takes tx ID?

pkg/solana/txm/txm.go
```diff
@@ -377,139 +399,209 @@ func (txm *Txm) confirm() {
...
+		_, err := txm.txs.Remove(tx.id)
+		if err != nil {
+			txm.lggr.Errorw("failed to remove expired transaction", "id", tx.id, "error", err)
+			continue
+		}
+		rebroadcastTx := pendingTx{
+			tx:  tx.tx,
+			cfg: tx.cfg,
+			id:  tx.id, // using same id in case it was set by caller and we need to maintain it.
+		}
```
amit-momin: Just realized since we're just passing in the old tx cfg here, we aren't updating the compute unit price. Think we probably need to update `BaseComputeUnitPrice` in `tx.cfg` before setting it here.
```suggestion
		tx.cfg.BaseComputeUnitPrice = txm.fee.BaseComputeUnitPrice()
		rebroadcastTx := pendingTx{
			tx:  tx.tx,
			cfg: tx.cfg,
			id:  tx.id, // using same id in case it was set by caller and we need to maintain it.
		}
```

pkg/solana/txm/txm.go
```diff
@@ -630,7 +722,13 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
 		tx:  *tx,
 		cfg: cfg,
-		id:  id,
+	}
+
+	// If ID was not set by caller, create one.
+	if txID != nil && *txID != "" {
+		msg.id = *txID
+	} else {
+		msg.id = uuid.New().String()
```
amit-momin: I think we can remove this code. Looks like the old code we had for this (below) got pulled back in with prebroadcast error changes. If anything we can maybe just update it to use the dedicated `NewString()` method. But I think we probably need to keep the original code. We want to generate the ID before estimating compute unit limits since it's needed to set prebroadcast errors but we only want to create the pending tx after the compute unit limit is estimated and set in the cfg.
```
id := uuid.NewString()
if txID != nil && *txID != "" {
	id = *txID
}
```

pkg/solana/txm/txm_internal_test.go
```diff
@@ -767,6 +772,7 @@ func TestTxm_disabled_confirm_timeout_with_retention(t *testing.T) {
 	cfg.Chain.TxRetentionTimeout = relayconfig.MustNewDuration(5 * time.Second)
 	mc := mocks.NewReaderWriter(t)
 	mc.On("GetLatestBlock", mock.Anything).Return(&rpc.GetBlockResult{}, nil).Maybe()
+	mc.On("SlotHeight", mock.Anything).Return(uint64(0), nil).Maybe()
```
amit-momin: Do we still need these `SlotHeight` mocks if we've switched to `GetLatestBlock`?

pkg/solana/txm/txm.go
```diff
@@ -377,139 +399,209 @@ func (txm *Txm) confirm() {
...
+	} else {
+		txm.lggr.Debugw("marking transaction as finalized", "id", id, "signature", sig)
+	}
+}
+
+// rebroadcastExpiredTxs attempts to rebroadcast all transactions that are in broadcasted state and have expired.
+// An expired tx is one where it's blockhash lastValidBlockHeight is smaller than the current slot height.
+// If any error occurs during rebroadcast attempt, they are discarded, and the function continues with the next transaction.
+func (txm *Txm) rebroadcastExpiredTxs(ctx context.Context, client client.ReaderWriter) {
+	currBlockHeight, err := client.GetLatestBlock(ctx)
```
amit-momin: nit: this is probably a misnomer for this variable. Think we should call this something like `currBlock` or `latestBlock` since that's what's really being returned by the RPC call

pkg/solana/txm/pendingtx_test.go
```diff
@@ -1137,3 +1144,157 @@ func randomSignature(t *testing.T) solana.Signature {
 
 	return solana.SignatureFromBytes(sig)
 }
+
+func TestPendingTxContext_ListAllExpiredBroadcastedTxs(t *testing.T) {
```
amit-momin, line 1176: Nice!

pkg/solana/txm/pendingtx.go
```diff
@@ -211,12 +206,32 @@ func (c *pendingTxContext) Remove(sig solana.Signature) (id string, err error) {
...
 }
 
-func (c *pendingTxContext) ListAll() []solana.Signature {
+func (c *pendingTxContext) ListAllSigs() []solana.Signature {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 	return maps.Keys(c.sigToID)
 }
 
+func (c *pendingTxContext) ListAllTxsIDs() []string {
```
amit-momin: Noticed this is only really used in a test. Could we just use `maps.Values(c.sigToID)` directly to avoid adding a new method? No worries if there's a risk of race or anything.

pkg/solana/txm/txm.go
Farber98: good catch

pkg/solana/txm/txm.go
Farber98: Oh you are right, thought it was handled by buildTx [here](https://github.com/smartcontractkit/chainlink-solana/blob/0e38174d35b6d6ee9f03dba526bd2bf135ad9809/pkg/solana/txm/txm.go#L257-L264) but it's just using the old value

pkg/solana/txm/txm.go
Farber98: good catch x1000

pkg/solana/txm/pendingtx.go
Farber98: removed!

pkg/solana/txm/txm.go
Farber98: Thanks for this! Fixed

pkg/solana/txm/txm_internal_test.go
Farber98: Removed!

amit-momin: Noticed this `SlotHeight` mock is still here

amit-momin (dismissed):

pkg/solana/txm/pendingtx.go
```diff
@@ -48,12 +51,13 @@ type PendingTxContext interface {
...
-	id         string
-	createTs   time.Time
-	state      TxState
+	tx                   solana.Transaction
+	cfg                  TxConfig
+	signatures           []solana.Signature
+	id                   string
+	createTs             time.Time
+	state                TxState
+	lastValidBlockHeight uint64 // to track expiration
```
silaslenihan: Out of curiosity, what is an invalid block height? Like a reorged out chain?

pkg/solana/txm/pendingtx.go
```diff
@@ -211,12 +206,26 @@ func (c *pendingTxContext) Remove(sig solana.Signature) (id string, err error) {
...
+func (c *pendingTxContext) ListAllExpiredBroadcastedTxs(currBlockHeight uint64) []pendingTx {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	broadcastedTxes := make([]pendingTx, 0, len(c.broadcastedProcessedTxs)) // worst case, all of them
+	for _, tx := range c.broadcastedProcessedTxs {
+		if tx.state == Broadcasted && tx.lastValidBlockHeight < currBlockHeight {
+			broadcastedTxes = append(broadcastedTxes, tx)
+		}
+	}
+	return broadcastedTxes
```
```suggestion
	expiredBroadcastedTxs := make([]pendingTx, 0, len(c.broadcastedProcessedTxs)) // worst case, all of them
	for _, tx := range c.broadcastedProcessedTxs {
		if tx.state == Broadcasted && tx.lastValidBlockHeight < currBlockHeight {
			expiredBroadcastedTxs = append(expiredBroadcastedTxs, tx)
		}
	}
	return expiredBroadcastedTxs
```

silaslenihan (dismissed):

pkg/solana/txm/pendingtx.go
prashantkumar1982: +1. I assume this is related to using latest blockhash in tx, and tracking its expiration. But the variable name is confusing, if it tracks a particular block number, or just a depth.

pkg/solana/txm/pendingtx.go
```diff
@@ -68,9 +72,9 @@ type pendingTxContext struct {
 	cancelBy map[string]context.CancelFunc
 	sigToID  map[solana.Signature]string
 
-	broadcastedTxs      map[string]pendingTx  // transactions that require retry and bumping i.e broadcasted, processed
-	confirmedTxs        map[string]pendingTx  // transactions that require monitoring for re-org
-	finalizedErroredTxs map[string]finishedTx // finalized and errored transactions held onto for status
+	broadcastedProcessedTxs map[string]pendingTx  // broadcasted and processed transactions that may require retry and bumping
```
prashantkumar1982, line 75: In EVM, we just call these Unconfirmed.
Unconfirmed means tx was successfully broadcasted, but not yet mined into a block.
My preference would be to reuse that same name here too?

pkg/solana/txm/pendingtx.go
```diff
@@ -93,7 +97,7 @@ func (c *pendingTxContext) New(tx pendingTx, sig solana.Signature, cancel contex
 			return ErrSigAlreadyExists
 		}
 		// validate id does not exist
-		if _, exists := c.broadcastedTxs[tx.id]; exists {
+		if _, exists := c.broadcastedProcessedTxs[tx.id]; exists {
```
prashantkumar1982, line 100: What is this tx.id? Is it the one caller provided as an idempotency key?

If this is checking if this id exists, then we likely need to do better checks across many other status.

pkg/solana/txm/pendingtx.go
```diff
@@ -211,12 +206,26 @@ func (c *pendingTxContext) Remove(sig solana.Signature) (id string, err error) {
...
 }
 
-func (c *pendingTxContext) ListAll() []solana.Signature {
+func (c *pendingTxContext) ListAllSigs() []solana.Signature {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 	return maps.Keys(c.sigToID)
 }
 
+// ListAllExpiredBroadcastedTxs returns all the txes that are in broadcasted state and have expired for given block height compared against their lastValidBlockHeight.
```
prashantkumar1982: nit: could we use blockNumber instead of blockHeight in all these places?
Block number is a more unambiguous term.

pkg/solana/txm/txm.go
```diff
@@ -175,197 +179,215 @@ func (txm *Txm) run() {
...
-		var finalSig [64]byte
-		copy(finalSig[:], sigBytes)
-		newTx.Signatures = append(newTx.Signatures, finalSig)
-
-		return newTx, nil
+	if blockhash == nil || blockhash.Value == nil {
+		return solanaGo.Transaction{}, "", solanaGo.Signature{}, errors.New("nil pointer returned from LatestBlockhash")
 	}
+	msg.tx.Message.RecentBlockhash = blockhash.Value.Blockhash
+	msg.lastValidBlockHeight = blockhash.Value.LastValidBlockHeight
```
prashantkumar1982: nit: better name seems to me recentBlockhashNumber.
It indicates what block number's hash was used in this tx.

pkg/solana/txm/txm.go
```diff
@@ -175,197 +179,215 @@ func (txm *Txm) run() {
...
 
-	initTx, initBuildErr := buildTx(ctx, baseTx, 0)
-	if initBuildErr != nil {
-		return solanaGo.Transaction{}, "", solanaGo.Signature{}, initBuildErr
+	// Build and sign initial transaction setting compute unit price and limit
+	initTx, err := txm.buildTx(ctx, msg, 0)
```
prashantkumar1982, line 187: Do we know what happens to the previous tx(original one before the retry) which this is intending to replace?

Since this is updating the recentBlockHash, I would assume Solana treats this as a brand new Tx, not something which replaces the previous one. So maybe Solana confirms both of these Txs separately?
On EVM, we were sure only 1 would get confirmed, since they shared the same nonce. But here, they don't share anything.

pkg/solana/txm/txm.go
```diff
@@ -175,197 +179,215 @@ func (txm *Txm) run() {
 	}
 }
 
+// sendWithRetry attempts to send a transaction with exponential backoff retry logic.
+// It builds, signs and sends the initial tx with a new valid blockhash, and starts a retry routine with fee bumping if needed.
```
prashantkumar1982: I commented below too, but if you use a new valid blockHash, are you sure Solana would still figure out and treat this as a replacement of the older Tx?
If not, then we may have both of these txs get included.

pkg/solana/txm/pendingtx.go
Farber98: In Solana, each transaction includes a `"recent blockhash"` that expires when that blockhash is no longer `"recent enough"`

During transaction processing, Solana validators will check if the `"recent blockhash"` of each transaction is recorded within the 150 most recent stored hashes. If the transaction's `"recent blockhash"` is older than this maximum processing age, the transaction won't be processed.

To track the expiration of each transaction, we call the RPC method [GetLatestBlockhash](https://solana.com/docs/rpc/http/getlatestblockhash) and store the return value `lastValidBlockHeight` within our `pendingTx` object. This value represents the height of the last block at which the returned `blockhash` value will be valid. 

Each slot (period of time a validator can produce a block) fluctuates between 400 ms and 600 ms. This means transactions can only use a given `blockhash` for 150 blocks, which is between 60 to 90 seconds since creation. After this, it will be considered expired by the network and you'll need a new `blockhash` to make it go through

pkg/solana/txm/pendingtx.go
Farber98: Names except `Broadcasted` were chosen to be consistent with the [Solana Commitment Status](https://docs.anza.xyz/consensus/commitments/)

I feel it's good to reuse the same names across chains, but I also think `Processed` is clearer than `Unconfirmed` in this case. In my opinion, `Unconfirmed` does not make it clear whether the transaction was `Processed` or simply `Broadcasted`

- `Broadcasted`: Our TXM sent the tx
- `Processed`: tx was Processed by network and included in a block

@amit-momin @silaslenihan what do you think about `Broadcasted` vs `Unconfirmed`?

pkg/solana/txm/pendingtx.go
Farber98: Yes! If set by caller we use that one. If not set, we generate a random `UUID` [when enqueueing txes](https://github.com/smartcontractkit/chainlink-solana/blob/6bcdd8f23ff2b87aaa6122cf74ca4ceeb186e304/pkg/solana/txm/txm.go#L700-L704).

What checks do you think are missing?

pkg/solana/txm/txm.go
Farber98: We are not able to change `Message`, it is coming from https://github.com/gagliardetto/solana-go/blob/da2193071f56059aa35010a239cece016c4e827f/message.go#L89-L116

pkg/solana/txm/txm.go
Farber98: This previous response makes sense here: https://github.com/smartcontractkit/chainlink-solana/pull/946#discussion_r1877977390

We will proceed and Rebroadcast a tx when the first one has already expired and become invalid

pkg/solana/txm/txm.go
Farber98: Also quoting https://github.com/smartcontractkit/chainlink-solana/pull/946#discussion_r1877977390 here

It won't be a `"replacement"` for the old TX, but the old TX became invalid in the eyes of validators and won't be picked up after 150 blocks have passed

pkg/solana/txm/pendingtx.go
Farber98: I think it's also easier in terms of `block number` than `block height`. 

Guys, do you think it makes sense to rename `lastValidBlockHeight` to `lastValidBlockNumber` inside our TXM? The `lastValidBlockHeight` is the official one, but `lastValidBlockNumber` may be easier to deal with. 

Another alternative would be to put comments when we use the term `block height` to be explicit that we are referring to `block number`

CC: @amit-momin @silaslenihan

pkg/solana/txm/pendingtx.go
silaslenihan: Ah yeah this became clear to me later in the PR review, thanks for clarifying though that makes sense

pkg/solana/txm/pendingtx.go
amit-momin: I agree with Juan. `Unconfirmed` might be misleading in the context of Solana since we have this `Processed` state in between. In EVM, we could consider a broadcasted transaction as `Unconfirmed` since the next step in the lifecycle is `Confirmed`

pkg/solana/txm/pendingtx.go
amit-momin: I'm ok either way but I thought both of those terms referred to the same thing? What's the ambiguity with block height? Since I thought they were the same, I was on the side of using the official name from the [RPC](https://solana.com/docs/rpc/http/getlatestblockhash) response

pkg/solana/txm/pendingtx.go
prashantkumar1982: The previously sent Tx by the caller with same id, may be still enqueued in the DB, not yet broadcasted, or may be confirmed, or finalizedErroredTx, or just finalized right?
We can't assume that a previously used id will only be in this broadcastedProcessedTxs list.

pkg/solana/txm/pendingtx.go
prashantkumar1982: If Solana has these standard terms, then yes, better to align with that. I didn't know.

pkg/solana/txm/pendingtx.go
prashantkumar1982: Yes if Solana defines a specific meaning to some terms, lets use that.
Didn't know that.
In this case, lets then just add a comment saying what it means.
For example: lastValidBlockHeight is the highest block number till which this Tx is valid. After this, this Tx gets treated as expired.

pkg/solana/txm/txm.go
prashantkumar1982: Yes, saw it now. Thanks

pkg/solana/txm/txm.go
```diff
@@ -175,197 +179,215 @@ func (txm *Txm) run() {
...
+	newTx.Signatures = append(newTx.Signatures, finalSig)
+
+	return newTx, nil
+}
+
+// retryTx contains the logic for retrying the transaction, including exponential backoff and fee bumping.
+// Retries until context cancelled by timeout or called externally.
+// It uses handleRetry helper function to handle each retry attempt.
+func (txm *Txm) retryTx(ctx context.Context, msg pendingTx, currentTx solanaGo.Transaction, sigs *signatureList) {
+	deltaT := 1 // initial delay in ms
```
prashantkumar1982, line 273: Isn't 1 ms too aggressive?
I assume just a round-trip time to Solana RPC call would be in order of 10s of ms atleast.
So after initial send, we may want to wait atleast 10 ms to give the initial tx some time to get confirmed.

Otherwise if say we have initial deltaT as 1ms, then next exponential backoffs are at 1, 2 ,4 8, 16, 31, 64, 128 ms.

But I see that the confirm logic, which checks Tx confirmation, has a polling that's 500 ms:
ConfirmPollPeriod:   config.MustNewDuration(500 * time.Millisecond)

Seems like we will retry many times even before trying to check confirmation once.

amit-momin: This was logic that carried over from the existing code for Data Feeds. We chose to keep this the same so we don't introduce any regressions from DF. We could maybe make this configurable in the future especially if we need to reduce RPC calls but seems like this was out of necessity to make sure our transaction gets out in front of the current validator. The multinode integration probably helps a bit with that though so spamming this much maybe isn't needed anymore. I had the same thoughts as you initially but was reluctant to relax the existing code in case it introduced any unforeseen issues.

pkg/solana/txm/pendingtx.go
amit-momin: I think that's a fair point. We should check across all of the maps when adding a new transaction to our in-memory store

Farber98: I'll add this

pkg/solana/txm/txm.go
Farber98: We [decided here](https://chainlink-core.slack.com/archives/C04Q61M2MFT/p1733936968071019?thread_ts=1733426993.295679&cid=C04Q61M2MFT) to create a [ticket](https://smartcontract-it.atlassian.net/browse/NONEVM-1031) to make this retry configurable. We will keep the default values as they are for now. This will give us the flexibility to adjust them based on the use case. For example, we could relax these configurations for CCIP and make them more aggressive for DF

amit-momin (dismissed):

prashantkumar1982 (dismissed):

pkg/solana/txm/txm.go
```diff
@@ -377,139 +399,206 @@ func (txm *Txm) confirm() {
...
+		txm.lggr.Errorw("failed to mark transaction as finalized", "id", id, "signature", sig, "error", err)
+	} else {
+		txm.lggr.Debugw("marking transaction as finalized", "id", id, "signature", sig)
+	}
+}
+
+// rebroadcastExpiredTxs attempts to rebroadcast all transactions that are in broadcasted state and have expired.
+// An expired tx is one where it's blockhash lastValidBlockHeight (last valid block number) is smaller than the current block height (block number).
+// If any error occurs during rebroadcast attempt, they are discarded, and the function continues with the next transaction.
+func (txm *Txm) rebroadcastExpiredTxs(ctx context.Context, client client.ReaderWriter) {
```
aalu1418, line 556: this feels like there's a potential race condition here

* tx is broadcast to the network and is included just before blockhash expiration
* confirmer does not see it yet because it waits for `confirmed` commitment level
* TXM marks a base transaction as expired because the block threshold has passed 
* deletes the saved tx id from the in-memory storage, and rebroadcasts a tx with a new blockhash
* this could go on forever because it can never confirm that a successful tx is confirmed because the tx.id is always deleted before

pkg/solana/txm/txm_internal_test.go
```diff
@@ -1186,3 +1209,432 @@ func addSigAndLimitToTx(t *testing.T, keystore SimpleKeystore, pubkey solana.Pub
 	require.NoError(t, fees.SetComputeUnitLimit(&txCopy, limit))
 	return &txCopy
 }
+
+func TestTxm_ExpirationRebroadcast(t *testing.T) {
```
aalu1418, line 1210: this tests only runs against a mocked endpoints under ideal behavior - please add additional tests that run against a live validator

aalu1418 (changes requested):

pkg/solana/txm/txm.go
Farber98: Wouldn't this be handled gracefully by the existing code? 

The `processConfirmations` method is called before `rebroadcastExpiredTxs` updating each signature status. Here we would detect if we have [transitioned from Broadcasted to Processed](https://github.com/smartcontractkit/chainlink-solana/blob/b4b4fd54f8999f83f66f194bb48305499c511a8f/pkg/solana/txm/txm.go#L467-L470).

Then inside `rebroadcastExpiredTxs` we call `ListAllExpiredBroadcastedTxs`, [where we only consider the ones which have `Broadcasted` state](https://github.com/smartcontractkit/chainlink-solana/blob/b4b4fd54f8999f83f66f194bb48305499c511a8f/pkg/solana/txm/pendingtx.go#L230C1-L241C1)

amit-momin: I think we would avoid that scenario since we only call `rebroadcastExpiredTxs` after processing the latest confirmation status changes. Since this method only processes `Broadcasted` transaction, the confirmation logic would handle any last minute state changes before we assess transaction for expiry. So in your scenario, we would notice that the transaction advanced to `Processed`, mark it accordingly in our in-memory storage, then move on to the expiry check. The expiry check would notice there aren't any transactions in `Broadcasted` state that are expired and just move on.

pkg/solana/txm/txm_internal_test.go
Farber98: Hey! We've found that testing this scenario without mocking is challenging because we don't have a way to force a transaction to expire in a live validator.

The only solution we've identified involves modifying some methods to simulate this behavior for the tests. However, this approach is not ideal, as it would require adding additional flows solely to make the test pass and to trigger a rebroadcast in the live validator tests.

Do you have any ideas on how to achieve this that we might be missing?

pkg/solana/txm/txm.go
aalu1418: what happens if a transaction skips the `processed` phase? 
`processed` is just the view of a single RPC node - so if a transaction can go from `processed` to non-existent, then it might also be possible to go from non-existent to `confirmed`

it just feels odd to delete all previous attempts at a transaction and then recreate - rather than just updating the transaction with a new blockhash

pkg/solana/txm/txm_internal_test.go
aalu1418: you could potentially introduce some sort of proxy that takes the transaction and returns a transaction signature, but doesn't allow any more transactions through until it knows the first transaction blockhash has expired

pkg/solana/txm/txm.go
amit-momin: We would advance the state from non-existent (internally what we consider `Broadcasted`) to `Confirmed` and also avoid rebroadcasting with a new hash in that case. Our state transition methods don't require us to move one step up at a time in case of slow polling or other reasons. We can even jump straight to `Finalized`. We only delete all previous attempts/sigs if we can confirm that the tx has a non-existent state and its blockhash has expired. I was thinking it was better to clean up those old attempts/sigs than keep them around and waste cycles on status checks on them since they're expired anyways.

pkg/solana/txm/txm_internal_test.go
Farber98: Would it be okay to address this in a separate ticket under the [Testing](https://smartcontract-it.atlassian.net/browse/NONEVM-698) track? So we are able to get this one in before the freeze and finish the re-org work that depends on this PR to complete the implementation track

pkg/solana/txm/txm.go
aalu1418: ok - the chances of this race happening might be low

is there logging around a transaction signature that doesn't match any tx id? something that fires if an assumption is broken?

pkg/solana/txm/txm_internal_test.go
aalu1418: why does the freeze impact the timeline? afaik that's just for production environments, not development?

also this PR touches significant portions of the most sensitive part of a solana integration, test coverage as the changes are made imo is a must

pkg/solana/txm/txm.go
amit-momin: If one of the state transition methods gets a sig that doesn't match to an ID in the internal map, it should return a `ErrSigDoesNotExist` error. Depending where the method was called, we have some logging to surface an assumption violation. But not in all cases like the confirmation expiration where we can maybe assume a different process cleaned up the signatures.

pkg/solana/txm/txm_internal_test.go
aalu1418: just clarifying a few things:
* there is a production code freeze, but that shouldn't impact development here
* i'd like to see an attempt at a proper integration tests, but it can be time-boxed to a day or two (if it still doesn't work, then a follow-up ticket is fine)

Farber98: Thanks for clarifying, thought the freeze involved not merging new code in `chainlink repo`, even if it was development 😅. I'll start working on these integration tests

pkg/solana/txm/txm.go
```diff
@@ -568,27 +551,48 @@ func (txm *Txm) handleFinalizedSignatureStatus(sig solanaGo.Signature) {
...
+
+		blockhash, err := client.LatestBlockhash(ctx)
+		if err != nil {
+			txm.lggr.Errorw("failed to get latest blockhash for rebroadcast", "error", err)
+			return
+		}
+		if blockhash == nil || blockhash.Value == nil {
+			txm.lggr.Errorw("nil pointer returned from LatestBlockhash for rebroadcast")
+			return
+		}
```
amit-momin: Think it might be better to make this call outside of the for loop? We'll loop through these tx quick enough that I don't think we have to worry about refreshing the blockhash every time.

pkg/solana/txm/txm_integration_test.go
```diff
@@ -0,0 +1,186 @@
...
+			} else {
+				require.Empty(t, rebroadcastLogs, "Rebroadcast should not occur")
+				require.Empty(t, rebroadcastLogs2, "Rebroadcast should not occur")
+			}
+		})
+	}
+}
+
+func setup(t *testing.T, url string, txExpirationRebroadcast bool) (context.Context, *solanaClient.Client, *txm.Txm, solana.PublicKey, solana.PublicKey, *observer.ObservedLogs) {
+	ctx := context.Background()
```
amit-momin: nit: think we should use the test context here. Think it's something like `tests.Context(t)`

pkg/solana/txm/txm.go
Farber98: Sure it is. The duplication part of [this definition](https://solana.com/docs/core/transactions#recent-blockhash) made me think that blockhash should be unique but after researching a little it only affects two txes that are exactly equal within the same block (very unlikely)
> All transactions include a recent blockhash to act as a timestamp for the transaction. The blockhash is used to prevent duplications and eliminate stale transactions.

pkg/solana/txm/txm_integration_test.go
```diff
@@ -0,0 +1,186 @@
...
+				require.Equal(t, initSenderBalance, finalSenderBalance, "Sender balance should remain the same")
+				require.Equal(t, uint64(0), finalReceiverBalance, "Receiver should not receive any funds")
+			}
+
+			// Verify rebroadcast logs
+			rebroadcastLogs := observer.FilterMessageSnippet("rebroadcast transaction sent").All()
+			rebroadcastLogs2 := observer.FilterMessageSnippet("transaction expired, rebroadcasting").All()
+			if tc.expectRebroadcast {
+				require.NotEmpty(t, rebroadcastLogs, "Expected rebroadcast log message not found")
+				require.NotEmpty(t, rebroadcastLogs2, "Expected rebroadcast log message not found")
```
amit-momin: I think it might be worth also ensuring that we only rebroadcast once here. Could we maybe leverage `Len()` and check that it's explicitly 1 here?
```suggestion
			rebroadcastLogs := observer.FilterMessageSnippet("rebroadcast transaction sent").Len()
			rebroadcastLogs2 := observer.FilterMessageSnippet("transaction expired, rebroadcasting").Len()
			if tc.expectRebroadcast {
				require.Equal(t, 1, rebroadcastLogs, "Expected rebroadcast log message not found")
				require.Equal(t, 1, rebroadcastLogs2, "Expected rebroadcast log message not found")
```

pkg/solana/relay.go
```diff
@@ -24,7 +24,16 @@ import (
...
 
 type TxManager interface {
-	Enqueue(ctx context.Context, accountID string, tx *solana.Transaction, txID *string, txCfgs ...txm.SetTxConfig) error
+	// Enqueue adds a tx to the txm queue for processing and submitting to the Solana network.
+	// An error is returned if the txm is not ready, if the tx is invalid, or if the queue is full.
+	//
+	// Important Notes:
+	// - The tx must contain at least one account key. The first account will be used to sign the tx (fee payer's public key).
+	// - txCfgs can be used to set custom tx configurations.
+	// - If an txID is provided, it will be used to identify the tx. Otherwise, a random UUID will be generated.
```
amit-momin: nit:
```suggestion
	// - If a txID is provided, it will be used to identify the tx. Otherwise, a random UUID will be generated.
```

pkg/solana/relay.go
```diff
@@ -24,7 +24,16 @@ import (
...
+	// - The caller needs to set the tx.Message.RecentBlockhash and provide the corresponding lastValidBlockHeight.
+	//   These values are obtained from the GetLatestBlockhash RPC call.
```
amit-momin: Could you combine this with the line above since they go together? Think it's ok if the line gets long.
```suggestion
	// - The caller needs to set the tx.Message.RecentBlockhash and provide the corresponding lastValidBlockHeight. These values are obtained from the GetLatestBlockhash RPC call.
```

pkg/solana/txm/txm_integration_test.go
Farber98: this is good!

pkg/solana/relay.go
Farber98: 🇪🇸 😢

amit-momin (approved):

aalu1418 (approved):

PR #950, opened 2024-11-27
LogPoller Boilerplate

EasterTheBunny:
basic boilerplate that connects loader to base log poller


pkg/solana/logpoller/poller.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+type EventSaver interface {
+	SaveEvent(evt ProgramEvent) error
```
reductionista, line 16: This will need to accept a slice of events, more like this:
```suggestion
	SaveEvents(events []ProgramEvent) error
```
But we already have an ORM method that does basically that, so we can probably just use that unless we want to wrap it with some conversion layer:
```
     InsertLogs(logs []Log) error
```

pkg/solana/logpoller/poller.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+		Start: p.start,
+		Close: p.close,
+	}.NewServiceEngine(lggr)
+	p.lggr = p.engine.SugaredLogger
+
+	return p
+}
+
+func (p *Service) AddFilter(name string) error {
```
reductionista, line 57: This looks like `RegisterFilter()` we have that implemented and is in review.

reductionista: The interface for LogPoller (as well as all of the data structures that will be passed between QueryKey & LogPoller) is fully specified in the design doc.  In principle, that should be enough for QueryKey to be implemented, and should be considered the source of truth--so I'd recommend starting there, and if there is anything missing or that needs to be updated in light of more recent conversations we can discuss.

pkg/solana/logpoller/poller.go
EasterTheBunny: That works. This method is only a placeholder and can use the ORM method instead.

PR #951, opened 2024-11-28
[NONEVM-984][solana] - Reorg Detection + lighter rpc call

Farber98:
### Description
- Track transaction statuses on per signature basis to identify which signature for a transaction was included to detect re-orgs specifically for it
- Update the confirmation logic to identify regression in a signature’s transaction status when no or processed status is received to detect re-orgs
	- A transaction can revert back to no status or processed from confirmed status
	- A transaction can revert back to no status from processed status
- If a re-org is detected 
	- from confirmed, restart the retry/bump loop
	- from processed, we don't do nothing and it's handled by expiration rebroadcast if needed later

### Tickets
- https://smartcontract-it.atlassian.net/browse/NONEVM-984
- https://smartcontract-it.atlassian.net/browse/NONEVM-1015

### Soak Testing
- https://github.com/smartcontractkit/chainlink-solana/pull/969


pkg/solana/txm/pendingtx.go
```diff
@@ -66,11 +70,21 @@ type finishedTx struct {
 	state       TxState
 }
 
+type txInfo struct {
+	id    string
+	state TxState
+}
+
+type retryCtx struct {
```
Farber98: Heads up here. Changed `cancelBy` to `sigToRetryCtx`. The idea is to be able to retrieve the `retryCtx` if a reorg was detected and only restart the retry/bumping cycle if the original `retryCtx` is still alive

pkg/solana/txm/pendingtx.go
```diff
@@ -66,11 +70,21 @@ type finishedTx struct {
 	state       TxState
 }
 
+type txInfo struct {
```
Farber98, line 81: removed state from `pendingTx` and moved it to sig level, as we now want to track state per sig

pkg/solana/txm/pendingtx.go
```diff
@@ -81,21 +95,18 @@ type pendingTxContext struct {
...
 
 		broadcastedProcessedTxs: map[string]pendingTx{},
 		confirmedTxs:            map[string]pendingTx{},
 		finalizedErroredTxs:     map[string]finishedTx{},
 	}
 }
 
-func (c *pendingTxContext) New(tx pendingTx, sig solana.Signature, cancel context.CancelFunc) error {
+// New adds a new tranasction in Broadcasted state to the storage
+func (c *pendingTxContext) New(tx pendingTx) error {
```
Farber98, line 112: Previously we were adding the first signature inside `New` and associating cancel func. Removed that part so `New` is only in charge of creating the new pendingTx and `AddSignature` adds the signature and associates the cancel func. This is when a reorg regressing from confirmed state is detected and we need to rebroadcast the tx restarting a new retry/bumping cycle

pkg/solana/txm/pendingtx.go
```diff
@@ -187,52 +194,59 @@ func (c *pendingTxContext) Remove(sig solana.Signature) (id string, err error) {
...
-			delete(c.cancelBy, id)
+			delete(c.confirmedTxs, info.id)
 		}
 
-		// remove all signatures associated with transaction from sig map
+		// remove all signatures associated with transaction from sig map and cancel any associated contexts to stop retry/bumping cycles
 		for _, s := range tx.signatures {
-			delete(c.sigToID, s)
+			delete(c.sigToTxInfo, s)
+			if rtryCtx, exists := c.sigToRetryCtx[s]; exists {
```
Farber98: `rtryCtx` because linter was angry with `retryCtx` shadowing the struct def

pkg/solana/txm/txm_internal_test.go
```diff
@@ -1329,23 +1329,21 @@ func TestTxm_ExpirationRebroadcast(t *testing.T) {
 			// First transaction should be rebroadcasted.
 			if time.Since(nowTs) < cfg.TxConfirmTimeout()-2*time.Second {
 				return nil
-			} else {
```
Farber98: linter angry with `else` cases

pkg/solana/txm/pendingtx.go
amit-momin: I think we would want to still keep the state in `pendingTx` too. Think this would help us avoid having to aggregate the state across different signatures and we can keep a lot of the existing logic the same. We would sort of be aggregating the state on the fly with our state transition methods.

pkg/solana/txm/pendingtx.go
amit-momin: At least from my initial look, I thought we could keep this the same. If the retry context is still alive, that should mean the retry loop is still running so we probably don't need to do anything in that case. But if a confirmed transaction is re-org'd, we would have cancelled the retry context during state transition so that context would no longer be alive. We would still want to restart the retry loop for it though.

pkg/solana/txm/pendingtx.go
Farber98: deprecated

pkg/solana/txm/pendingtx.go
```diff
@@ -556,6 +567,129 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+			pTx = tx
+		}
+
+		if !broadcastedProcessedExists && !confirmedExists {
+			// transaction does not exist in any non finalized/errored maps
+			return "", ErrTransactionNotFound
+		}
+
+		// reset state to broadcasted and update the transaction in the broadcasted map
+		info.state, pTx.state = Broadcasted, Broadcasted
```
amit-momin: I think this should still work but wanted to run an edge case by you. If the TXM is configured to poll more slowly for statuses, there's a chance a confirmed tx gets re-org'd but also processed before we poll for statuses again. We could still detect a re-org if a tx goes from confirmed to processed. But I think this should still work since we handle broadcasted and processed the same way. We could just wait for the next status poll to update it to processed or if it jumps to confirmed by the next time that's ok too

pkg/solana/txm/utils.go
```diff
@@ -51,6 +51,35 @@ func (s TxState) String() string {
...
+
+// isStatusRegression checks if the current status is a regression compared to the previous status:
+// - Confirmed -> Processed, Broadcasted: should not regress
+// - Processed -> Broadcasted: should not regress
+// Returns true if a regression is detected, indicating a possible re-org.
+func isStatusRegression(previous, current TxState) (regressionType, bool) {
+	switch previous {
+	case Confirmed:
+		// Confirmed transactions should not regress to Processed or Broadcasted.
+		if current != Confirmed && current != Finalized && current != Errored {
```
amit-momin: Couldn't we simplify this if by checking for Broadcasted or Processed?
```suggestion
		if current == Broadcasted || current == Processed {
```

pkg/solana/txm/utils.go
```diff
@@ -51,6 +51,35 @@ func (s TxState) String() string {
...
+			return FromConfirmed, true
+		}
+	case Processed:
+		// Processed transactions should not regress to Broadcasted.
+		if current != Processed && current != Confirmed && current != Finalized && current != Errored {
```
amit-momin: Same here
```suggestion
		if current == Broadcasted {
```

pkg/solana/txm/txm.go
```diff
@@ -467,17 +475,19 @@ func (txm *Txm) processConfirmations(ctx context.Context, client client.ReaderWr
...
 				case rpc.ConfirmationStatusFinalized:
 					// if signature is finalized, end polling
 					txm.handleFinalizedSignatureStatus(sig)
-					continue
 				default:
 					txm.lggr.Warnw("unknown confirmation status", "signature", sig, "status", status.ConfirmationStatus)
+				}
+
+				// check if a potential re-org has occurred for this sig and handle it
+				err := txm.handleReorg(ctx, sig, status)
```
amit-momin: Since we can only have a re-org if a status was not found or processed, I think we should move this in those cases so we're not processing the sig statuses even if we get confirmed or finalized status.

pkg/solana/txm/pendingtx.go
```diff
@@ -556,6 +567,129 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+// It achieves this by comparing the highest aggregated state across all associated signatures with the current state of the transaction.
+// If the highest aggregated state is less than the current state, a reorg has occurred and we need to handle it.
+func (c *pendingTxContext) TxHasReorg(id string) bool {
+	var pTx pendingTx
+	var broadcastedExists, confirmedExists bool
+	statePriority := map[TxState]int{
+		Broadcasted: 1,
+		Processed:   2,
+		Confirmed:   3,
+	}
```
amit-momin: Not against this if it helps but I think the `TxState` enum is ordered in this way already. Couldn't we just compare these states against each other without the map?

pkg/solana/txm/utils.go
Farber98: really good use of DeMorgan laws 😄

pkg/solana/txm/pendingtx.go
Farber98: yes! this makes things simpler

pkg/solana/txm/pendingtx.go
Farber98: do you think passing the `currentStatus` to this method and update directly to that state should be better?

amit-momin: Tbh I think it might be better to just keep it this way for simplicity but maybe we can add a comment in case anyone else has the same question. I think it's ok if we always mark it as Broadcasted here. We can pick up that it's really `Processed` on the next cycle. I don't think this introduces any risk with the expiration logic either. We wouldn't falsely pick up this transaction for expiration since we make sure to check any changes in confirmation status before expiration detection.

Farber98: added comment to explain this

pkg/solana/txm/pendingtx.go
```diff
@@ -547,6 +560,148 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+			pTx = tx
+		}
+
+		if !broadcastedProcessedExists && !confirmedExists {
+			// transaction does not exist in any non finalized/errored maps
+			return "", ErrTransactionNotFound
+		}
+
+		// Reset the transaction state to 'Broadcasted' upon detecting a reorg.
+		// Even if the transaction might have already progressed to 'Processed' before the reorg,
```
amit-momin: nit: I think this comment is only relevant if the tx progressed after the re-org but our confirmation polling was too slow
```suggestion
		// Even if the transaction might have already progressed to 'Processed' after the reorg,
```

pkg/solana/txm/txm.go
```diff
@@ -466,19 +476,17 @@ func (txm *Txm) processConfirmations(ctx context.Context, client client.ReaderWr
 				switch status.ConfirmationStatus {
 				case rpc.ConfirmationStatusProcessed:
 					// if signature is processed, keep polling for confirmed or finalized status
+					// we also need to check if a potential re-org has occurred for this sig and handle it
 					txm.handleProcessedSignatureStatus(sig)
-					continue
+					txm.handleReorg(ctx, sig, status)
```
amit-momin: Thinking a little more about this, I think it might be better if we handle re-orgs before `handleProcessedSignatureStatus`. Otherwise, we would falsely log an error in `handleProcessedSignatureStatus` because the tx would not be found in the broadcasted/processed map if a re-org has occurred. If we handle re-orgs first, we would put the tx back into the broadcasted state to avoid the error log.

pkg/solana/txm/pendingtx.go
amit-momin: If we make the change from my other comment, we probably wouldn't even have to wait till the next polling cycle to advance the state. We would just be marking the tx as `Processed` right after we're done handling the re-org. We can probably adjust this comment accordingly if you make that change

pkg/solana/txm/txm.go
Farber98: done! lmk if that is what you were thinking for both cases

pkg/solana/txm/txm.go
```diff
@@ -570,14 +640,14 @@ func (txm *Txm) handleFinalizedSignatureStatus(sig solanaGo.Signature) {
 // An expired tx is one where it's blockhash lastValidBlockHeight (last valid block number) is smaller than the current block height (block number).
 // If any error occurs during rebroadcast attempt, they are discarded, and the function continues with the next transaction.
 func (txm *Txm) rebroadcastExpiredTxs(ctx context.Context, client client.ReaderWriter) {
-	currBlock, err := client.GetLatestBlock(ctx)
-	if err != nil || currBlock == nil || currBlock.BlockHeight == nil {
+	blockHeight, err := client.GetLatestBlockHeight(ctx)
```
Farber98, line 604: this is out of scope of the reorg, but pushing it here so we don't need to make a separate merge: https://smartcontract-it.atlassian.net/browse/NONEVM-1015

pkg/solana/client/client.go
```diff
@@ -331,6 +333,19 @@ func (c *Client) GetLatestBlock(ctx context.Context) (*rpc.GetBlockResult, error
 	return v.(*rpc.GetBlockResult), err
 }
 
+// GetLatestBlockHeight returns the latest block height of the node based on the configured commitment type
```
Farber98, line 336: this is out of scope of the reorg, but pushing it here so we don't need to make a separate merge: [smartcontract-it.atlassian.net/browse/NONEVM-1015](https://smartcontract-it.atlassian.net/browse/NONEVM-1015)

pkg/solana/txm/pendingtx.go
```diff
@@ -112,11 +130,8 @@ func (c *pendingTxContext) New(tx pendingTx, sig solana.Signature, cancel contex
 		return err
 	}
 
-	// upgrade to write lock if sig or id do not exist
+	// upgrade to write lock if id do not exist
```
amit-momin: nit: 
```suggestion
	// upgrade to write lock if id does not exist
```

pkg/solana/txm/pendingtx.go
```diff
@@ -560,6 +579,162 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+			return ErrTransactionNotFound
+		}
+		return nil
+	})
+	if err != nil {
+		// If transaction or sig are not found, return
+		return err
+	}
+
+	var pTx pendingTx
```
amit-momin: nit: think we can move this into the write lock block since we don't really need it outside

pkg/solana/txm/txm.go
```diff
@@ -505,6 +514,82 @@ func (txm *Txm) handleErrorSignatureStatus(sig solanaGo.Signature, status *rpc.S
...
+
+		// Determine if the sig regression affects the transaction state.
+		// If the tx isn't considered re-orged, skip further processing.
+		// Multiple signatures may be in-flight for a single transaction, so a re-org
+		// for one signature doesn't necessarily mean the transaction state has regressed.
+		if !txm.txs.TxHasReorg(txInfo.id) {
+			return
+		}
+
+		txm.lggr.Warnw("re-org detected for transaction", "txID", txInfo.id, "signature", sig, "previousStatus", txInfo.state, "currentStatus", currentTxState)
```
amit-momin: Think we can consider re-orgs as part of the normal TXM processes since it's an expected occurrence eventually. `Warn` might be too alarming here. I think `Info` or even `Debug` might be a better log level here.

amit-momin: Actually let's do `Debug` to keep this the same as our expiration log

pkg/solana/txm/txm.go
```diff
@@ -505,6 +514,82 @@ func (txm *Txm) handleErrorSignatureStatus(sig solanaGo.Signature, status *rpc.S
...
+		// reset tx state to broadcasted in our in memory map
+		err := txm.txs.OnReorg(txInfo.id)
+		if err != nil {
+			txm.lggr.Errorw("failed to handle re-org", "signature", sig, "id", txInfo.id, "error", err)
+			return
+		}
+
+		// For regressions from "Confirmed, we'll need to rebroadcast the tx.
+		if regressionType == FromConfirmed {
```
amit-momin: IIRC I thought I remember you mentioning that if we detect a re-org, we'd have to replace the blockhash in all cases since the re-org implies that the block hash we were using is invalid. So rebroadcasting with the same blockhash would fail anyways. If that's the case, I think we can ditch this `regressionType` and just assign a new blockhash always. We could also skip the specific `confirmed re-orged tx` log below too then.

pkg/solana/txm/txm.go
```diff
@@ -505,6 +514,82 @@ func (txm *Txm) handleErrorSignatureStatus(sig solanaGo.Signature, status *rpc.S
...
+		txm.lggr.Errorw("failed to get signature info when checking for potential re-orgs", "signature", sig, "error", err)
+		return
+	}
+
+	// Check if the sig status has regressed to indicate a re-org.
+	// A regression is identified when the state transitions as follows:
+	// - Confirmed -> Processed || Broadcasted || Not Found
+	// - Processed -> Broadcasted || Not Found
+	currentTxState := convertStatus(status)
+	if regressionType, isRegressed := isStatusRegression(txInfo.state, currentTxState); isRegressed {
```
amit-momin: If we don't need `regressionType` anymore based on my other comment, I think we can just use `TxHasReorg` for this and eliminate `GetSignatureInfo` and `UpdateSignatureStatus` to minimize the number of storage methods we use.
1. We could rework `TxHasReorg` to take in a sig and a current state and return the id and bool. That way we can do the `isStatusRegression` on the storage side and we can drop `GetSignatureInfo`
2. We could also update `TxHasReorg` to fetch the id from the sig map to fetch the tx. Since we'd have access to the current sig and its new state, I don't think we'd have to call `UpdateSignatureStatus`. We'd be dropping all old sigs anyways.

pkg/solana/txm/txm.go
amit-momin: Even if that's not an issue, we can maybe still consider keeping the same logic for both cases if it helps simplify things. What do you think?

pkg/solana/txm/txm_integration_test.go
```diff
@@ -185,3 +189,185 @@ func createTransaction(ctx context.Context, t *testing.T, client *solanaClient.C
...
+		finalSenderBalance, err := cl.Balance(ctx, senderPubKey)
+		require.NoError(t, err)
+		finalReceiverBalance, err := cl.Balance(ctx, receiverPubKey)
+		require.NoError(t, err)
+		require.Less(t, finalSenderBalance, initSenderBalance, "Sender balance should decrease after re-finalization")
+		require.Equal(t, amount, finalReceiverBalance, "Receiver should receive transferred amount after re-finalization")
+		status, errGetStatus = txmInstance.GetTransactionStatus(ctx, txID)
+		require.NoError(t, errGetStatus)
+		require.Equal(t, types.Finalized, status, "tx should be finalized after reorg")
+	})
```
amit-momin, line 303: Could you also include a test case where we detect a re-org from `Processed` state?

pkg/solana/txm/utils.go
```diff
@@ -51,6 +51,33 @@ func (s TxState) String() string {
...
 
+type regressionType int
+
+const (
+	FromConfirmed regressionType = iota
+	FromProcessed
+)
+
+// isStatusRegression checks if the current status is a regression compared to the previous status:
+// - Confirmed -> Processed, Broadcasted, Not Found: should not regress
```
prashantkumar1982: nit: Comment part ": should not regress" doesn't make sense.
Did you mean that any state change from confirmed/processed to broadcasted/NotFound means that it did regress?
Atleast that's that the code says, so the comment isn't reflecting the code.

pkg/solana/txm/txm.go
Farber98: Hey! Great catch with this one. This approach simplifies the design a lot. Let me test it first to ensure we don’t unintentionally introduce races on `Processed` ones, though I have a good feeling about this.

To back up our thoughts, here are some relevant quotes from the Solana Docs:

**What is a [blockhash](https://solana.com/docs/advanced/confirmation#what-is-a-blockhash)?**
> A blockhash refers to the last Proof of History (PoH) hash for a slot. Since Solana uses PoH as a trusted clock, a transaction's recent blockhash can be thought of as a timestamp.

**Proof of [History refresher](https://solana.com/docs/advanced/confirmation#proof-of-history-refresher)**
> Solana's Proof of History mechanism uses a very long chain of recursive SHA-256 hashes to build a trusted clock. The “history” part of the name comes from the fact that block producers hash transaction id's into the stream to record which transactions were processed in their block.
> 
> PoH can be used as a trusted clock because each hash must be produced sequentially. Each produced block contains a blockhash and a list of hash checkpoints called “ticks” so that validators can verify the full chain of hashes in parallel and prove that some amount of time has actually passed.


**How does [transaction expiration](https://solana.com/docs/advanced/confirmation#how-does-transaction-expiration-work) work?**

> Each transaction includes a “recent blockhash” which is used as a PoH clock timestamp and expires when that blockhash is no longer “recent enough”.
> 
> As each block is finalized, the final hash of the block is added to the BlockhashQueue which stores a maximum of the 300 most recent blockhashes. During transaction processing, Solana Validators will check if each transaction's recent blockhash is recorded within the most recent 151 stored hashes (aka "max processing age"). If the transaction's recent blockhash is older than this max processing age, the transaction is not processed.

In conclusion, I have a good feeling this approach will be viable. If a block gets reorged, my understanding is that its blockhash will no longer be present in the mentioned `BlockhashQueue`. Therefore, any retry attempt using the previous blockhash would not be accepted, mitigating the risk of duplicate transactions. The only transaction that should go through would be the new one created with a fresh (and hopefully valid) blockhash.

pkg/solana/txm/txm_integration_test.go
Farber98: I couldn’t achieve this because the local live validator transitions to `Confirmed` instantly. I can try spamming a large number of transactions to overwhelm it, which might delay the transition and provide a window to stop it

pkg/solana/txm/txm.go
Farber98: I like this

pkg/solana/txm/txm.go
Farber98: Hey, we’re considering the premise that a signature regression implies an invalid blockhash. If this holds true, it seems we wouldn’t need to check and aggregate state across all signatures since they all rely on the same blockhash, which would likely be invalid as well.

Are we in the same page with this?

amit-momin: Ya I was thinking the same thing. Not 100% sure if this can happen but I also want to be careful with potentially mistaking a status regression with a lagging RPC. In case the multinode switches to a new node that does not have a processed block with 1 of our tx propagated to it, we could maybe falsely consider the signature regressed. Maybe this risk is low enough, if it's real, and not that impactful that we can ignore it though. Aggregating statuses feels like could reduce the risk but we'd probably still have it if bumping is disabled and only have 1 valid sig at a time anyways.

pkg/solana/txm/txm_integration_test.go
amit-momin: Ah gotcha. In that case I think this is ok. Your mock testing for that scenario might be enough

pkg/solana/txm/txm.go
prashantkumar1982: +1
In general, what's the harm in using a new BlockHash whenever we suspect a reorg?

pkg/solana/txm/txm.go
```diff
@@ -926,6 +992,34 @@ func (txm *Txm) InflightTxs() int {
 	return len(txm.txs.ListAllSigs())
 }
 
+// rebroadcastWithGivenBlockhash attempts to rebroadcast a pending tx with a new blockhash.
+// Removes all signatures associated with the prior tx, cancels prior ctx, updates compute unit price and sets given blockhash for rebroadcasting.
+// Calls sendWithRetry directly to avoid enqueuing the transaction. It logs the error when rebroadcast fails and returns the new signature when successful.
+func (txm *Txm) rebroadcastWithGivenBlockhash(ctx context.Context, pTx pendingTx, blockhash solana.Hash, lastValidBlockHeight uint64) (solana.Signature, error) {
+	// Removes all signatures associated to prior tx and cancels context.
```
prashantkumar1982: nit: Just say "Remove the previous tx from state".
Because it does that, in addition to signatures.

pkg/solana/txm/pendingtx.go
```diff
@@ -47,6 +48,18 @@ type PendingTxContext interface {
...
 	// TrimFinalizedErroredTxs removes transactions that have reached their retention time
 	TrimFinalizedErroredTxs() int
+	// GetSignatureInfo returns the transaction ID and TxState for the provided signature
+	GetSignatureInfo(sig solana.Signature) (txInfo, error)
+	// UpdateSignatureStatus updates the status of a signature in the SigToTxInfo map
+	UpdateSignatureStatus(sig solana.Signature, status TxState) error
+	// TxHasReorg determines whether a reorg has occurred for a given tx.
+	// It achieves this by comparing the highest aggregated state across all associated signatures with the current state of the transaction.
+	// If the highest aggregated state is less than the current state, a reorg has occurred and we need to handle it.
+	TxHasReorg(id string) bool
```
prashantkumar1982: nit: rename to IsTxReorged

The previous name, to me, looked like was telling this interface that a reorg has occurred.

pkg/solana/txm/pendingtx.go
```diff
@@ -81,20 +99,20 @@ type pendingTxContext struct {
...
 }
 
-func (c *pendingTxContext) New(tx pendingTx, sig solana.Signature, cancel context.CancelFunc) error {
+func (c *pendingTxContext) New(tx pendingTx) error {
 	err := c.withReadLock(func() error {
-		// validate signature does not exist
-		if _, exists := c.sigToID[sig]; exists {
-			return ErrSigAlreadyExists
+		// Check if ID already exists in any of the maps
+		if _, exists := c.broadcastedProcessedTxs[tx.id]; exists {
```
prashantkumar1982, line 115: Now that you're using tx.id for already existing entries, won't that cause unnecessary errors when calling this function from rebroadcastExpiredTxs()?

I see that rebroadcastExpiredTxs() will collect all expired txs, and then call txm.sendWithRetry() on them, using the same tx.id. That looks to me will fail because this function will return the error ErrIDAlreadyExists.

pkg/solana/txm/txm.go
Farber98: Thanks for your comments guys! I’ve updated the code so that both cases are handled the same way 🙌. It looks much cleaner now.

> Not 100% sure if this can happen but I also want to be careful with potentially mistaking a status regression with a lagging RPC

How can we evaluate this risk? Should we reach out to Dylan for more insight?

> In general, what's the harm in using a new BlockHash whenever we suspect a reorg?

This approach is the best one since a regression could imply an invalid blockhash. It’s better to be on the safe side and rebroadcast with a new one as you suggest

pkg/solana/txm/pendingtx.go
Farber98: It should be handled gracefully. The `rebroadcastExpiredTxs` function calls [rebroadcastWithGivenBlockhash method in this line](https://github.com/smartcontractkit/chainlink-solana/blob/nonevm-984-reorg/pkg/solana/txm/txm.go#L629) where we [remove the previous tx from state](https://github.com/smartcontractkit/chainlink-solana/blob/404708355148ebdfdd9737b1d46d8d6393c72e89/pkg/solana/txm/txm.go#L963) before calling [sendWithRetry](https://github.com/smartcontractkit/chainlink-solana/blob/404708355148ebdfdd9737b1d46d8d6393c72e89/pkg/solana/txm/txm.go#L975)

pkg/solana/txm/utils.go
Farber98: Thanks for the feedback! Ended up removing the method, and handling this within [IsTxReorged](https://github.com/smartcontractkit/chainlink-solana/blob/ae95e4ff704f3b4f0136344804698f40f2c56c4c/pkg/solana/txm/pendingtx.go#L589-L602) body. It feels cleaner to have it in-line rather than using a separate function on a separate file

pkg/solana/txm/txm_internal_test.go
```diff
@@ -510,59 +511,6 @@ func TestTxm(t *testing.T) {
 				mc.On("SendTx", mock.Anything, tx).Panic("SendTx should not be called anymore").Maybe()
 			})
 
-			// tx passes sim, shows processed, moves to nil (timeout should cleanup)
```
Farber98, line 513: removed this because it's basically a reorg that is tested below

amit-momin: Hmm it might still be worth keeping this test to make sure a transaction that doesn't reach confirmed state in time is still cleaned up even if we try to handle re-orgs.

pkg/solana/txm/utils/utils.go
```diff
@@ -51,6 +51,13 @@ func (s TxState) String() string {
 	}
 }
 
+type regressionType int
```
amit-momin: Think we aren't using this anymore right?

pkg/solana/txm/pendingtx_test.go
```diff
@@ -1327,3 +1408,163 @@ func TestPendingTxContext_ListAllExpiredBroadcastedTxs(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestPendingTxContext_GetPendingTx(t *testing.T) {
+	t.Parallel()
+	txs := newPendingTxContext()
+
+	t.Run("successfully retrieve broadcasted transaction", func(t *testing.T) {
```
amit-momin, line 1544: nit: based on the code I think the description should be
```suggestion
	t.Run("successfully retrieve processed transaction", func(t *testing.T) {
```

amit-momin: Might also be worth adding a test case for Broadcasted too though?

pkg/solana/txm/txm.go
amit-momin: We could ask Dylan but that might be a Solana chain behavior he might not be too familiar with. Maybe we can check with him what the multinode considers when choosing the lead RPC? Regardless I think this scenario might be rare and it doesn't have any extreme implications on the TXM process. We'd really just spend some extra gas on it.

pkg/solana/txm/utils/utils.go
Farber98: ohh good catch, I forgot about it

pkg/solana/txm/pendingtx_test.go
Farber98: good catch, adding it

pkg/solana/txm/txm_internal_test.go
Farber98: Added again. Refactored so instead of regressing it just stays processed and then is cleaned up

amit-momin: I think the test `fail_confirm_processed` right above this is already testing this behavior. I think it's ok if we regress to nil and rebroadcast the tx. If the rebroadcasted tx also stalls at processed, we just want to make sure that's cleaned up as well

Farber98: addressed!

amit-momin (approved):

pkg/solana/txm/pendingtx.go
```diff
@@ -563,6 +576,52 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+	if !exists {
+		return "", false
+	}
+
+	// Compare our in-memory state of the sig with the current on-chain state to determine if the sig had a regression
+	sigInMemoryState := txInfo.state
+	var hasReorg bool
+	switch sigInMemoryState {
+	case txmutils.Confirmed:
+		if sigOnChainState == txmutils.Processed || sigOnChainState == txmutils.Broadcasted || sigOnChainState == txmutils.NotFound {
```
silaslenihan, line 594: Wouldn't there be a case where the in-memory state differs from the on-chain state due to the chain being polled more recently, which wouldn't be a true reorg?

pkg/solana/txm/pendingtx.go
```diff
@@ -563,6 +576,52 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
...
+	defer c.lock.RUnlock()
+	var tx, tempTx pendingTx
+	var broadcastedExists, confirmedExists bool
+	if tempTx, broadcastedExists = c.broadcastedProcessedTxs[id]; broadcastedExists {
+		tx = tempTx
+	}
+	if tempTx, confirmedExists = c.confirmedTxs[id]; confirmedExists {
+		tx = tempTx
+	}
+
```
silaslenihan, line 618: Shouldn't it be impossible for a tx to exist in both confirmed and broadcasted since it's deleted during `OnCofirmed`? Not saying this is a huge issue mostly just testing my understanding

pkg/solana/txm/pendingtx.go
```diff
@@ -66,14 +75,21 @@ type pendingTx struct {
...
+	state txmutils.TxState
 }
 
 var _ PendingTxContext = &pendingTxContext{}
 
 type pendingTxContext struct {
-	cancelBy map[string]context.CancelFunc
-	sigToID  map[solana.Signature]string
+	cancelBy    map[string]context.CancelFunc
+	sigToTxInfo map[solana.Signature]txInfo
```
silaslenihan, line 92: Why not just use the existing state maps? If these diverge it could be an issue so wouldn't it be better to have a singular source of truth?

pkg/solana/txm/pendingtx.go
```diff
@@ -563,6 +576,52 @@ func (c *pendingTxContext) TrimFinalizedErroredTxs() int {
 	return len(expiredIDs)
 }
 
+func (c *pendingTxContext) IsTxReorged(sig solana.Signature, sigOnChainState txmutils.TxState) (string, bool) {
```
silaslenihan, line 579: Might be useful to add some comments above these new functions

pkg/solana/txm/txm.go
```diff
@@ -499,6 +508,51 @@ func (txm *Txm) handleErrorSignatureStatus(sig solanaGo.Signature, status *rpc.S
...
 
+// handleReorg detects and manages state regressions (re-orgs) for a given signature.
+//
+// A re-org occurs when the on-chain state of a signature regresses as follows:
+// - Confirmed -> Processed || Not Found
+// - Processed -> Not Found
+//
+// When a signature re-org is detected, the following steps are taken:
+// - Remove the prior transaction, along with all associated signatures, and cancel the prior context.
+// - Rebroadcast the prior transaction with a new blockhash and an updated compute unit price.
```
silaslenihan, line 519: Kudos on the comments and readability on this function!

silaslenihan (approved):

pkg/solana/txm/pendingtx.go
Farber98: I added it at the [interface level](https://github.com/smartcontractkit/chainlink-solana/blob/ea4d0fa7633b7a714c38ddf38b1872f61c6dbfe5/pkg/solana/txm/pendingtx.go#L54-L59), thinking it's the first place a user of the module would check.

I'm OK with duplicating the comment if it improves readability, but we would need to ensure consistency between both places, which could become an issue. Maybe we could add a smaller comment pointing to the interface for the more detailed explanation?

silaslenihan: Ah, no that's fine sorry I missed that!

pkg/solana/txm/pendingtx.go
Farber98: Yes, this shouldn't happen. What is your concern about this?

silaslenihan: Well this function seems to be okay with that case where `broadcastedExists` and `confirmedExists`, I guess the reorg would just be caught elsewhere?

pkg/solana/txm/pendingtx.go
Farber98: There's no problem that we have different states as long as the on-chain one is greater or equal than our in-memory one. 

If this premise is false, it's because:
- A re-org occurred and we need to handle it 
- [Potential Lagging RPC edge case](https://github.com/smartcontractkit/chainlink-solana/pull/951#discussion_r1902131441) 
	- We are not sure if this is a real risk

silaslenihan: > https://github.com/smartcontractkit/chainlink-solana/pull/951#discussion_r1902131441

Yeah this is what I was getting at... I guess you guys have already thought it through though and nothing has come up in soak testing

pkg/solana/txm/pendingtx.go
Farber98: We catch the reorgs here: 

- When [getting a NotFound](https://github.com/smartcontractkit/chainlink-solana/blob/ea4d0fa7633b7a714c38ddf38b1872f61c6dbfe5/pkg/solana/txm/txm.go#L443) sig, we check if our in-memory state is higher for it (`Processed` or `Confirmed`) and needs to be handled
- When [getting a Processed signature](https://github.com/smartcontractkit/chainlink-solana/blob/ea4d0fa7633b7a714c38ddf38b1872f61c6dbfe5/pkg/solana/txm/txm.go#L458) sig, if our in-memory state is higher for it (`Confirmed`) and needs to be handled
- By definition, no reorgs from `Finalized` state should happen, that's why they are not handled

prashantkumar1982 (approved):

jadepark-dev (approved):
Amazing work, looks good to me!

PR #953, opened 2024-12-02
logs are forwarded to a processor in slot and trx order

EasterTheBunny:
### Description

Log ingestion heavily leverages async go-routines but `LogPoller` will rely on logs delivered in slot/block order. This commit adds a new component to the log ingestion to order logs before forwarding them to `LogPoller`.


pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,218 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+		}
+
+		// if expectations set and met -> forward, remove, and continue
+
+		// to ensure ordered delivery, break from the loop if a ready block isn't found
+		// this function should be preceded by clearEmptyBlocks
+		rIdx, ok := getIdx(p.ready, block)
+		if !ok {
+			return nil
+		}
```
reductionista, line 461:
```suggestion
		rIdx := slices.Index(p.ready, block)
		if rIdx == -1 {
			return nil
		}
```

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,218 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+			p.clearExpectations(block)
+			rmvIdx = append(rmvIdx, idx)
+
+			continue
+		}
+
+		// if expectations set and met -> forward, remove, and continue
+
+		// to ensure ordered delivery, break from the loop if a ready block isn't found
+		// this function should be preceded by clearEmptyBlocks
```
reductionista, line 457: Which function is this comment referring to? Not sure I follow

reductionista: Not sure I understand this comment. Which function is it referring to that's preceded by `cleanEmptyBlocks`?

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,218 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+			return idx, true
+		}
+	}
+
+	return -1, false
+}
+
+func remove[T any](slice []T, s int) []T {
+	return append(slice[:s], slice[s+1:]...)
+}
```
reductionista: Could use `slices.Delete`:
https://pkg.go.dev/slices#Delete

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,218 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+
+func getIdx[T any](slice []T, match T) (int, bool) {
+	for idx, value := range slice {
+		if reflect.DeepEqual(value, match) {
+			return idx, true
+		}
+	}
+
+	return -1, false
+}
```
reductionista: Could use `slices.Index` instead of defining are own:
https://pkg.go.dev/slices#Index

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,218 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.blocks = append(p.blocks, block)
+
+	// ensure sort ascending
+	sort.Slice(p.blocks, func(i, j int) bool {
+		return p.blocks[i] < p.blocks[j]
+	})
+}
```
reductionista, line 358: This is a perfect use case for a heap:
https://pkg.go.dev/container/heap#pkg-index

Instead of re-sorting the list each time a new block is added, you just call `h.Push()` to add each block and it inserts it efficiently into the tree. Then calling `h.Pop()` will remove them in order.

pkg/solana/logpoller/loader.go
EasterTheBunny: I think this was just a comment from a previous iteration. I'll remove it.

pkg/solana/logpoller/loader.go
EasterTheBunny: Good to know. I have not used the `slices` package yet.

pkg/solana/logpoller/loader.go
EasterTheBunny: Same here on the `slices` package.

pkg/solana/logpoller/loader.go
```diff
@@ -475,34 +469,44 @@ func (p *orderedParser) sendReadySlots() error {
...
+	return x
 }
 
-var (
-	errExpectationsNotSet = errors.New("expectations not set")
-)
+func (h *blockHeap) Delete(idx int) {
+	old := *h
+	*h = slices.Delete(old, idx, idx+1)
+}
```
reductionista: I don't think you can use `slices.Delete` with a heap, because it has to do do some fancy rebalancing while removing an element to preserve the heap invariant.

I think we just need to call:
```
heap.Remove(h, idx)
```
and then it does it all for you. No need to define a custom `Delete` method

pkg/solana/logpoller/loader.go
```diff
@@ -434,7 +431,7 @@ func (p *orderedParser) sendReadySlots() error {
 	rmvIdx := make([]int, 0)
 
 	// start at the lowest block and find ready blocks
-	for idx, block := range p.blocks {
+	for idx, block := range *p.blocks {
```
reductionista: Iterating over them directly like this won't work right. The order of the elements of a heap is well-defined, but not necessarily the same order as the underlying elements of the backing slice.
To get the right order, you pop them one at a time. Something like this:
```
for p.blocks.Len() > 0 {
	block := heap.Pop(p.blocks)
	...
}
```
I left another comment somewhere else suggesting to replace the `Delete` method with a call to `heap.Remove`, but now that I'm looking at this part of the code I realize you shouldn't need either of those, I think `Pop` is all we need?

pkg/solana/logpoller/loader.go
```diff
@@ -451,12 +448,9 @@ func (p *orderedParser) sendReadySlots() error {
 			continue
```
reductionista, line 449: I see... you'll probably want to delay the `heap.Pop` call I have above till here, so that if you break or return early it's not already removed. Instead of the initial `heap.Pop` you can "peak" at the minimum block number on the heap with `block := (*p.blocks)[0]`. Even though the rest may be out of order, this should always be guaranteed to be the min element.

I think the two places you'll need the `heap.Pop` are in place of the 2 instances of `rmvIdx = append(rmvIndex, idx)`, does that sound right?

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,222 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+			errs = errors.Join(errs, p.parser.Process(evt))
+		}
+
+		// need possible retry
+		if errs != nil {
+			return errs
+		}
+
+		p.ready = slices.Delete(p.ready, rIdx, rIdx+1)
+		rmvIdx = append(rmvIdx, idx)
```
reductionista:
```suggestion
		_ = heap.Pop(p.blocks)
```

pkg/solana/logpoller/loader.go
reductionista: Update: you shouldn't need either this or `heap.Remove`, just `heap.Pop`--see other comments

pkg/solana/logpoller/loader.go
reductionista: Update: replace `heap.Pop` line above with `block := (*p.blocks)[0]`, to delay the actual Pop for later when you are sure it can be removed.

reductionista:
Almost there, I think this is coming together. Sorry for updating a couple of the comments after writing them hopefully it's clear what needs to be done. I think once you've got the iteration replaced with peaking and popping the heap it should simply things a bit.

pkg/solana/logpoller/loader.go
EasterTheBunny: I'm not sure the `heap.Pop` call is useful in this instance though. Pop will remove the first element, but all I want to do is iterate through them in order without alterations. Then, when I want to remove specific indexes, Pop will not work. I need to use `heap.Remove` to specify which index to remove.

pkg/solana/logpoller/loader_test.go
```diff
@@ -53,42 +58,178 @@ func TestEncodedLogCollector_ParseSingleEvent(t *testing.T) {
...
+
+			return &result, nil
+		})
 
 	tests.AssertEventually(t, func() bool {
 		return parser.Called()
 	})
+}
+
+func TestEncodedLogCollector_MultipleEventOrdered(t *testing.T) {
```
reductionista, line 113: I think this test only passes because it's testing only 2 slots. For a sequence of 3 or more slots I think it should fail with the existing code.

The reason 2 slots passes is because the first element in the heap is always the first element in the slice backing it. But beyond that I think it depends on the order in which elements were added. Can we make it test 3 or more slots, to catch this regression?

EasterTheBunny: I updated the test to have 4 slots with blocks and two are returned slow to force out of order loading. The test still passes.

pkg/solana/logpoller/loader.go
reductionista: Maybe I'm missing something, but is there a reason why you can't just Pop each block as soon as you've verified that the expectations are met? I don't see any case where you go back and look at old blocks or look ahead at future blocks, so I'm not sure why we have to flag the block for removal and then go back through later and locate all the ones flagged for removal? If we're done with it, why not just discard it and move on?

pkg/solana/logpoller/loader_test.go
reductionista: Hmm... I think it must still be a false positive for some reason, I'm pretty sure the code as written does not guarantee the order.

Just ran the debugger on this test, and I notice that the elements are always getting added to `p.blocks` in ascending order (41, 42, 43, 44).  If that's guaranteed to happen in production, then we can probably just go back to having it be an ordinary slice rather than a heap, and don't need to sort it.  But I suspect there is a way it can happen--we just need to find a way of testing it. Maybe it would happen if there are two different requests for `GetBlocks()` sent and those came back out of order?  I see that you added a 1s delay in the mocked `GetBlockWithOpts` request... but I think that doesn't make a difference here, because `ExpectBlock()` is called before that's ever called.

reductionista: I added a more granular test `TestBlockHeap` to check that out of order elements added are iterated over in order:
https://github.com/smartcontractkit/chainlink-solana/pull/975

It does fail for ordinary iteration and succeeds when using `Pop`

reductionista: I think we should add this test and fix the iteration, then comment out the part of the test that's failing.

Would be nice to also have a test where multiple calls to `GetBlocks` come back in the wrong order, but maybe this is enough?

pkg/solana/logpoller/loader.go
dhaidashenko: +1 to Domino's comment. I do not see a reason to defer removal from blocks once we've verified that block met expectations.

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,217 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+		// if expectations are 0 -> remove and continue
+		if exp == 0 {
+			p.clearExpectations(block)
+			rmvIdx = append(rmvIdx, idx)
+
+			continue
+		}
+
+		// to ensure ordered delivery, break from the loop if a ready block isn't found
+		rIdx := slices.Index(p.ready, block)
```
dhaidashenko: Why do we need p.ready to guarantee order? Is it possible that p.expectations returns true, when p.ready does not contain the block?

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,217 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+		}
+
+		evts, ok := p.actual[block]
+		if !ok {
+			return errors.New("invalid state")
+		}
+
+		var errs error
+		for _, evt := range evts {
+			errs = errors.Join(errs, p.parser.Process(evt))
```
dhaidashenko, line 459: We should pass to the processor the whole block and all related events instead of a single event. 
This way, it knows that the whole block was processed instead of a single event. (On the startup we won't need to refetch data for the block of latest log).
Also, for finality detection work, it would be helpful to store even empty blocks.

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,217 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+
+		// need possible retry
+		if errs != nil {
+			return errs
+		}
+
+		p.ready = slices.Delete(p.ready, rIdx, rIdx+1)
+		rmvIdx = append(rmvIdx, idx)
+
+		delete(p.expect, block)
```
dhaidashenko: clearExpectations already deletes from p.expect and p.actual

pkg/solana/logpoller/loader_test.go
```diff
@@ -53,42 +58,204 @@ func TestEncodedLogCollector_ParseSingleEvent(t *testing.T) {
...
+
+					break
+				}
+			}
+
+			// imitate loading block data out of order
+			if slot == 43 || slot == 41 {
+				// force slot 43 to return after 44
+				// same for 41 after 42
+				time.Sleep(1 * time.Second)
```
dhaidashenko: This may cause flakes when running in CI, it would be nice to use waitGroups or different sync method.

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +291,217 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+
+	heap.Push(p.blocks, block)
+}
+
+func (p *orderedParser) ExpectTxs(block uint64, quantity int) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.expect[block] = quantity
+	p.actual[block] = make([]ProgramEvent, 0, quantity)
```
dhaidashenko, line 365: nit: if quantity is 0 - signal that bock is ready

pkg/solana/logpoller/loader_test.go
dhaidashenko: I agree with Domino that iteration over the heap's slice does not guarantee order. 
Also I do not see why we need heap or sorting? Do we expect that `GetBlocks` may return blocks out-of-order? It seems like linked should be enough.

pkg/solana/logpoller/loader.go
EasterTheBunny: Domino brought this up as well in conversation and she will be doing the interface change as a followup.

pkg/solana/logpoller/loader.go
EasterTheBunny: The values in `p.ready` are set after the actual block data is retrieved (all transactions). The values in `p.expectations` is simply that we eventually expect something from those blocks.

reductionista (dismissed):
Alright, LGTM!

pkg/solana/logpoller/loader_test.go
reductionista: I think technically it can't cause a flake, worst case they would end up in order instead of out of order so it wouldn't be fully testing the ordering on that run. But still, better to avoid relying on a Sleep for testing this

pkg/solana/logpoller/loader.go
reductionista: I hadn't even thought of the reasons @dhaidashenko gives, just was worried that writing 1 event at a time would put more strain on the db than writing batches at a time as they become available.

aalu1418 (dismissed):

pkg/solana/logpoller/loader.go
dhaidashenko: But it returns `met` which is only true when we've actually fetched all transaction. Am I missing something?

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +303,191 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+			errs = errors.Join(errs, p.parser.Process(evt))
+		}
+
+		// need possible retry
+		if errs != nil {
+			return errs
+		}
+
+		p.ready = slices.Delete(p.ready, rIdx, rIdx+1)
+		p.blocks = p.blocks[1:]
```
dhaidashenko: Seems like this is a memory leak. We never rest `p.blocks`

EasterTheBunny: Why do we need to reset `p.blocks`? This simply removes the first element.

reductionista: If this were an array of pointers or slices, it would be a memory leak without doing `p.blocks[0] = nil`. But at least based on this answer, it should not be an issue because it's just an array of primitives:
https://stackoverflow.com/questions/55045402/memory-leak-in-golang-slice

reductionista: Actually, maybe that answer is wrong? there's an example of a memory leak here with `[]byte` that does seem similar to what we're doing here:
https://go.dev/blog/slices-intro

reductionista: Would be curious to hear @jmank88 's thoughts, is using a slice of primitives as a FIFO queue like this considered a memory leak?

EasterTheBunny: I asked @jmank88 about this and he said it could be a memory leak like @dhaidashenko says if used incorrectly. However, in this case since `append` is being used to add values, Go _should_ resize the underlying array correctly. To be safe, since we don't know the magic underneath for certain in this circumstance *and* the array should remain small anyway, I will use the linked list.

pkg/solana/logpoller/loader.go
```diff
@@ -279,3 +304,189 @@ func (c *EncodedLogCollector) loadSlotBlocksRange(ctx context.Context, start, en
...
+func (p *orderedParser) run(_ context.Context) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	_ = p.sendReadySlots()
+}
+
+func (p *orderedParser) sendReadySlots() error {
+	// start at the lowest block and find ready blocks
+	for element := p.blocks.Front(); element != nil; element = element.Next() {
```
dhaidashenko:
```suggestion
	for element := p.blocks.Front(); element != nil; element = p.blocks.Front() {
```
To drop 
```
                temp := element.Prev()

		if temp == nil {
			break
		}

		element = temp
```

dhaidashenko: Also due to break, there is a leak of expectations.

EasterTheBunny: Ahh ok. I was just using the iteration method in the docs. This works better. Thanks

dhaidashenko (approved):

reductionista (approved):

jadepark-dev (approved):
LGTM

PR #959, opened 2024-12-04
LogPoller DSL Parser

EasterTheBunny:
### Description

This PR adds a DSL parser that converts basic query filters into postgresql queries for logpoller. Additionally solana address and transaction hashes are handled.

[NONEVM-744](https://smartcontract-it.atlassian.net/browse/NONEVM-744)

### Requires Dependencies

- https://github.com/smartcontractkit/chainlink-solana/pull/921

[NONEVM-744]: https://smartcontract-it.atlassian.net/browse/NONEVM-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhaidashenko (dismissed):

pkg/solana/logpoller/query.go
```diff
@@ -31,6 +31,12 @@ func (q *queryArgs) withField(fieldName string, value any) *queryArgs {
 	return args
 }
```
reductionista, line 33: I just noticed there is a `parser_test.go` but no `query_test.go`, while in evm logpoller we have both. Are the functions in `query.go` unit tested anywhere?

Seems like we should have unit tests for them, but not sure whether they should go in this PR or in @dhaidashenko 's db models PR it's based on.

pkg/solana/logpoller/orm.go
```diff
@@ -156,3 +157,27 @@ func (o *DSORM) SelectLogs(ctx context.Context, start, end int64, address Public
 	}
 	return logs, nil
 }
+
+func (o *DSORM) FilteredLogs(ctx context.Context, filter []query.Expression, limitAndSort query.LimitAndSort, _ string) ([]Log, error) {
```
reductionista, line 205: I was thinking the `FilteredLogs` method that calls this would also be in this PR, but I guess it makes more sense if I just include it in my Main LogPoller Loop PR... probably too difficult to add it here without the struct that's defined in RegisterFilters PR.

pkg/solana/logpoller/parser_test.go
```diff
@@ -0,0 +1,278 @@
...
+					}},
+				},
+				BoolOperator: query.AND,
+			}},
+		}
+		limiter := query.LimitAndSort{}
+
+		result, args, err := parser.buildQuery(chainID, expressions, limiter)
+		expected := logsQuery(
+			" WHERE chain_id = :solana_chain_id " +
```
reductionista: I think this should just be `:chain_id` rather than `:solana_chain_id`, to match the db schema.

Also, hoping we can move this code to `chainlink-framework` at some point to re-use it for many chain families, abstracting it a bit to fill in the custom data types.

pkg/solana/logpoller/orm.go
reductionista: Oh, I do think we should add some tests of this function to `orm_tests.go`.

In evm, there are several tests that test it in different ways:  `TestORM_IndexedLogs`,  `TestORM_SelectLogsCreatedAfter`, `TestORM_SelectIndexedLogsByTxHash`, `TestORM_SelectLogsWithSigsByBlockRangeFilter`. Those also test the corresponding legacy functions which we don't need. But maybe we could just combine the parts that test `FilteredLogs` together into a single `TestFilteredLogs` here?

pkg/solana/logpoller/parser.go
```diff
@@ -1,5 +1,413 @@
...
+)
+
 var (
-	logsFields = [...]string{"id", "filter_id", "chain_id", "log_index", "block_hash", "block_number", "block_timestamp", "address", "event_sig", "subkey_values", "tx_hash", "data", "created_at", "expires_at", "sequence_num"}
+	ErrInvalidComparator   = errors.New("invalid comparison operator")
+	ErrInvalidConfidence   = errors.New("invalid confidence level; solana only supports finalized")
+	ErrInvalidCursorDir    = errors.New("invalid cursor direction")
+	ErrInvalidCursorFormat = errors.New("invalid cursor format")
+	ErrInvalidSortDir      = errors.New("invalid sort direction")
+	ErrInvalidSortType     = errors.New("invalid sort by type")
```
reductionista, line 33: Good call on defining these all together! This will help later with abstraction.

pkg/solana/logpoller/parser.go
```diff
@@ -1,5 +1,413 @@
...
+	limit := v.limitClause(limiter)
+	if len(limit) > 0 {
+		clauses = append(clauses, limit)
+	}
+
+	return strings.Join(clauses, " "), v.args, nil
+}
+
+func (v *pgDSLParser) whereClause(expressions []query.Expression, limiter query.LimitAndSort) (string, error) {
+	segment := fmt.Sprintf("WHERE %s = :solana_chain_id", chainIDFieldName)
```
reductionista:
```suggestion
	segment := fmt.Sprintf("WHERE %s = :chain_id", chainIDFieldName)
```
(see my comment in `parser_test.go` about this... probably should have left it here instead)

pkg/solana/logpoller/parser.go
```diff
@@ -1,5 +1,413 @@
...
+		Primitive: &eventSigFilter{eventSig: sig},
+	}
+}
+
+func (f *eventSigFilter) Accept(visitor primitives.Visitor) {
+	switch v := visitor.(type) {
+	case *pgDSLParser:
+		v.VisitEventSigFilter(f)
+	}
+}
```
reductionista, line 414: I think we're missing an essential feature here that CCIP will require, and that we won't be able to implement QueryKey without. You have `eventSigFilter` and `eventByAddressFitler`, but we also need an `eventBySubKeyFilter` so that we can filter by subkey.

We'll need that to implement support for `IndexedSequencesKeyFilter` and several other `IndexedSequences...` filters that can be passed to `QueryKey`. CCIP definitely uses some of those.

reductionista: See:  `func (b *EventBinding) encodeComparator(comparator *primitives.Comparator) (query.Expression, error)` in `chainlink/core/services/relay/evm/read/event.go`)

We'll have to call `NewEventBySubKey` there instead of `NewEventByTopicFilter` and `NewEventByWordFilter`.

pkg/solana/logpoller/query.go
reductionista: Update: I think this probably belongs in the other PR, but let me know what you both think.

reductionista (dismissed):
Mostly missing `eventBySubKeyFilter`, but see comments suggesting adding some tests for `FilteredLogs` and one minor change (`:solana_chain_id` -> `:chain_id`)

pkg/solana/logpoller/parser.go
```diff
@@ -1,5 +1,413 @@
...
+	if err != nil {
+		v.err = err
+
+		return
+	}
+
+	v.expression = fmt.Sprintf(
+		"%s = :%s",
+		txHashFieldName,
+		v.args.withIndexedField(txHashFieldName, txHash),
```
dhaidashenko: solana.PublicKey does not implement Value method, so not sure if the query won't error. Might need to wrap it into type defined in `types` file.

pkg/solana/logpoller/query.go
dhaidashenko: They are tested, but indirectly, the DB schema PR provides 70% coverage. Combined with this PR, it should be even higher. Given the lack of complex logic, I'm not sure dedicated tests are needed at this stage.

pkg/solana/logpoller/parser.go
EasterTheBunny: I will be adding this when I do the remapping.

dhaidashenko (approved):

reductionista (approved):

aalu1418 (approved):

PR #960, opened 2024-12-05
MultiNode Adaptor Managed Subscriptions

DylanTinianov:
### Description
Ticket: https://smartcontract-it.atlassian.net/browse/BCFR-213

Added a managed subscription which enables the MultiNode Adaptor to remove subscriptions when `Unsubscribe()` is called.

### Soak Test
Namespace: solana-gauntlet-embedded-6c6ca


pkg/solana/client/multinode/client.go
```diff
@@ -72,11 +72,16 @@ func (m *MultiNodeClient[RPC, HEAD]) RegisterSub(sub Subscription, stopInFLightC
 		return fmt.Errorf("failed to register subscription - all in-flight requests were canceled")
 	default:
 	}
-	// TODO: BCI-3358 - delete sub when caller unsubscribes.
 	m.subs[sub] = struct{}{}
 	return nil
 }
 
+func (m *MultiNodeClient[RPC, HEAD]) removeSub(sub Subscription) {
+	m.subsSliceMu.Lock()
```
dhaidashenko: This causes a deadlock on failed m.registerSub

pkg/solana/client/multinode/client.go
```diff
@@ -244,7 +259,10 @@ func (m *MultiNodeClient[RPC, HEAD]) UnsubscribeAllExcept(subs ...Subscription)
 
 	for sub := range m.subs {
 		if _, keep := keepSubs[sub]; !keep {
+			// Release lock to avoid deadlock on unsubscribe
+			m.subsSliceMu.Unlock()
```
dhaidashenko: It does not seem right. https://go.dev/doc/faq#atomic_maps
Concurrently reading/writing data from a map is not safe. I'm not 100% sure it will occur in this case, but I'd rather take a safer path.
You might be able to call Unsubscribe of underlying subscription instead of managed in the `UnsubscribeAllExcept`

pkg/solana/client/multinode/client_test.go
```diff
@@ -94,6 +94,31 @@ func TestMultiNodeClient_HeadSubscriptions(t *testing.T) {
 			t.Fatal("failed to receive finalized head: ", ctx.Err())
 		}
 	})
+
+	t.Run("Remove Subscription on Unsubscribe", func(t *testing.T) {
```
dhaidashenko: Could you also add a test to verify that UnsubscribeAllExecept won't remove & unsubscribe passed sub?

dhaidashenko (dismissed):

dhaidashenko (changes requested):
left a couple of comments

pkg/solana/client/multinode/client_test.go
DylanTinianov: Yes, I added a test for `UnsubscribeAllExcept`

PR #961, opened 2024-12-05
fix: add chainlink/deployment bump to updater workflow

aalu1418:
### Description

New test dependency introduced (`chainlink/deployments`) - also bump this dep when needed


jadepark-dev (approved):
👍

PR #965, opened 2024-12-09
fix: handle compute unit + lookup tables, and improve flaky tests

aalu1418:
### Description

- fix flaky test for `GetBlocks` slot generation != block generation so it can lag behind - increase timeout to reduce flakiness
- fix compute budget handling instructions for lookup tables (need to account for shifted account indices when using lookup tables)

### DIFF
--- a/pkg/solana/fees/computebudget.go
+++ b/pkg/solana/fees/computebudget.go
@@ -151,6 +151,17 @@ func set(tx *solana.Transaction, baseData instruction, appendToFront bool) error
 
 		// https://github.com/gagliardetto/solana-go/blob/618f56666078f8131a384ab27afd918d248c08b7/transaction.go#L293
 		tx.Message.Header.NumReadonlyUnsignedAccounts++
+
+		// lookup table addresses are indexed after the tx.Message.AccountKeys
+		// higher indices must be increased
+		// https://github.com/gagliardetto/solana-go/blob/da2193071f56059aa35010a239cece016c4e827f/transaction.go#L440
+		for i, ix := range tx.Message.Instructions {
+			for j, v := range ix.Accounts {
+				if int(v) >= programIdx {
+					tx.Message.Instructions[i].Accounts[j]++
+				}
+			}
+		}
 	}
 
 	// get instruction data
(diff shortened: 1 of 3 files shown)


jmank88 (dismissed):

pkg/solana/fees/computebudget_test.go
```diff
@@ -136,6 +136,44 @@ func testSet[V instruction](t *testing.T, builder func(uint) V, setter func(*sol
...
+		receiver, err := solana.NewRandomPrivateKey()
+		require.NoError(t, err)
+
+		// build base tx (no fee)
+		tx, err := solana.NewTransaction([]solana.Instruction{
+			system.NewTransferInstruction(
+				0,
+				key.PublicKey(),
+				receiver.PublicKey(),
+			).Build(),
```
amit-momin: Since we could have a scenario where we need to adjust multiple indices, could we also test that here? I assume you'd probably have to use a custom instruction here.

aalu1418: adjusted the test case to have multiple instructions and multiple indices to adjust!

amit-momin (approved):

pkg/solana/fees/computebudget.go
```diff
@@ -151,6 +151,17 @@ func set(tx *solana.Transaction, baseData instruction, appendToFront bool) error
 
 		// https://github.com/gagliardetto/solana-go/blob/618f56666078f8131a384ab27afd918d248c08b7/transaction.go#L293
 		tx.Message.Header.NumReadonlyUnsignedAccounts++
+
+		// lookup table addresses are indexed after the tx.Message.AccountKeys
+		// higher indices must be increased
+		// https://github.com/gagliardetto/solana-go/blob/da2193071f56059aa35010a239cece016c4e827f/transaction.go#L440
```
agusaldasoro, line 157: If solana-go lib does the same, why do we need a wrapper here? Can not we use just the library instead? I'm thinking that we need to avoid having any other bug on our side in the future

aalu1418: it doesn't do the exact same thing - this comment is mostly just pointing out that lookup table addresses are indexes that are larger than the normal message.accountkeys array

solana-go requires a user to predeclare all instructions and then compile things together
our utility allows modification of an compiled transaction to include compute budget instructions

jadepark-dev (approved):

PR #987, opened 2024-12-19
[Non-EVM-1062] Solana Codec events support, Hookup Fuzz tests and cleanup Codec init

ilija42:


pkg/solana/codec/codec_test.go
```diff
@@ -25,6 +27,27 @@ func TestCodec(t *testing.T) {
 	tester := &codecInterfaceTester{}
 	RunCodecInterfaceTests(t, tester)
 	RunCodecInterfaceTests(t, looptestutils.WrapCodecTesterForLoop(tester))
+
+	t.Run("Events are encode-able and decode-able for a single item", func(t *testing.T) {
```
ilija42, line 32: Events are encoded/decoded in the same way that accounts are. I only tested this one thing which confirms that the IDL event fields are properly parsed into Codec entries, all the encoding variations are already tested by the codec interface tests that are encoding/decoding accounts

PR #991, opened 2024-12-23
[NONEVM-1091] - Refactor/Setup CR init and cfg and codec init

ilija42:
### Description

1. Refactor CR init and codec init to be more like EVM CR
2. CR now discerns codec input and output modifiers
3. Rewrite cfg to match the new development
4. Rewrite cfg validation and unmarshal to allow IDL to be passed in both as a string or structured
5. Align naming for all type naming vars for CR to be genericName<>chainSpecificName


pkg/solana/chainreader/chain_reader.go
```diff
@@ -217,54 +227,99 @@ func (s *SolanaChainReaderService) CreateContractType(readIdentifier string, for
...
-			}
-
-			idlCodec, err := codec.NewIDLAccountCodec(idl, config.BuilderForEncoding(method.Encoding))
-			if err != nil {
-				return err
-			}
-
-			s.lookup.addReadNameForContract(namespace, methodName)
+func (s *SolanaChainReaderService) addCodecDef(forEncoding bool, namespace, genericName string, readType codec.ChainConfigType, idl codec.IDL, idlDefinition interface{}, modCfg commoncodec.ModifiersConfig) error {
+	mod, err := modCfg.ToModifier(codec.DecoderHooks...)
```
silaslenihan, line 234: The ChainWriter should have similar logic to this, correct? Instead of having a codec for each program and adding modifiers to that codec, you have a codec entry for each method essentially

pkg/solana/chainreader/chain_reader.go
```diff
@@ -217,54 +227,99 @@ func (s *SolanaChainReaderService) CreateContractType(readIdentifier string, for
...
+		createRPCOpts(readDefinition.RPCOpts),
+	))
+
+	return nil
+}
+
 // injectAddressModifier injects AddressModifier into OutputModifications.
 // This is necessary because AddressModifier cannot be serialized and must be applied at runtime.
-func injectAddressModifier(outputModifications codeccommon.ModifiersConfig) {
+func injectAddressModifier(inputModifications, outputModifications commoncodec.ModifiersConfig) {
```
silaslenihan: Can you elaborate on this a little bit more? Why would someone use `commoncodec.AddressBytesToStringModifierConfig` instead of `codec.SolanaAddressModifier{}`?

ilija42: Because this  `AddressBytesToStringModifierConfig` is the actual serializeable  modifier config, the actual chain specific impl. which is a function and therefore cannot be serialized is injected during CR startup

pkg/solana/chainreader/chain_reader.go
ilija42: Yes, same as the evm impl.

pkg/solana/config/chain_reader.go
```diff
@@ -8,94 +8,85 @@ import (
...
-	// Encoding defines the type of encoding used for on-chain data. Currently supported
-	// are 'borsh' and 'bincode'.
-	Encoding  EncodingType         `json:"encoding" toml:"encoding"`
-	Procedure ChainReaderProcedure `json:"procedure" toml:"procedure"`
+type ReadDefinition struct {
+	ChainSpecificName   string                      `json:"chainSpecificName"`
+	ReadType            ReadType                    `json:"readType,omitempty"`
+	InputModifications  commoncodec.ModifiersConfig `json:"inputModifications,omitempty"`
+	OutputModifications commoncodec.ModifiersConfig `json:"outputModifications,omitempty"`
+	RPCOpts             *RPCOpts                    `json:"rpcOpts,omitempty"`
```
EasterTheBunny: Are the `RPCOpts` necessary anymore? We are explicitly using finalized commitment and encoding will probably always be `base64`. The only remaining option is for data slices. Do we use that or are we expecting to use that?

ilija42: Idk, I haven't looked into this yet, just kept it because it was already there

ilija42: it can most likely be removed

silaslenihan (approved):
LGTM

jadepark-dev (approved):

PR #1002, opened 2025-01-07
[NONEVM-916] logpoller log processing & decoding

reductionista:
JIRA: [LINK-916](https://smartcontract-it.atlassian.net/browse/NONEVM-916)

This adds the implementation of `lp.Process()`, and also initializes LogPoller with a client that gets passed along to initialize the EncodedLogCollector.

In order, the main 3 tasks performed by `lp.Process` are:
1. Filter out any logs not matching registered filters
2. Decode each matching log, and extract the subkeys
3. Save matching logs along with their subkey values to the db


pkg/solana/chain.go
```diff
@@ -304,10 +308,14 @@ func newChain(id string, cfg *config.TOMLConfig, ks core.Keystore, lggr logger.L
 			return result.Signature(), result.Error()
 		}
 
-		tc = internal.NewLoader[client.ReaderWriter](func() (client.ReaderWriter, error) { return ch.multiNode.SelectRPC() })
-		bc = internal.NewLoader[monitor.BalanceClient](func() (monitor.BalanceClient, error) { return ch.multiNode.SelectRPC() })
+		// TODO: Can we just remove these? They nullify the lazy loaders initialized earlier, don't they?
+		//lc = internal.NewLoader[client.Reader](func() (client.Reader, error) { return ch.multiNode.SelectRPC() })
```
dhaidashenko: @DylanTinianov do you recall reason to add this?

pkg/solana/logpoller/filters.go
```diff
@@ -100,6 +113,15 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
 
 	filter.ID = filterID
 
+	idl := codec.IDL{
+		Events: []codec.IdlEvent{filter.EventIdl.IdlEvent},
+		Types:  filter.EventIdl.IdlTypeDefSlice,
+	}
+	filter.decoder, err = codec.NewIDLEventCodec(idl, binary.LittleEndian())
```
dhaidashenko: we should do it before Insert to ensure that db always has valid filters

pkg/solana/logpoller/filters.go
```diff
@@ -118,6 +140,21 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
...
 	}
+
+	programID := filter.Address.ToSolana().String()
+	if _, ok = fl.knownPrograms[programID]; !ok {
+		fl.knownPrograms[programID] = 1
+	} else {
+		fl.knownPrograms[programID]++
+	}
+
+	discriminatorHead := filter.Discriminator()[:10]
```
dhaidashenko, line 154: Are we sure that Distriminator has at least 10 bytes? Should we add validation?

pkg/solana/logpoller/filters.go
```diff
@@ -118,6 +140,21 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
 	if !filter.IsBackfilled {
 		fl.filtersToBackfill[filter.ID] = struct{}{}
 	}
+
+	programID := filter.Address.ToSolana().String()
+	if _, ok = fl.knownPrograms[programID]; !ok {
```
dhaidashenko: No need for if else statement we can rely on default values and just increment it

pkg/solana/logpoller/filters.go
```diff
@@ -180,6 +217,26 @@ func (fl *filters) removeFilterFromIndexes(filter Filter) {
...
+	if refcount, ok := fl.knownPrograms[programID]; ok {
+		refcount--
+		if refcount > 0 {
+			fl.knownPrograms[programID] = refcount
+		} else {
+			delete(fl.knownPrograms, programID)
+		}
+	}
+
+	discriminatorHead := filter.Discriminator()[:10]
```
dhaidashenko, line 231: Same here, we should add validation to avoid panic

pkg/solana/logpoller/filters.go
```diff
@@ -210,6 +267,38 @@ func (fl *filters) MatchingFilters(addr PublicKey, eventSignature EventSignature
...
+	if _, ok := fl.knownPrograms[event.Program]; !ok {
+		return nil
+	}
+
+	// The first 64-bits of the event data is the event sig. Because it's base64 encoded, this corresponds to
+	// the first 10 characters plus 4 bits of the 11th character. We can quickly rule it out as not matching any known
+	// discriminators if the first 10 characters don't match. If it passes that initial test, we base64-decode the
+	// first 11 characters, and use the first 8 bytes of that as the event sig to call MatchingFilters. The address
+	// also needs to be base58-decoded to pass to MatchingFilters
+	if _, ok := fl.knownDiscriminators[event.Data[:10]]; !ok {
```
dhaidashenko: Need validation that data is large enough.

pkg/solana/logpoller/filters.go
```diff
@@ -210,6 +267,38 @@ func (fl *filters) MatchingFilters(addr PublicKey, eventSignature EventSignature
...
+		return nil
+	}
+	decoded, err := base64.StdEncoding.DecodeString(event.Data[:11])
+	if err != nil {
+		fl.lggr.Errorw("failed to decode event data", "EventProgram", event)
+		return nil
+	}
+	eventSig := EventSignature(decoded[:8])
+
+	return fl.MatchingFilters(PublicKey(addr), eventSig)
```
dhaidashenko: There is no need for MatchingFilters to be public

pkg/solana/logpoller/filters.go
```diff
@@ -32,6 +40,11 @@ func newFilters(lggr logger.SugaredLogger, orm ORM) *filters {
 	}
 }
 
+func (fl *filters) IncrementSeqNums(filterID int64) int64 {
+	fl.seqNums[filterID]++
```
dhaidashenko, line 51: It might be nice to ensure that filters were loaded

pkg/solana/logpoller/filters.go
```diff
@@ -180,6 +217,26 @@ func (fl *filters) removeFilterFromIndexes(filter Filter) {
 	if len(filtersForAddress) == 0 {
 		delete(fl.filtersByAddress, filter.Address)
 	}
+
+	programID := filter.Address.ToSolana().String()
```
dhaidashenko, line 221: we should also delete sequence counter

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+	}
+
+	var logs []Log
+	for filter := range matchingFilters {
+		log := Log{
+			FilterID:       filter.ID,
+			ChainID:        lp.orm.ChainID(),
+			LogIndex:       makeLogIndex(blockData.TransactionIndex, blockData.TransactionLogIndex),
+			BlockHash:      Hash(blockData.BlockHash),
+			BlockNumber:    int64(blockData.BlockHeight),
```
dhaidashenko: Should we use the slot number instead? BlockHeight defines how many nonempty slots are behind this block. And it's not used by RPC API or explorers, so does not provide any debugging value

pkg/solana/logpoller/job.go
```diff
@@ -114,12 +115,18 @@ func (j *getTransactionsFromBlockJob) Run(ctx context.Context) error {
...
+	detail.blockHeight = *block.BlockHeight
+
+	if block.BlockTime == nil {
+		return fmt.Errorf("received block %d from rpc with missing block time", block.BlockHeight)
+	}
+	detail.blockTime = *block.BlockTime
 
 	if len(block.Transactions) != len(blockSigsOnly.Signatures) {
-		return fmt.Errorf("block %d has %d transactions but %d signatures", j.slotNumber, len(block.Transactions), len(blockSigsOnly.Signatures))
+		return fmt.Errorf("block %d has %d transactions but %d signatures", block.BlockHeight, len(block.Transactions), len(blockSigsOnly.Signatures))
```
dhaidashenko: We should use slot number. Block Height does not allow us to lookup block in the explorer or fetch it via API.

pkg/solana/logpoller/loader.go
```diff
@@ -170,7 +170,7 @@ func (c *EncodedLogCollector) runSlotPolling(ctx context.Context) {
 			ctxB, cancel := context.WithTimeout(ctx, c.rpcTimeLimit)
 
 			// not to be run as a job, but as a blocking call
-			result, err := c.client.GetLatestBlockhash(ctxB, rpc.CommitmentFinalized)
+			result, err := c.client.LatestBlockhash(ctxB)
```
dhaidashenko, line 173: LatestBlockhash return block hash with a configured commitment. Default value is `CommitmentConfirmed` it might be better to control it on LP level as TXM or other components might have different requirements

pkg/solana/logpoller/log_poller.go
```diff
@@ -2,40 +2,64 @@ package logpoller
...
 	InsertFilter(ctx context.Context, filter Filter) (id int64, err error)
 	SelectFilters(ctx context.Context) ([]Filter, error)
 	DeleteFilters(ctx context.Context, filters map[int64]Filter) error
 	MarkFilterDeleted(ctx context.Context, id int64) (err error)
 	MarkFilterBackfilled(ctx context.Context, id int64) (err error)
+	InsertLogs(context.Context, []Log) (err error)
+	SelectSeqNums(ctx context.Context) (map[int64]int64, error)
+}
+
+type ILogPoller interface {
```
dhaidashenko: Why do we need to define interface?

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
 
 func (lp *LogPoller) start(context.Context) error {
+	if lp.typeProvider == nil {
+		return ErrMissingEventTypeProvider
+	}
+	cl, err := lp.client.Get()
+	if err != nil {
+		return err
+	}
+	lp.collector = NewEncodedLogCollector(cl, lp, lp.lggr)
```
dhaidashenko: Seems like we need to implement proper client that wrapps multinode or use lazy-loader instead. Passing raw client directly to the loader drop MultiNode functionality.

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+
 	lp.eng.Go(lp.run)
 	lp.eng.Go(lp.backgroundWorkerRun)
 	return nil
 }
 
+func makeLogIndex(txIndex int, txLogIndex uint) int64 {
+	if txIndex < 0 || txIndex > math.MaxUint32 || txLogIndex > math.MaxUint32 {
```
dhaidashenko: nit: if txIndex is equal to MaxUint32, we'll have an overflow. We should check that `txIndex > math.MaxInt32`

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+		panic(fmt.Sprintf("txIndex or txLogIndex out of range: txIndex=%d, txLogIndex=%d", txIndex, txLogIndex))
```
dhaidashenko: Does it make sense to kill the whole CL Node in case of the issues in Solana?

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+	}
+	return int64(math.MaxUint32*uint32(txIndex) + uint32(txLogIndex))
```
dhaidashenko: This function produces same results for (0, 1), (2,1) https://go.dev/play/p/z7vPLU9kBQT

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+
+// Process - process stream of events coming from log ingester
+func (lp *LogPoller) Process(programEvent ProgramEvent) (err error) {
+	ctx, cancel := utils.ContextFromChan(lp.eng.StopChan)
+	defer cancel()
+
+	blockData := programEvent.BlockData
+
+	matchingFilters := lp.filters.MatchingFiltersForEncodedEvent(programEvent)
+	if matchingFilters == nil {
```
dhaidashenko, line 91: nit: it should be safe to iterate through nil seq

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+
+		log.SequenceNum = lp.filters.IncrementSeqNums(filter.ID)
+
+		expiresAt := time.Now().Add(filter.Retention).UTC()
+		log.ExpiresAt = &expiresAt
+
+		logs = append(logs, log)
+	}
+
+	err = lp.orm.InsertLogs(ctx, logs)
```
dhaidashenko, line 156: nit: we might want to abort insert if there is no logs to skip db transaction creation

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+		}
+
+		log.Data, err = base64.StdEncoding.DecodeString(programEvent.Data)
+		if err != nil {
+			return err
+		}
+
+		log.SubkeyValues = make(pq.ByteaArray, 0, len(filter.SubkeyPaths))
+		for _, path := range filter.SubkeyPaths {
+			subKeyVal, err2 := filter.DecodeSubKey(ctx, log.Data, path)
```
dhaidashenko: err2, err3 can produce diificult to find bugs, when we return err instead of err2 or 3. 
Should we also wrap them to add more context on what happened and for which log?

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+			}
+			err = log.SubkeyValues.Scan(indexedVal)
+			if err != nil {
+				return err
+			}
+		}
+
+		log.SequenceNum = lp.filters.IncrementSeqNums(filter.ID)
+
+		expiresAt := time.Now().Add(filter.Retention).UTC()
```
dhaidashenko: Should not set expiresAt if filter.Retention is 0

pkg/solana/logpoller/orm.go
```diff
@@ -200,3 +204,13 @@ func (o *DSORM) SelectLogs(ctx context.Context, start, end int64, address Public
 	}
 	return logs, nil
 }
+
+func (o *DSORM) SelectSeqNums(ctx context.Context) (map[int64]int64, error) {
+	seqNums := make(map[int64]int64)
+	query := "SELECT id, MAX(sequence_num) FROM solana.logs WHERE chain_id=%s GROUP BY id"
```
dhaidashenko: Should be filter id, otherwise we'll load whole logs table

pkg/solana/logpoller/types.go
```diff
@@ -76,42 +82,131 @@ func (p SubkeyPaths) Value() (driver.Value, error) {
...
-}
+// IndexedValue represents a value which can be written to, read from, or compared to an indexed BYTEA
+// postgres field. Maps, structs, and slices or arrays (of anything but byte) are not supported. For signed
+// or unsigned integer types, strings, or byte arrays, the SQL operators <, =, & > should work in the expected
+// way.
+type IndexedValue []byte
 
-const EventSignatureLength = 8
+func (v *IndexedValue) FromInt64(i int64) {
+	v.FromUint64(uint64(i + math.MaxInt64))
```
dhaidashenko: Could you add comment to explain why we need to add MaxInt64?

pkg/solana/logpoller/utils/anchor.go
```diff
@@ -0,0 +1,290 @@
...
+	"time"
+
+	bin "github.com/gagliardetto/binary"
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/stretchr/testify/require"
+)
+
+var ZeroAddress = [32]byte{}
```
dhaidashenko: Most of the methods in this file are not used. Do we need them?

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
 		Start: lp.start,
 	}.NewServiceEngine(lggr)
 	lp.lggr = lp.eng.SugaredLogger
+
 	return lp
 }
 
 func (lp *LogPoller) start(context.Context) error {
+	if lp.typeProvider == nil {
```
dhaidashenko: Type provider is always nil

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +68,94 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
 
+func makeLogIndex(txIndex int, txLogIndex uint) int64 {
+	if txIndex < 0 || txIndex > math.MaxUint32 || txLogIndex > math.MaxUint32 {
+		panic(fmt.Sprintf("txIndex or txLogIndex out of range: txIndex=%d, txLogIndex=%d", txIndex, txLogIndex))
+	}
+	return int64(math.MaxUint32*uint32(txIndex) + uint32(txLogIndex))
+}
+
+// Process - process stream of events coming from log ingester
+func (lp *LogPoller) Process(programEvent ProgramEvent) (err error) {
```
dhaidashenko: Loader does not provide a retry mechanic for Process. If Process fails - we'll skip logs

pkg/solana/logpoller/log_poller.go
reductionista: Yeah, I replaced `lp.typeProvider` just last night with a `decoder` object attached to each filter, just forgot to remove it.

pkg/solana/logpoller/orm.go
reductionista: Good catch!

pkg/solana/logpoller/log_poller.go
dhaidashenko: I've created a wrapped client in [this PR](https://github.com/smartcontractkit/chainlink-solana/pull/1006), so that we can seamlessly interact with multinode client or legacy lazy-loaded client.

pkg/solana/logpoller/log_poller.go
reductionista: I thought so, but for some reason it panic's if it's nil and I try to iterate over it

pkg/solana/logpoller/log_poller.go
reductionista: Oh, I thought block number and block height were the same (counting only actual blocks) while slot number counts the empty slots that don't include any block as well.  Is block number just another name for slot number then?

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +66,93 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+			LogIndex:       makeLogIndex(blockData.TransactionIndex, blockData.TransactionLogIndex),
+			BlockHash:      Hash(blockData.BlockHash),
+			BlockNumber:    int64(blockData.BlockHeight),
+			BlockTimestamp: blockData.BlockTime.Time().UTC(),
+			Address:        filter.Address,
+			EventSig:       filter.EventSig,
+			TxHash:         Signature(blockData.TransactionHash),
+		}
+
+		log.Data, err = base64.StdEncoding.DecodeString(programEvent.Data)
```
dhaidashenko: errors related to decoding should be logged as critical as LP won't be able to ever process the msg. 
We might also want just to abandon this log instead of constantly retrying, as we may be able to process other msgs.

pkg/solana/logpoller/log_poller.go
reductionista: I think we'll at least need it so we can generate a mock LogPoller that chainreader can use for their unit tests. But it also helps clearly define what the LogPoller API is. Eventually, once we standardize the API across chains, this should get moved into chainlink-common but for now the EVM and Solana one don't line up perfectly.

That reminds me, I wanted to name it `LogPoller` and rename the struct to `logPoller` so it's more similar to what we have in EVM. But at the time, I was working on top of several branches of open PR's, and didn't want to introduce any additional merge conflicts by renaming the struct everywhere.

I'm open to taking it out for now, but I have a feeling we'll have to put it back in at some point, either for the mock or other reasons. What do you think?

pkg/solana/logpoller/filters.go
reductionista: The `utils.Discriminator` explicitly returns [8]byte, and the `filter.Discriminator` wrapper around that just calls `base64.EncodeToString` on that which will always return []byte backed by [12]byte array. 

I can make this a bit more explicit by changing the return type of `filter.Discriminator` to `[12]byte`, since we have to generate a new slice of it anyway (at least in this case).

pkg/solana/chain.go
DylanTinianov: We must still use these internal loaders, since the above lazy loaders will not properly use MultiNode functionality as they hold onto a single RPC rather than using MultiNode selection on each `getClient()` call.

Looking at this now, we can probably remove the following from ch.getClient() since MultiNode should just use the internal loaders and not lazy loaders:

`	if c.cfg.MultiNode.Enabled() {
		return c.multiNode.SelectRPC()
	}
`

pkg/solana/chain.go
```diff
@@ -90,6 +92,7 @@ type chain struct {
 	services.StateMachine
 	id             string
 	cfg            *config.TOMLConfig
+	lp             logpoller.ILogPoller
```
EasterTheBunny: Why use an interface here when we are always creating an instance of the solana log poller? A function returning an interface is very much a Go anti-pattern and should be avoided unless absolutely necessary.

In this case, we really should store and return a `*logpoller.LogPoller` instead of an interface. If we need to change it in the future, it is easy, but I think it is best to stay with convention before making exceptions.

pkg/solana/chain.go
```diff
@@ -90,6 +92,7 @@ type chain struct {
 	services.StateMachine
```
EasterTheBunny, line 99: Using the following combined as service state helps clean up boilerplate:

```
	services.Service
	engine *services.Engine
```

And in the init:

```
	p.Service, p.engine = services.Config{
		Name: "LogPollerService",
		NewSubServices: func(lggr logger.Logger) []services.Service {
			p.loader = NewEncodedLogCollector(client, p, lggr)

			return []services.Service{p.loader}
		},
		Start: p.start,
		Close: p.close,
	}.NewServiceEngine(lggr)
	p.lggr = p.engine.SugaredLogger
```

pkg/solana/logpoller/job.go
EasterTheBunny: +1

pkg/solana/logpoller/log_data_parser.go
```diff
@@ -78,8 +80,9 @@ func parseProgramLogs(logs []string) []ProgramOutput {
 
 			if len(dataMatches) > 1 {
 				instLogs[lastLogIdx].Events = append(instLogs[lastLogIdx].Events, ProgramEvent{
-					Prefix: prefixBuilder(depth),
-					Data:   dataMatches[1],
+					Program: instLogs[lastLogIdx].Program,
+					Prefix:  prefixBuilder(depth),
```
EasterTheBunny: If `Prefix` isn't needed/used, we can get rid of it. The idea is that it shows overall log nesting.

pkg/solana/logpoller/log_poller.go
EasterTheBunny: Creating unused mocks just adds unnecessary code and brings our code coverage down with no value added. Ideally, mocks should be small and isolated to ONLY what a component needs as a dependency. Since the `LogPoller` doesn't use this interface as a dependency, it shouldn't be here.

pkg/solana/logpoller/log_poller.go
```diff
@@ -2,40 +2,62 @@ package logpoller
...
+
+type ILogPoller interface {
+	Start(context.Context) error
+	Close() error
+	RegisterFilter(ctx context.Context, filter Filter) error
+	UnregisterFilter(ctx context.Context, name string) error
+	Process(programEvent ProgramEvent) error
 }
 
 type LogPoller struct {
```
EasterTheBunny: This really should be something like `Service` so that imports into other packages read as `logpoller.Service`. Using `logpoller.LogPoller` is repetative. I understand it is in an effort to be more like the `EVM` implementation, but we shouldn't copy something that is wrong for the sake of being consistently wrong.

pkg/solana/logpoller/log_poller.go
```diff
@@ -2,40 +2,62 @@ package logpoller
...
+	lggr      logger.SugaredLogger
+	orm       ORM
+	client    internal.Loader[client.Reader]
+	collector *EncodedLogCollector
 
 	filters *filters
 }
 
-func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
+func New(lggr logger.SugaredLogger, orm ORM, cl internal.Loader[client.Reader]) ILogPoller {
```
EasterTheBunny: Using `New` here is great and helps provide a better coding experience. You can use `logpoller.New` instead of `logpoller.NewLogPoller` which would be repetative. However, we really should not be returning an interface here. Instead we should return a pointer type, which is Go convention.

pkg/solana/logpoller/log_poller.go
EasterTheBunny: A retry is built into the loader. If an error is returned here, the loader should retry using exponential backoff.

pkg/solana/logpoller/utils/anchor.go
```diff
@@ -0,0 +1,41 @@
+package utils
```
EasterTheBunny: Let's avoid a `utils` package. Maybe call it `discriminator.go` and leave it in the `logpoller` package for now?

EasterTheBunny (changes requested):

pkg/solana/logpoller/filters.go
reductionista: I changed this, but ended up having to change it back due to other refactoring.  It's stored in `fl.decoders[ID]` now, but needs to get the ID from inserting the db row first.

Unfortunately, having the decoder as part of the `filter` object broke all of the tests which depend on `assert.Equal(filter, expectedFilter)`. I tried fixing that in various ways, but in the end decided it was cleaner to just have the decoders in a separate index instead of a part of the filter.

The order shouldn't matter, since the filters mutex is held throughout this process.

pkg/solana/logpoller/utils/anchor.go
reductionista: Yeah, originally the plan was to import this file from the private repo once it's made public, so I copied it over for now. But then @dhaidashenko pointed out this is the only method actually used, so I got rid of the others. I'll just move it into the main package.

pkg/solana/logpoller/log_data_parser.go
reductionista: Good point, as far as I know we don't need it

pkg/solana/chain.go
reductionista: Ah, I see... the lazy loader would only call `multiNode.SelectRPC()` once, then get stuck on whichever rpc client is returned? Makes sense.

dhaidashenko: Duplicating here my comment from the MultiClient PR. 
We can't remove the `if c.cfg.MultiNode.Enabled() { return c.multiNode.SelectRPC() }` logic from chain's getClient method as getClient is used in other methods of the chain type.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Your understanding is correct. But BlockHeight is not useful for debugging as we can't use it in explorer or in RPC API. All the calls use SlotNumber.

pkg/solana/logpoller/filters.go
```diff
@@ -2,36 +2,53 @@ package logpoller
...
-	loadedFilters     atomic.Bool
+	filtersByID         map[int64]*Filter
+	filtersByName       map[string]int64
+	filtersByAddress    map[PublicKey]map[EventSignature]map[int64]struct{}
+	filtersToBackfill   map[int64]struct{}
+	filtersToDelete     map[int64]Filter
+	filtersMutex        sync.RWMutex
+	loadedFilters       atomic.Bool
+	knownPrograms       map[string]uint // fast lookup to see if a base58-encoded ProgramID matches any registered filters
+	knownDiscriminators map[string]uint // fast lookup by first 10 characters (60-bits) of a base64-encoded discriminator
```
ilija42, line 34: maybe bool instead of uint

pkg/solana/logpoller/filters.go
```diff
@@ -210,6 +271,45 @@ func (fl *filters) MatchingFilters(addr PublicKey, eventSignature EventSignature
...
 
+// MatchchingFiltersForEncodedEvent - similar to MatchingFilters but accepts a raw encoded event. Under normal operation,
+// this will be called on every new event that happens on the blockchain, so it's important it returns immediately if it
+// doesn't match any registered filters.
+func (fl *filters) MatchingFiltersForEncodedEvent(event ProgramEvent) iter.Seq[Filter] {
+	if _, ok := fl.knownPrograms[event.Program]; !ok {
+		return nil
+	}
+
+	if len(event.Data) < 12 {
```
ilija42, line 278: why are we validating this, a small comment would be helpful

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +66,96 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+				return err2
+			}
+			indexedVal, err3 := NewIndexedValue(subKeyVal)
+			if err3 != nil {
+				return err3
+			}
+			log.SubkeyValues = append(log.SubkeyValues, indexedVal)
+		}
+
+		log.SequenceNum = lp.filters.IncrementSeqNums(filter.ID)
```
ilija42: Just want to double check if this is thread safe. logs are processed in parallel, is it possible to start processing a new batch of logs with the same filterID before the previous batch is done processing

pkg/solana/logpoller/log_poller.go
```diff
@@ -44,15 +66,96 @@ func New(lggr logger.SugaredLogger, orm ORM) *LogPoller {
...
+			Address:        filter.Address,
+			EventSig:       filter.EventSig,
+			TxHash:         Signature(blockData.TransactionHash),
+		}
+
+		eventData, err := base64.StdEncoding.DecodeString(programEvent.Data)
+		if err != nil {
+			return err
+		}
+		if len(eventData) < 8 {
```
ilija42, line 123: don't we already check this above in `MatchingFiltersForEncodedEvent`

ilija42 (commented):
Also made improvements on how codec is being used in Log Poller https://github.com/smartcontractkit/chainlink-solana/pull/1014

pkg/solana/logpoller/filters.go
reductionista: It was bool originally, had to make it a uint so it can reference count them, knowing when it's safe to delete a discriminator (once the last filter with that event sig is removed).

pkg/solana/logpoller/log_poller.go
reductionista: It's subtle, but yes this is thread safe.

I got confused by this too, and just tried to add a lock inside `filters.IncrementSeqNums` an hour ago due to the same worry. It resulted in a deadlock, because the mutex is already held here. The function it's locked in is the iterator returned by `fl.matchingFilters`, it gets unlocked and then locked again each time it moves to the next filter in the for loop.

I'll add some comments making this more clear, as it's not at all obvious.

pkg/solana/logpoller/log_poller.go
reductionista: Yes, it's a redundant check which is why I made it an `AssumptionViolation` rather than an ordinary error associated with the blockchain. If the blockchain emits an event less than 8 bytes it should just move on and ignore it. But if it already matched an 8 byte discriminator and now somehow  it's less than 8 bytes, then something in our code must have gone wrong.  I added this in response to @dhaidashenko 's comment asking for more validation, which reminded me that a panic will kill the whole node instead of just reporting a critical error. It could only happen here if someone modifies the code to introduce a bug, but I figured it doesn't hurt to protect against that just in case.

pkg/solana/logpoller/loader.go
reductionista: Hmm... does the txm & logpoller have to share the same multinode client? Or could it make sense to have one for each? I'd prefer to just configure it once for finalized up front rather than having to keep setting finalized every time. Forgetting to set it in one place could easily cause an issue we don't notice for a while.

reductionista: Also, I would think that using unfinalized blocks would lead to similar issues with the tx mgr. But I guess I should check with the ChainWriter team on that.

pkg/solana/logpoller/filters.go
reductionista: I don't think this is the right place for that check, because its only purpose is for `lp.Process` to be able to increment the sequence number each time a new log is processed.

I should probably add a check to `lp.Process`, since we shouldn't be processing logs yet if filters are not loaded. The main loop of `Process` is iterating through the sequence returned by `matchingFilters()`, and I notice this already logs a critical error if it's called before filters are loaded:
```
Invariant violation: expected filters to be loaded before call to matchingFilters
```
But checking prior to this call is better, since then it can return an error.

Looking over this code reminded me, I still need to add a call to `lp.collector.Start()` just after `lp.loadFilters()` returns successfully

reductionista: Oh right, nevermind I forgot there is no seq_num field in filter table. No need for the second one :-)

pkg/solana/logpoller/filters.go
reductionista: I forgot that the return type is `string` not `[]byte`. So instead of changing the return type, I added an AssumptionViolation check that panics if the base64 encoder doesn't encode the 8 bytes passed to it as a 12-byte base64 string. I think the panic makes sense here, because the only way this could potentially happen is during development, if we're updating to a new version of `go` and they've broken backwards compatibility in `encoder/base64` or introduced a bug.  The panic will let us know immediately that we need to roll back or update our code accordingly.

pkg/solana/chain.go
reductionista: Is this comment in reference to `Chain` in this file, or to `LogPoller` in the logpoller file? The example you give looks like LogPoller but you left it on the `Chain` interface.

EasterTheBunny: Ahh my bad. It references `LogPoller` and may not be relevant. It looks like you did it this way so ignore my comment.

pkg/solana/logpoller/discriminator.go
```diff
@@ -0,0 +1,14 @@
+package logpoller
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+const DiscriminatorLength = 8
+
+func Discriminator(namespace, name string) [DiscriminatorLength]byte {
```
EasterTheBunny, line 10: @ilija42 has a PR that also adds a discriminator: [here](https://github.com/smartcontractkit/chainlink-solana/pull/1014)

EasterTheBunny (dismissed):

pkg/solana/logpoller/discriminator.go
reductionista: Yeah, that PR will go in after this one... this will use his discriminator once that gets merged.

pkg/solana/logpoller/filters.go
```diff
@@ -310,7 +411,86 @@ func (fl *filters) LoadFilters(ctx context.Context) error {
...
+		return fmt.Errorf("failed to select sequence numbers from db: %w", err)
+	}
+
 	fl.loadedFilters.Store(true)
 
 	return nil
 }
+
+func (fl *filters) DecodeSubKey(ctx context.Context, raw []byte, ID int64, subKeyPath []string) (any, error) {
+	filter, ok := fl.filtersByID[ID]
```
dhaidashenko, line 436: nit: we might want to mark this method as unsafe to signal that it should only be called from locked context of matching filters

pkg/solana/logpoller/filters.go
```diff
@@ -210,6 +266,49 @@ func (fl *filters) MatchingFilters(addr PublicKey, eventSignature EventSignature
 	}
 }
 
+// MatchchingFiltersForEncodedEvent - similar to MatchingFilters but accepts a raw encoded event. Under normal operation,
+// this will be called on every new event that happens on the blockchain, so it's important it returns immediately if it
+// doesn't match any registered filters.
+func (fl *filters) MatchingFiltersForEncodedEvent(event ProgramEvent) iter.Seq[Filter] {
+	if _, ok := fl.knownPrograms[event.Program]; !ok {
```
dhaidashenko: concurrent read & write to the map leads to panic

pkg/solana/logpoller/types.go
```diff
@@ -76,42 +82,127 @@ func (p SubkeyPaths) Value() (driver.Value, error) {
...
+		v.FromUint64(math.Float64bits(f) + math.MaxInt64 + 1)
+		return
+	}
+	v.FromUint64(math.MaxInt64 + 1 - math.Float64bits(f))
 }
 
-// Value implements valuer for database/sql.
-func (s EventSignature) Value() (driver.Value, error) {
-	return s[:], nil
+func NewIndexedValue(typedVal any) (iVal IndexedValue, err error) {
```
dhaidashenko, line 177: Nice work! I thought we'd have to build something super complex to support `<` `>` operators.

pkg/solana/logpoller/filters.go
reductionista: Oh yeah, I should add that to this as well as `IncrementSeqNums`

pkg/solana/logpoller/filters.go
reductionista: Good catch!

dhaidashenko (approved):

EasterTheBunny (approved):

ilija42 (approved):

aalu1418 (approved):

PR #1003, opened 2025-01-07
Enable ChainReader to read PDA account state

amit-momin:
Enable the ChainReader to read PDA account state
- Added a `PDADefintion` field to the read definition to allow us to configure the fields needed for a PDA account read
- `PDADefintion` includes a `Prefix` and a `Seeds` field
  - `Prefix` is a static string that is converted to bytes and prepended to the list of seeds used for PDA calculation
  - `Seeds` is a list of seeds that are expected to be provided through `params`
- Updated `accountReadBinding`'s `GetAddress` method to intake `params` to use for seeds when calculating the PDA
- Added a new method to `accountReadBinding` to encode and build the seeds list


pkg/solana/chainreader/chain_reader.go
```diff
@@ -299,10 +295,18 @@ func (s *SolanaChainReaderService) addAccountRead(namespace string, genericName
 
 	s.lookup.addReadNameForContract(namespace, genericName)
 
-	s.bindings.AddReadBinding(namespace, genericName, newAccountReadBinding(
-		namespace,
-		genericName,
-	))
+	var reader readBinding
+	// Create PDA read binding if Seeds config is non-nil
+	// Note: Empty seeds list is a valid configuration for a PDA so a length check is intentionally skipped
```
silaslenihan: Would an empty seeds list just be a normal account read?

pkg/solana/chainreader/pda_read_binding.go
```diff
@@ -0,0 +1,97 @@
...
+	return b.codec.Decode(ctx, bts, outVal, codec.WrapItemType(false, b.namespace, b.genericName, codec.ChainConfigTypeAccountDef))
+}
+
+func (b *pdaReadBinding) buildSeedsSlice(params any) ([][]byte, error) {
+	if b.seeds == nil {
+		return [][]byte{}, nil
+	}
+
+	seedByteArray := make([][]byte, 0, len(b.seeds))
+	for _, seed := range b.seeds {
```
silaslenihan: thanks for reusing some of the chainwriter seeds functionality! makes it a lot easier to understand both

silaslenihan (commented):
This all looks great to me! Really readable and straightforward. Not sure if you're done adding functionality for CCIP but this should unblock the interface tests

pkg/solana/chainreader/chain_reader.go
amit-momin: I think `FindProgramAddress` still outputs a different programID even with empty seeds since it's still applying the bump seed. I think every PDA in the on-chain code uses some sort of seed though. So guess it wouldn't hurt to also validate empty seeds for our current use case but I wanted to keep this open in case something changes in the future.

pkg/solana/codec/solana.go
```diff
@@ -388,6 +388,29 @@ func asVec(parentTypeName string, idlVec *IdlTypeVec, refs *codecRefs) (commonen
...
 
+func asSeeds(seeds []SeedDefinition) (commonencodings.TypeCodec, error) {
+	named := make([]commonencodings.NamedTypeCodec, len(seeds))
+
+	for idx, field := range seeds {
+		var typedCodec commonencodings.TypeCodec
+		switch field.Type {
+		case SeedPubKey:
+			typedCodec = NewPublicKey()
+		case SeedUint64:
```
ilija42: why can't we reuse asStruct here so that every type is handled?

pkg/solana/codec/solana.go
```diff
@@ -388,6 +388,29 @@ func asVec(parentTypeName string, idlVec *IdlTypeVec, refs *codecRefs) (commonen
...
 }
 
+func asSeeds(seeds []SeedDefinition) (commonencodings.TypeCodec, error) {
+	named := make([]commonencodings.NamedTypeCodec, len(seeds))
+
+	for idx, field := range seeds {
+		var typedCodec commonencodings.TypeCodec
+		switch field.Type {
+		case SeedPubKey:
+			typedCodec = NewPublicKey()
```
ilija42: Why is special handling for public key needed? Why doesn't regular public key encoded by Solana codec work?

pkg/solana/chainreader/chain_reader.go
```diff
@@ -284,25 +301,26 @@ func (s *SolanaChainReaderService) init(namespaces map[string]config.ChainContra
...
-	))
+	var reader readBinding
+	// Create PDA read binding if PDA prefix or seeds configs are populated
+	if len(readDefinition.PDAPrefix) > 0 || len(readDefinition.Seeds) > 0 {
+		if err := s.addSeedsCodecDef(namespace, genericName, readDefinition.Seeds, readDefinition.InputModifications); err != nil {
+			return fmt.Errorf("failed to add codec entry for seed configs: %w", err)
+		}
+		reader = newPdaReadBinding(namespace, genericName, readDefinition.PDAPrefix)
+	} else {
+		if err := s.addCodecDef(true, namespace, genericName, codec.ChainConfigTypeAccountDef, idl, codec.NilIdlTypeDefTy, readDefinition.InputModifications); err != nil {
```
ilija42: What does this do? There are no inputs for non PDA account reads

pkg/solana/codec/solana.go
amit-momin: I was thinking `asStruct` seemed strongly tied to IDL types and since PDA seeds aren't officially defined in the IDL I thought it made sense to keep this separate. But your comments making me rethink that. I think I can use `IDLField` in place of `SeedDefinition` and reuse the existing `IDLType` instead of my custom enum. I'll give this a shot

pkg/solana/codec/solana.go
amit-momin: Sorry this was just a miss on my part. I see [this](https://github.com/smartcontractkit/chainlink-solana/blob/0cd58cca8aa21e3ea168b75a3f53e830720a2d94/pkg/solana/codec/solana.go#L464-L465) now which would get automatically picked up once I address the previous comment.

pkg/solana/chainreader/chain_reader.go
```diff
@@ -249,6 +249,23 @@ func (s *SolanaChainReaderService) addCodecDef(forEncoding bool, namespace, gene
...
+	if err != nil {
+		return fmt.Errorf("failed to create a codec entry for seed definitions %v: %w", seedDefs, err)
+	}
+
+	// Seed codec entry is only used for encoding. Read type is not used by WrapItemType for encoding string.
+	s.parsed.EncoderDefs[codec.WrapItemType(true, namespace, genericName, "")] = seedEntry
+
+	return nil
+}
+
```
ilija42: I think this can go away in favour of just using `addCodecDef()`. You just need to define seeds as a new IDL type and handle it like events, args and accounts are handled here `CreateCodecEntry()` which is a func used by `addCodecDef()`.

pkg/solana/chainreader/chain_reader.go
amit-momin: I think this is unused for normal account reads. For encoding PDAs, we only really need the new `addSeedsCodecDef` method I added. But when I tried to remove this, I started getting some errors. I can debug and see what's going on but since this was already in the existing code I opted to keep it as is

ilija42: my bad actually, I wrote this code lol

ilija42: I thought I left it as a placeholder for PDA to be input, not sure why it throws errors

pkg/solana/codec/anchoridl.go
```diff
@@ -140,6 +140,18 @@ type IdlField struct {
...
+// PDA is a struct that does not correlate to an official IDL type
+// It is needed to encode seeds to calculate the address for PDA account reads
+type PDATypeDef struct {
+	Prefix string    `json:"prefix,omitempty"`
+	Seeds  []PDASeed `json:"seeds,omitempty"`
+}
+
+type PDASeed struct {
+	Name string          `json:"name"`
+	Type IdlTypeAsString `json:"type"`
```
amit-momin, line 152: Is it ok if I use `IdlTypeAsString` here? This would expose Solana specific types to CR config which I could see being something we want to avoid. I wanted to double check before introducing a new human readable type just for seeds.

ilija42: Thats fine, this config is Solana specific anyway

pkg/solana/chainreader/pda_read_binding.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/gagliardetto/solana-go"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codec"
+)
+
+// pdaReadBinding provides calculating PDA addresses with the provided seeds and reading decoded PDA Account data using a defined codec
+type pdaReadBinding struct {
```
ilija42: Do you think we could merge this into account_read_binding.go? They essentially do the same thing, since we’re reading an account in both cases. The only difference is that we need to calculate the PDA address if parameters are provided.

amit-momin: Hmm good point. I think I could. I'll give this a shot

amit-momin: I merged the two in the latest commit

ilija42 (approved):
looks good

EasterTheBunny (approved):
LGTM

aalu1418 (approved):

PR #1006, opened 2025-01-10
Wrapped MultiNode client

dhaidashenko:
Wrapped over MultiNode and lazy-loaded client to ensure components like LogPoller can interact with them without need to handle error returned on RPC selection. 
See [comment](https://github.com/smartcontractkit/chainlink-solana/pull/1002#discussion_r1909185678).


DylanTinianov (dismissed):

pkg/solana/client/multinode_wrapped_client.go
```diff
@@ -0,0 +1,168 @@
...
+
+var _ ReaderWriter = (*MultiClient)(nil)
+
+// MultiClient - wrapper over multiple RPCs, underlying provider can be MultiNode or LazyLoader.
+// Main purpose is to eliminate need for frequent error handling on selection of a client.
+type MultiClient struct {
+	getClient func() (ReaderWriter, error)
+}
+
+func NewMultiNodeWrappedClient(getClient func() (ReaderWriter, error)) *MultiClient {
```
DylanTinianov:
```suggestion
func NewMultiClient(getClient func() (ReaderWriter, error)) *MultiClient {
```

DylanTinianov: Should we maybe rename the file to `multiclient.go` since it's not only wrapping MultiNode?

dhaidashenko: Good point. Addressed

DylanTinianov (dismissed):

pkg/solana/chain.go
```diff
@@ -291,6 +293,9 @@ func newChain(id string, cfg *config.TOMLConfig, ks core.Keystore, lggr logger.L
 
 		ch.multiNode = multiNode
 		ch.txSender = txSender
+		ch.multiClient = client.NewMultiClient(func() (client.ReaderWriter, error) {
+			return ch.multiNode.SelectRPC()
+		})
```
reductionista: Isn't this redundant with the prior initialization?
```
ch.multiClient = client.NewMultiClient(ch.getClient)
```
Either way, if the multinode is enabled the same `SelectRPC()` function will be called, right?
```
// getClient returns a client, randomly selecting one from available and valid nodes
// If multinode is enabled, it will return a client using the multinode selection instead.
func (c *chain) getClient() (client.ReaderWriter, error) {
	if c.cfg.MultiNode.Enabled() {
		return c.multiNode.SelectRPC()
	}
```

reductionista: Actually I just re-read @DylanTinianov's comment on my PR and saw he's proposing to get rid of the MultiNode path from `c.getClient`.  If we do that, then I think this makes sense as-is.

reductionista: We could either do that in this PR, or in my PR

reductionista (dismissed):

pkg/solana/chain.go
dhaidashenko: Removed second initialization of multiClient. 
We can't remove the MultiNode path from `c.getClient` as it's by other methods of the `chain` struct.

DylanTinianov (approved):

reductionista (approved):

jadepark-dev (approved):

jadepark-dev (approved):

PR #1007, opened 2025-01-10
Updated getSeedBytes to handle array seeds

silaslenihan:
### Description
This PR updated the ChainWriter's PDA parsing capabilities to handle the case where a seed resolves to an array. 

This will be necessary for handling CCIP messages which contain multiple token transfers.

### DIFF
--- a/pkg/solana/chainwriter/helpers.go
+++ b/pkg/solana/chainwriter/helpers.go
@@ -84,6 +84,22 @@ func errorWithDebugID(err error, debugID string) error {
 // traversePath recursively traverses the given structure based on the provided path.
 func traversePath(data any, path []string) ([]any, error) {
 	if len(path) == 0 {
+		val := reflect.ValueOf(data)
+
+		// If the final data is a slice or array, flatten it into multiple items,
+		if val.Kind() == reflect.Slice || val.Kind() == reflect.Array {
+			// don't flatten []byte
+			if val.Type().Elem().Kind() == reflect.Uint8 {
+				return []any{val.Interface()}, nil
+			}
+
+			var results []any
+			for i := 0; i < val.Len(); i++ {
+				results = append(results, val.Index(i).Interface())
+			}
+			return results, nil
+		}
+		// Otherwise, return just this one item
 		return []any{data}, nil
 	}
 
@@ -124,9 +140,6 @@ func traversePath(data any, path []string) ([]any, error) {
 		}
 		return traversePath(value.Interface(), path[1:])
 	default:
-		if len(path) == 1 && val.Kind() == reflect.Slice && val.Type().Elem().Kind() == reflect.Uint8 {
-			return []any{val.Interface()}, nil
-		}
 		return nil, errors.New("unexpected type encountered at path: " + path[0])
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/solana/chainwriter/lookups.go
```diff
@@ -210,63 +210,103 @@ func decodeBorshIntoType(data []byte, typ reflect.Type) (interface{}, error) {
 }
 
 // getSeedBytes extracts the seeds for the PDALookups.
-// It handles both AddressSeeds (which are public keys) and ValueSeeds (which are byte arrays from input args).
-func getSeedBytes(ctx context.Context, lookup PDALookups, args any, derivedTableMap map[string]map[string][]*solana.AccountMeta, reader client.Reader) ([][]byte, error) {
-	var seedBytes [][]byte
+// The return type is [][][]byte, where each element of the outer slice is
+// one combination of seeds. This handles the case where one seed can resolve
+// to multiple addresses, multiplying the combinations accordingly.
+func getSeedBytes(
```
amit-momin: nit: With the new functionality, think we could maybe update the name to reflect. Maybe `getSeedBytesCombinations`

pkg/solana/chainwriter/lookups.go
```diff
@@ -210,63 +210,103 @@ func decodeBorshIntoType(data []byte, typ reflect.Type) (interface{}, error) {
...
 				}
 			}
 		}
+
+		// expansions is the list of possible seed bytes for this single seed lookup.
+		// Multiply the existing combinations in allCombinations by each item in expansions.
+		newCombinations := make([][][]byte, 0, len(allCombinations)*len(expansions))
+		for _, existingCombo := range allCombinations {
+			for _, expandedSeed := range expansions {
+				comboCopy := make([][]byte, len(existingCombo))
```
amit-momin: nit: If we know we're adding an element on top of this, should we initialize this with +1 length? Then instead of append below we can just set the new element through the index.

silaslenihan: Good suggestion, fixed!

pkg/solana/chainwriter/helpers.go
```diff
@@ -84,6 +84,22 @@ func errorWithDebugID(err error, debugID string) error {
 // traversePath recursively traverses the given structure based on the provided path.
 func traversePath(data any, path []string) ([]any, error) {
 	if len(path) == 0 {
+		val := reflect.ValueOf(data)
+
+		// If the final data is a slice or array, flatten it into multiple items,
+		if val.Kind() == reflect.Slice || val.Kind() == reflect.Array {
+			// don't flatten []byte
```
prashantkumar1982, line 91: Is it possible for us to infer if a byte array is actually a Solana PublicKey?
I'm thinking, theoretically, a byte can be a seed right?
So a []byte may also technically be a list of seeds sometimes. And other times, it can be a Solana Address

silaslenihan: Hm that's a good point, a single byte could potentially be a seed, and if we want each element in an array of bytes to be a seed, that wouldn't be possible with this current logic. I'm unsure what validation would make it possible to enable this kind of functionality and whether its worth adding complexity for a case we may not even need.

prashantkumar1982: Sounds good, yes it is ok to let things be as it is for now.

prashantkumar1982 (approved):

amit-momin (approved):

jadepark-dev (approved):

PR #1012, opened 2025-01-13
NONEVM-1115 logpoller backfill

dhaidashenko:
Implemented main LogPoller loop. 
Forward-looking block polling was replaced by global backfill to reduce the complexity and need to maintain two ways of fetching logs. 
Draft of a performance test showed ~10x decrease in time needed to catchup from 10,000 slots lag on Solana Mainnet (assuming that CCIP uses a single contract for all interactions and that transaction frequency is similar to the top bridge available on Solana).


pkg/solana/logpoller/log_poller.go
```diff
@@ -19,36 +23,77 @@ type ORM interface {
...
+	lp.processBlocks = lp.processBlocksImpl
+
 	lp.Service, lp.eng = services.Config{
 		Name:  "LogPollerService",
 		Start: lp.start,
+		NewSubServices: func(l logger.Logger) []services.Service {
+			loader := NewEncodedLogCollector(client, lggr)
+			lp.loader = loader
+			return []services.Service{loader}
+		},
```
reductionista, line 84: I avoided doing this because the `Start()` method on `Service` starts all of the subservices before it starts itself. So this will start the EncodedLogCollector before filters are loaded.

It shouldn't be a big problem, since `Process()` will error out until the filters are ready. But this could result in logs getting fetched and piling up in memory while it waits for the filters to be loaded.  I don't feel strongly about it, but lean toward starting things in the right order over having this formally listed as a subservice.

This is true even with the call to `loadFilters()` being at the beginning of `lp.start` where you've got it now.

pkg/solana/logpoller/log_poller.go
```diff
@@ -19,36 +23,77 @@ type ORM interface {
...
+	lp := New(lggr, orm, client)
+	lp.processBlocks = processBlocks
+	return lp
+}
+
+func (lp *LogPoller) start(ctx context.Context) error {
+	err := lp.filters.LoadFilters(ctx)
+	if err != nil {
+		return fmt.Errorf("failed loading filters: %w", err)
+	}
```
reductionista: I think we usually try not to have any I/O blocking calls within Start() functions, since any top-level services started after this will have to wait for it.  (As mentioned above, it will not stop the LogCollector from starting up before the filters are loaded, just other services like tx mgr, or whatever gets loaded after logpoller.)

That said, loading the filters will hopefully be a very quick operation so I think it's probably fine.

pkg/solana/logpoller/log_poller.go
```diff
@@ -67,58 +112,163 @@ func (lp *LogPoller) UnregisterFilter(ctx context.Context, name string) error {
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case block, ok := <-blocks:
+			if !ok {
+				break consumedAllBlocks
+			}
+
+			batch := []Block{block}
+			batch = appendBuffered(blocks, 16, batch)
```
reductionista: Would be good to define this constant in `block_sorter.go` ( BlocksBufferSize=16 ?) so we can reference the same constant in both places.

Also, curious how picked 16... have you compared performance with different buffer sizes yet, or is this just a guess for us to start out with?

pkg/solana/logpoller/job_get_block.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+)
+
+// getBlockJob is a job that fetches blocks with transactions, coverts logs into ProgramEvents and writes them into blocks channel
```
reductionista: Maybe "fetches a block with transactions" would be more clear? I got confused in a couple places about whether this job was fetching a single block or many

also: covert -> converts

pkg/solana/logpoller/job_get_block.go
```diff
@@ -0,0 +1,144 @@
...
+	return j.done
+}
+
+func (j *getBlockJob) Run(ctx context.Context) error {
+	var excludeRewards bool
+	version := client.MaxSupportTransactionVersion
+	block, err := j.client.GetBlockWithOpts(
+		ctx,
+		j.slotNumber,
+		// NOTE: any change to the filtering arguments may affect calculation of txIndex, which could lead to events duplication.
```
reductionista, line 51: Are the filtering arguments this is referring to `MaxSupportedTransactionVersion` and `Rewards`?

I wonder if `MaxSupportedTransactionVersion` might change 🤔 if so, that does seem like something we should be very careful about handling--no action requested just thinking we might ask someone about this or note it somewhere to keep track.

reductionista:
This looks great! A few minor comments, very pleased that we don't have to fetch every block any more.  Eventually I'd like to experiment with just fetching transactions instead of blocks, in cases where that could speed things up, but that seems much harder to do while keeping everything in sync and would add a lot of complexity. This should be plenty fast enough for initial release.

pkg/solana/logpoller/loader.go
```diff
@@ -1,477 +1,163 @@
...
-		case job := <-c.chJobs:
-			if err := c.workers.Do(ctx, job); err != nil {
-				c.lggr.Errorf("failed to add job to queue: %s", err)
-			}
+func (c *EncodedLogCollector) scheduleBlocksFetching(ctx context.Context, slots []uint64) (<-chan Block, error) {
+	blocks := make(chan Block)
+	getBlocksJobs := make([]*getBlockJob, len(slots))
+	for i, slot := range slots {
+		getBlocksJobs[i] = newGetBlockJob(c.client, blocks, c.lggr, slot)
+		err := c.workers.Do(ctx, getBlocksJobs[i])
```
reductionista: I'd recommend renaming this to getBlockJobs instead of getBlocksJobs since it's only fetching a single block per job with GetBlockWithOpts not calling GetBlocks

pkg/solana/logpoller/job_get_block.go
```diff
@@ -0,0 +1,152 @@
...
+	result := Block{
+		SlotNumber: j.slotNumber,
+		BlockHash:  block.Blockhash,
+		Events:     events,
+	}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case j.blocks <- result:
+		close(j.done)
```
EasterTheBunny, line 116: I'm not 100% certain how `Done` is being used for the job, but it might be better to do a 

```
defer func(){close(j.done)}()
```

This would ensure that anything only watching for `Done` to return does so even with an error.

EasterTheBunny: It would probably work the same as the `context.Context` interface

pkg/solana/logpoller/job_get_slots_for_addr.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (f *getSlotsForAddressJob) Run(ctx context.Context) error {
+	isDone, err := f.run(ctx)
+	if err != nil {
+		return err
+	}
+
+	if isDone {
+		close(f.done)
+	}
```
EasterTheBunny, line 58: Is done handled this way because of retries?

pkg/solana/logpoller/log_poller.go
```diff
@@ -27,43 +28,80 @@ type ORM interface {
 	DeleteFilters(ctx context.Context, filters map[int64]Filter) error
 	MarkFilterDeleted(ctx context.Context, id int64) (err error)
 	MarkFilterBackfilled(ctx context.Context, id int64) (err error)
+	GetLatestBlock(ctx context.Context) (int64, error)
```
EasterTheBunny, line 31: Why use the ORM to give the latest block? Weren't we getting that from the RPC?

pkg/solana/logpoller/log_poller.go
```diff
@@ -27,43 +28,80 @@ type ORM interface {
...
 	InsertLogs(context.Context, []Log) (err error)
 	SelectSeqNums(ctx context.Context) (map[int64]int64, error)
 }
 
+type logsLoader interface {
+	BackfillForAddresses(ctx context.Context, addresses []PublicKey, fromSlot, toSlot uint64) (orderedBlocks <-chan Block, cleanUp func(), err error)
```
EasterTheBunny, line 37: We're only loading logs using the `BackfillForAddress` function now?

pkg/solana/logpoller/loader_test.go
```diff
@@ -268,378 +180,10 @@ type slotUnsync struct {
...
-		})
-
-	assert.NoError(t, collector.BackfillForAddress(ctx, pubKey.String(), 42))
-
-	tests.AssertEventually(t, func() bool {
-		return parser.Count() == 6
-	})
-}
-
-func BenchmarkEncodedLogCollector(b *testing.B) {
```
EasterTheBunny, line 388: Why remove the benchmark? It seems there was a major change in how we are collecting block data, but the means to measure efficiency was removed.

pkg/solana/logpoller/loader.go
```diff
@@ -1,477 +1,163 @@
...
-				return err
-			}
+			return nil, fmt.Errorf("could not shedule job to fetch slots for address: %w", err)
 		}
 	}
 
-	return nil
-}
-
-func (c *EncodedLogCollector) start(_ context.Context) error {
```
EasterTheBunny, line 147: I know there was discussion on changing the strategy for reading block data, but I see a lot of stuff removed and no proof that the change was an improvement.

EasterTheBunny (changes requested):
This PR is a big change related to a push from another team. I would really like to see the benchmark brought back to help prove the value of this new approach.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Loader no longer produces any event until we call `BackfillForAddresses`, so `Process` won't be called before filters were loaded

pkg/solana/logpoller/log_poller.go
dhaidashenko: Moved filters loading to the `run`

pkg/solana/logpoller/log_poller.go
dhaidashenko: Created a constant to hold the buffer size.
16 - is just a guess.

pkg/solana/logpoller/job_get_block.go
dhaidashenko: The version change should be okay. If a requested block contains a transaction with a version greater than `MaxSupportedTransactionVersion,` we'll get an error.
At the moment, only `Rewards` filters transactions, so any change to it will affect transaction's index => affect log's index

pkg/solana/logpoller/log_poller.go
dhaidashenko: LogPoller needs to provide a guarantee that it captures all events. If we only get the latest block from RPC, we'll miss some due to restarts or other downtime.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Yes, the main reason for this change is to reduce scope. The goals of this PR were
1. Introduce changes into the loader's Backfill operation to allow it to signal when the backfill is done.
2. Add synchronization between the backfill for a newly registered filter and the main forward-looking loop.
3. Add global backfill to speed up catchup in cases of prolonged downtime (~8 hours). Calculation showed that forward-looking processing needs `8(hours of downtime)*3600*2.5 (blocks per second)*2 (request per block) = 144,000` requests to catch up, even if there are no CCIP events. If an RPC support 20 block req/second, it would take 2 hours to process 8 hours of data.  

This PR implements 1 and 3. Adding 2 would have further increased the scope and complexity of the LogPoller. Given the short timeline, I've dropped the 2 to deliver a working implementation of LP as soon as possible.

pkg/solana/logpoller/loader_test.go
dhaidashenko: It does not seem to work for me on latest develop. I always get `0 events/sec` for some reason. Also, it focuses on forward-looking processing, so fixing it would require significant time. I still have performance testing [ticket](https://smartcontract-it.atlassian.net/browse/NONEVM-748), so memory and CPU usage can be measured there.
```
BenchmarkEncodedLogCollector-10         10626343              5479 ns/op                 0 events/sec            0.9790 rcp_calls/sec        699 B/op          3 allocs/op
``` 

If you are worried about the number of RPC requests during regular operation, here are my calculations. 
In the previous version we performed the following requests:
1.  fetch the latest slot (1 request)
2. fetch blocks available in the range (1 request as we are fetching every second).
3. fetch all blocks in the range (2.5 requests if we are fetching every second).
Total: 4.5 requests.
In this version: 
1. fetch the latest slot (1 request) 
2. for each contract fetch signatures. As CCIP uses a single contract and we can expect less than 1000 CCIP transactions per second => 1 request should be enough
3. fetch all blocks that contain transactions (2.5 requests in the worst case) 
Total: 2-4.5 requests.

pkg/solana/logpoller/loader.go
dhaidashenko: The main goal of this PR is to deliver a functioning LogPoller. We can alwaysys reintroduce forward-looking processing as a followup.

pkg/solana/logpoller/loader_test.go
EasterTheBunny: That's is essentially what I was looking for in the the benchmark. However I think we are sacrificing scale to make this change and I wanted that to be apparent in the benchmark. 

The initial concern was that we were processing too much data per block (and downloading too much data). In this solution we are targeting specific addresses. However, what happens when we scale out the number of contract addresses we need to target?

While I know that we are primarily building for CCIP right now, I don't know that it's the right move to eliminate functionality for the sake of efficiency of a single product. I would like to have seen these improvements as an option if possible because there are legitimate future use cases where the new approach is the less efficient option.

pkg/solana/logpoller/loader.go
EasterTheBunny: I'm not sure delivering a functioning LogPoller required removing functionality though. Yes, we can reintroduce forward-looking processing, but I would expect that would be more difficult than leaving it in.

pkg/solana/logpoller/job_get_block.go
dhaidashenko: j.done channel signals when job was successfully done. It should not be closed if context was cancelled.

pkg/solana/logpoller/job_get_slots_for_addr.go
dhaidashenko: Yes, if there is an error and job will be retried we should not signal that job was done

pkg/solana/logpoller/loader.go
dhaidashenko: If you insist, I can roll back removal, but I'm not sure that leaving unused code is a good idea.

pkg/solana/logpoller/loader_test.go
dhaidashenko: > I don't know that it's the right move to eliminate functionality for the sake of efficiency of a single product.

The functionality was removed to reduce scope of the work. Implementation of synchronisation between LogPoller's backfill operations and forward-looking loop is quite complex. Domino described it in [this ticket](https://smartcontract-it.atlassian.net/browse/NONEVM-1115).

pkg/solana/logpoller/loader.go
EasterTheBunny: Well I'm not going to stonewall this, but I just think the approach of removing something for the sake of efficiency for one product and sacrificing future capability is generally bad practice.

EasterTheBunny (approved):

pkg/solana/logpoller/log_poller.go
reductionista: @EasterTheBunny These are two different things--we need to know what block we left off at before a restart, as well as what the latest block on chain is.  The naming can be confusing since "GetLatestBlock" doesn't indicate which, but I think it's okay because it's a part of the ORM so it's clear which one you're requesting. (Use the ORM to get latest in db, use Client to get latest from chain. Then backfill everything in between.)

reductionista (approved):

jadepark-dev (approved):

PR #1014, opened 2025-01-14
Improve log poller codec usage and optimise log event discriminator matching

ilija42:
### Description

- Fix event discriminator handling, fix codec event tests.
- Implement very optimised discriminator extractor for faster log poller event matching
- Init codec through NewCodec in LP which resolves discriminators automatically on Decode


EasterTheBunny (approved):

EasterTheBunny (dismissed):

EasterTheBunny (dismissed):

pkg/solana/logpoller/filters.go
```diff
@@ -228,7 +230,7 @@ func (fl *filters) removeFilterFromIndexes(filter Filter) {
 		}
 	}
 
-	discriminatorHead := filter.Discriminator()[:10]
```
reductionista, line 231: We must store only the first 10 characters (hence the name `discriminatorHead`). Storing all 12 is not useful and will cause us to miss logs. (more on this in another comment)

pkg/solana/logpoller/filters.go
```diff
@@ -278,6 +280,15 @@ func (fl *filters) MatchingFiltersForEncodedEvent(event ProgramEvent) iter.Seq[F
...
 	}
+
+	discriminator, err := base64.StdEncoding.DecodeString(event.Data[:12])
+	if err != nil {
+		fl.lggr.Errorw("failed to decode event discriminator", "event", event, "err", err)
+		return nil
+	}
+
+	discriminator = discriminator[:8]
+
```
reductionista: Moving this up here would defeat the purpose of having the `knownDiscriminators` lookup, which is to quickly tell whether it's known so we don't have to base64 decode anything 99%+ of the time.

The way it's currently implement is very efficient--it will just check the first 10 characters against the hash table quickly, and if they don't match it can return immediately. Granted, it may not add that much time but there's no need to base64-decode every single discriminator as they come back from the rpc

reductionista (dismissed):

pkg/solana/logpoller/filters.go
```diff
@@ -291,7 +302,7 @@ func (fl *filters) MatchingFiltersForEncodedEvent(event ProgramEvent) iter.Seq[F
 		// discriminators if the first 10 characters don't match. If it passes that initial test, we base64-decode the
 		// first 12 characters, and use the first 8 bytes of that as the event sig to call MatchingFilters. The address
 		// also needs to be base58-decoded to pass to MatchingFilters
-		_, ok = fl.knownDiscriminators[event.Data[:10]]
+		_, ok = fl.knownDiscriminators[base64.StdEncoding.EncodeToString(discriminator)]
```
reductionista: You decode it and then immediately encode it again. Then decode it again later? There's no need to do 3 steps we only need one quick check and then one decoding if it passes.

reductionista: (Also, the comments don't match any more what the code does.)

pkg/solana/logpoller/filters.go
ilija42: If you just check first 10 then its not accurate and you might get bad logs matched which we really don't want for CCIP

ilija42: I optimised it as much as possible now, while still maintaining the comparison of the whole discriminator

reductionista:
> Implement very optimised discriminator extractor for faster log poller event matching

Faster than what? It can't be faster than the existing implementation which just does a simple lookup in `knownDiscriminators` without decoding anything. Is the advantage that this is easier to read or more maintainable?

pkg/solana/codec/discriminator.go
```diff
@@ -69,3 +79,47 @@ func (d Discriminator) Size(_ int) (int, error) {
...
+		d[1] = (c2 << 4) | (c3 >> 2)
+		d[2] = (c3 << 6) | c4
+
+		// next 3 bytes and next 4 characters
+		d = d[3:]
+		s = s[4:]
+	}
+
+	return decodeBuffer[:discriminatorLength]
+}
```
reductionista, line 125: Can you explain the motivation for this? (Wanting to extract the discriminator). What is the benefit?

reductionista: The current implementation just compares the two strings directly. This would change it to quickly decode it and then re-encode it, then compare them. That's great that it's quick, but... what's the point? Am I missing something?

pkg/solana/logpoller/filters.go
```diff
@@ -23,24 +23,26 @@ type filters struct {
...
-	decoders            map[int64]Decoder
+	filtersByID            map[int64]*Filter
+	filtersByName          map[string]int64
+	filtersByAddress       map[PublicKey]map[EventSignature]map[int64]struct{}
+	filtersToBackfill      map[int64]struct{}
+	filtersToDelete        map[int64]Filter
+	filtersMutex           sync.RWMutex
+	loadedFilters          atomic.Bool
+	knownPrograms          map[string]uint // fast lookup to see if a base58-encoded ProgramID matches any registered filters
+	knownDiscriminators    map[string]uint // fast lookup by first 10 characters (60-bits) of a base64-encoded discriminator
```
reductionista:
```suggestion
	knownDiscriminators    map[string]uint // fast lookup based on 8-byte discriminator (casted to string)
```

pkg/solana/logpoller/filters.go
```diff
@@ -228,7 +231,7 @@ func (fl *filters) removeFilterFromIndexes(filter Filter) {
 		}
 	}
 
-	discriminatorHead := filter.Discriminator()[:10]
+	discriminatorHead := filter.DiscriminatorRawBytes()
 	if refcount, ok := fl.knownDiscriminators[discriminatorHead]; ok {
```
reductionista: I only named this var `discriminatorHead` to distinguish it from a full discriminator, since it was just the first 10 characters of 12. Suggest dropping the `Head` now that it's the full discriminator
```suggestion
	discriminator := filter.DiscriminatorRawBytes()
	if refcount, ok := fl.knownDiscriminators[discriminator]; ok {
```

pkg/solana/logpoller/log_poller.go
```diff
@@ -116,24 +116,25 @@ func (lp *Service) Process(programEvent ProgramEvent) (err error) {
...
-			return decodeErr
+		log.Data, err = base64.StdEncoding.DecodeString(programEvent.Data)
+		if err != nil {
+			return err
 		}
-		if len(eventData) < 8 {
-			err = fmt.Errorf("Assumption violation: %w, log.Data=%s", ErrMissingDiscriminator, log.Data)
+
+		// TODO isn't discrimintaor already checked above in MatchingFiltersForEncodedEvent?
+		if len(log.Data) < 8 {
```
reductionista: I added this check to address a PR comment from @dhaidashenko requesting more validation, in case it's not 8 bytes. I added it, but noted that it isn't something that can happen with the existing code--would just protect against someone accidentally introducing a bug later.

I don't care much either way whether we leave it or take it out, but if you want to take it out you should get @dhaidashenko approval.

pkg/solana/logpoller/models.go
```diff
@@ -26,16 +26,9 @@ func (f Filter) MatchSameLogs(other Filter) bool {
...
-//
-// This is the base64 encoding of the [8]byte discriminator returned by utils.Discriminator
-func (f Filter) Discriminator() string {
-	d := Discriminator("event", f.EventName)
-	b64encoded := base64.StdEncoding.EncodeToString(d[:])
-	if len(b64encoded) != 12 {
-		panic(fmt.Sprintf("Assumption Violation: expected encoding/base64 to return 12 character base64-encoding, got %d characters", len(b64encoded)))
-	}
-	return b64encoded
+// DiscriminatorRawBytes returns raw discriminator bytes as a string, this string is not base64 encoded.
```
reductionista: I think it would be worth mentioning here that this is guaranteed to return a string of length 8, in case someone calling it is unsure whether they need to validate what's returned.

reductionista:
The description of the PR as optimizing the event matching is misleading, since it's not optimizing anything relative to the current implementation. But it's a more self-contained way of handling the discriminators we might be able to re-use at some point, and adds negligible overhead so I am happy with it. Just pointed out a few minor cosmetic issues to fix.

pkg/solana/logpoller/log_poller.go
reductionista: (If not, remove TODO comment)

ilija42: DecodeSubKey uses codec decode which also checks for discriminator, so I don't think that there is any need for this extra validation step

pkg/solana/logpoller/filters.go
```diff
@@ -436,8 +436,9 @@ func (fl *filters) DecodeSubKey(ctx context.Context, raw []byte, ID int64, subKe
 	if err != nil || decodedEvent == nil {
 		return nil, err
 	}
-	err = decoder.Decode(ctx, raw, decodedEvent, filter.EventName)
-	if err != nil {
+	if err = decoder.Decode(ctx, raw, decodedEvent, filter.EventName); err != nil {
+		err = fmt.Errorf("failed to decode sub key raw data: %v, for filter: %s, for subKeyPath: %v, err: %w", raw, subKeyPath, filter.Name, err)
+		lggr.Criticalw(err.Error())
```
ilija42, line 467: critical because it shouldn't ever happen, it would mean that log discriminators match, but what the onchain contract emitted is malformed

reductionista (approved):

jadepark-dev (approved):

PR #1016, opened 2025-01-15
Query Filter Comparator Remaping

EasterTheBunny:
### Description

This PR adds support for remapping chain agnostic filters to solana specific filters.

[NONEVM-1063](https://smartcontract-it.atlassian.net/browse/NONEVM-1063)

### Requires Dependencies

- https://github.com/smartcontractkit/chainlink-common/pull/990

[NONEVM-1063]: https://smartcontract-it.atlassian.net/browse/NONEVM-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/logpoller/parser.go
```diff
@@ -396,19 +400,26 @@ func (f *addressFilter) Accept(visitor primitives.Visitor) {
...
 }
 
-func (f *eventSigFilter) Accept(visitor primitives.Visitor) {
+func (f *eventSubkeyFilter) Accept(visitor primitives.Visitor) {
 	switch v := visitor.(type) {
 	case *pgDSLParser:
-		v.VisitEventSigFilter(f)
+		v.VisitEventSubkeyFilter(f)
 	}
 }
```
reductionista, line 479: We can't combine the `eventSigFilter` and `eventSubkeyFIlter`.  Unlike in evm where the event sig is just "topic 0" the event sig in Solana has nothing to do with the event subkeys.  They are two separate columns in the db table, one has to compare the `event_sig` column and the other references an index into the `subkey_values` column.

Also the event sigs have a specific format, it must always be an 8 byte discriminator--we should validate that if we can I guess.  Whereas the subkey values are fields in a struct, and can be any primitive type; these have to be converted with `NewIndexedValue` before storing in the appropriate slot of `subkey_values`

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,229 @@
...
+	onChainName, ok := b.indexedSubkeys[comparator.Name]
+	if !ok {
+		return query.Expression{}, fmt.Errorf("%w: unknown indexed subkey mapping %s", types.ErrInvalidConfig, comparator.Name)
+	}
+
+	itemType := strings.Join([]string{b.namespace, b.genericName, comparator.Name}, ".")
+
+	for idx, comp := range comparator.ValueComparators {
+		// need to do a transform and then extract the value for the subkey
+		newValue, err := b.modifier.TransformToOnChain(comp.Value, itemType)
```
reductionista, line 146: We can't call this function, unless we modify the codec to skip the Borsh-encoding step.

I'm still not clear on whether we actually need to call this at all, but if we do then it either has to be without the Borsh-encoding or we'll need to decode it as soon as it finishes encoding.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logpoller"
+)
+
+type eventReadBinding struct {
+	namespace, genericName string
+	codec                  types.RemoteCodec
+	modifier               commoncodec.Modifier
+	key                    solana.PublicKey
+	remapper               remapHelper
+	indexedSubkeys         map[string]string
```
reductionista: I would lean towards calling this `subKeys` rather than `indexedSubkeys`. The latter might give someone the impression that we support non-indexed subKeys.

pkg/solana/codec/solana.go
```diff
@@ -174,6 +174,40 @@ func NewIDLInstructionsCodec(idl IDL, builder commonencodings.Builder) (commonty
 	return typeCodecs, nil
 }
 
+func NewIDLEventCodec(idl IDL, builder commonencodings.Builder) (commontypes.RemoteCodec, error) {
```
reductionista: There are two definitions of this and `eventFieldsAsStandardFields` now, probably happened because I included that in my PR and then you rebased onto it. I think they are identical, so just get rid of one or the other shouldn't matter.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	return sequences, nil
+}
+
+func (b *eventReadBinding) decodeLog(ctx context.Context, log *logpoller.Log, into any) error {
+	itemType := codec.WrapItemType(false, b.namespace, b.genericName, codec.ChainConfigTypeEventDef)
+
+	// decode non indexed topics and apply output modifiers
```
reductionista: This is an evm-specific comment. For Solana, it should just say "decode event data".

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: `TransformToOnchain` is a modifier function and doesn't handle encoding/decoding.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: This comes from the config where the name is `IndexedSubkeys` where I think it is good to imply they are indexed.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Good callout.

pkg/solana/logpoller/parser.go
EasterTheBunny: Ahhh. I messed that one up. I don't know what I was thinking. So essentially we need both: the event sig filter when subkey filters are provided.

EasterTheBunny: Looking at evm again, event sig is always provided. I'll make that change.

reductionista: Just backing out these changes to parser.go should be fine, as long as you're okay with merging my changes first (which add `VisitEventSubkeyFilter` while keeping `VisitEventSigFilter`)

pkg/solana/relay.go
```diff
@@ -162,7 +162,10 @@ func (r *Relayer) NewContractReader(_ context.Context, chainReaderConfig []byte)
 		return nil, fmt.Errorf("failed to init account reader err: %s", err)
 	}
 
-	return chainreader.NewChainReaderService(r.lggr, &chainreader.RPCClientWrapper{AccountReader: accountReader}, cfg)
+	// TODO: create a logpoller
+	var logpoller chainreader.EventsReader
```
reductionista: You don't need to create a new one here, it's already created in `NewChain()` which will have been called by the Relayer during its initialization:
https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/chain.go#L328-L329

All you should need is:
```suggestion
	logpoller := r.chain.LogPoller()
```
(or just pass that directly to `chainreader.NewContractReaderServices`) You could validate that it's not nil I guess, either here or inside there

reductionista: One thing that hasn't been added yet though is lp to the list of subservices started in `chain.Start()`:
https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/chain.go#L524

I figure there's no reason to start it yet, since it's not used by anything in production at the moment (ie, by DataFeeds). Q; Are we planning on having a feature flag for CCIP? Or are we going to have ChainReader & LogPoller started regardless of whether it's just an ocr node or a ccip node?

pkg/solana/codec/encoder_test.go
```diff
@@ -35,14 +35,15 @@ func (e *testErrEncodeTypeEntry) GetCodecType() commonencodings.TypeCodec {
 }
 
 func TestEncoder_Encode_Errors(t *testing.T) {
-	someType := "some-type"
+	someType := "intput.Some.Type"
```
reductionista: nit: Is that supposed to say input.Some.Type?

pkg/solana/logpoller/parser.go
```diff
@@ -21,6 +21,7 @@ const (
 	txHashFieldName       = "tx_hash"
 	addressFieldName      = "address"
 	eventSigFieldName     = "event_sig"
+	eventSubkeyFieldName  = "event_subkey"
```
reductionista: This file needs some more cleaning up. It looks like all of the changes to this file (and to `parser_test.go` I think?) should be reverted, aside from the addition of the `FormatContractReaderCursor(log Log) string` function.

For example, there is no `event_subkey` column in the logpoller tables, it's `subkey_values` which is already defined. The `eventSubkeyFilter` and all of its methods should be removed, we just need `eventBySubKeyFilter`.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+		// need to do a transform and then extract the value for the subkey
+		newValue, err := b.modifier.TransformToOnChain(comp.Value, itemType)
+		if err != nil {
+			return query.Expression{}, err
+		}
+
+		comparator.ValueComparators[idx].Value = newValue
+	}
+
+	return logpoller.NewEventSubkeyFilter(strings.Split(onChainName, "."), comparator.ValueComparators), nil
```
reductionista: This is calling the wrong constrctor. Needs to call this constructor:
```suggestion
	return logpoller.NewEventBySubkeyFilter(subKeyIndex, comparator.ValueComparators), nil
```

We don't actually need to find the `onChainName` at all here, since the subKeyIndex is the only thing that goes in the actual query (it's what's used to index the subkey_values column, which is an array).

Unfortunately, it looks like we'll probably have to do a bit of refactoring here to find the index, since that needs to be looked up in the ChainReader config.  There should be an array of subkeys specified in the ChainReader config for this event.  What needs to happen is it finds the index in that array and passes that down to LogPoller. The path itself isn't as helpful, because in order to use it it would have to construct a more complex query that JOIN's the logs table with the filters table.

pkg/solana/config/chain_reader.go
```diff
@@ -27,6 +31,9 @@ type ReadDefinition struct {
 	InputModifications  commoncodec.ModifiersConfig `json:"inputModifications,omitempty"`
 	OutputModifications commoncodec.ModifiersConfig `json:"outputModifications,omitempty"`
 	PDADefiniton        codec.PDATypeDef            `json:"pdaDefinition,omitempty"` // Only used for PDA account reads
+	IndexedFields       map[string]string           `json:"indexedFields,omitempty"`
+	// This will create a log poller filter for this event.
+	*PollingFilter `json:"pollingFilter,omitempty"`
```
reductionista, line 39: Ah, I see... this is trickier than I thought. We might need to change the config to have `IndexedFields` be a list of ordered pairs rather than a map 🤔

Either that, or we need enforce that they always get sorted using a deterministic sort before constructing the `logpoller.Filter` to pass to `RegisterFilter`, and in the event read binding before creating the `EventBySubKeyFilter`

pkg/solana/logpoller/parser.go
EasterTheBunny: What is the `SubKeyIndex`? How do I map a subkey path to an index?

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: I'm not following how we went from subkey paths to indexes.

pkg/solana/chainreader/chain_reader.go
```diff
@@ -300,14 +375,42 @@ func (s *SolanaChainReaderService) addAccountRead(namespace string, genericName
...
+	contractAddress solana.PublicKey,
+	_ codec.IDL,
+	_ codec.IdlEvent,
+	readDefinition config.ReadDefinition,
+	events EventsReader,
+) error {
+	subKeys := [][]string{}
+	for _, onChain := range readDefinition.IndexedFields {
+		subKeys = append(subKeys, strings.Split(onChain, "."))
+	}
```
reductionista: Ok, so this is the place where the ordering has to be deterministic.  I was thinking we could just fix this by sorting the keys of `readDefinition.IndexedFields` before we iterate over them like this:
```suggestion
	subKeys := [][]string{}
	for _, offChain := range slices.sorted(maps.Keys(readDefinition.IndexedFields)) {
		subKeys = append(subKeys, strings.Split(readDefinition.IndexedFields(offChain), "."))
	}
```
Unfortunately, this would only work as long as we never added any new subkeys to index. If we ever went back and updated the config, adding a new field of the event to index on, the new key would not be guaranteed to be placed at the end of the list... it would go somewhere in the middle, changing the index of all subkeys coming lexographicaly after it.

I think the only way to do this safely, without risking that we might accidentally start reading the wrong fields after a new one is added, is to have the order fixed in the config

We need to update the type of `IndexedFields` to be something like this:
```
IndexedFields []struct{
    onChainName string
    offChainName string
}
Then there is no need for sorting, and we can add new fields later without having to reindex the existing db tables.  Each subkey is indexed in a particular slot, so the only way to add a new field to index (if the slots aren't in a predictable order) would be to completely reindex all of the fields when a new one is added, by updating all of the logs associated with that event in the db.  That's an expensive operations, so it's better to have a predictable order of the indexed fields. Does that make sense?

File: pkg/solana/chainreader/chain_reader.go [hunk 12]
```diff
@@ -300,14 +375,42 @@ func (s *SolanaChainReaderService) addAccountRead(namespace string, genericName
...
+
+	filter := toLPFilter(readDefinition.PollingFilter, contractAddress, subKeys)
+
+	s.filters = append(s.filters, filter)
+	s.bindings.AddReadBinding(namespace, genericName, newEventReadBinding(
+		namespace,
+		genericName,
+		readDefinition.IndexedFields,
```
reductionista: The event binding won't need to know the onchain names.  Those are only used to decode the logs coming in from the rpc server.  (In other words, the purpose of the onChain field names is only to tell `Process` which fields in the struct it should extract and index).   Neither `QueryKey()` nor anything it calls needs to know the onchain field names, since the values have already been extracted and indexed.

So here, we could just pass the ordered slice of offChain names:
```suggestion
		offChainNames
```
where
```
var offChainNames []string
// build by iterating over IndexedFields
```
Although if you prefer to just pass the whole readDefinition.IndexedFields, and let the `AddReadBinding` method extract the part it needs that works too.

Ultimately, the most convenient form to save it in the `eventReadBinding` struct will be as a reverse map from offChainName to subKeyIndex:
```
var subKeyIndexes [string]int
```

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/logpoller"
+)
+
+type eventReadBinding struct {
+	namespace, genericName string
+	codec                  types.RemoteCodec
+	modifier               commoncodec.Modifier
+	key                    solana.PublicKey
+	remapper               remapHelper
+	indexedSubkeys         map[string]string
```
reductionista:
```suggestion
	subKeyIndexes         map[string]int
```
(could be int64 or uint64, etc. doesn't matter much I think? This is just going to be a number between 0 and NumDBIndexes - 1 (where NumDBIndexes is how many CREATE INDEX statements we have in the sql migration script, which determines the maximum # of subkey fields you can index.)

Currently we have NumDBIndexes = 4, but we can increase that if there's a product that needs more than 4 different fields of the same event indexed:
```
core/store/migrate/migrations/0262_add_solana_schema.sql:CREATE INDEX IF NOT EXISTS solana_logs_idx_subkey_one ON solana.logs ((subkey_values[1]));
core/store/migrate/migrations/0262_add_solana_schema.sql:CREATE INDEX IF NOT EXISTS solana_logs_idx_subkey_two ON solana.logs ((subkey_values[2]));
core/store/migrate/migrations/0262_add_solana_schema.sql:CREATE INDEX IF NOT EXISTS solana_logs_idx_subkey_three ON solana.logs ((subkey_values[3]));
core/store/migrate/migrations/0262_add_solana_schema.sql:CREATE INDEX IF NOT EXISTS solana_logs_idx_subkey_four ON solana.logs ((subkey_values[4]));
```

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+	events                 EventsReader
+	eventSig               [logpoller.EventSignatureLength]byte
+}
+
+func newEventReadBinding(
+	namespace, genericName string,
+	indexedSubkeys map[string]string,
+	events EventsReader,
+	eventSig [logpoller.EventSignatureLength]byte,
+) *eventReadBinding {
```
reductionista, line 36: Assuming the ordered slice of offChainNames is passed into this constructor, we'd build the `eventReadBinding.indexedSubKeys` map like this:
```
subKeyIndexes := make(map[string]int, len(onChainNames))
for i, onChainName := range onChainNames {
     subKeyIndexes[onChainName] = i
}
```

pkg/solana/chainreader/event_read_binding.go
reductionista: I left a bunch more comments, with code samples to illustrate what I'm suggesting and what I mean by the "subKeyIndex"

If you follow those suggestions or something like it, then all you'll need to pass here for the first arg to `NewEventBySubkeyFilter` is:
```
b.subKeyIndexes[comp.Value]
```

pkg/solana/chainreader/chain_reader.go
reductionista: > Although if you prefer to just pass the whole readDefinition.IndexedFields, and let the AddReadBinding method extract the part it needs that works too.

This might actually be better, come to think of it... it would eliminate the need for iterating over this list an extra time, just to extract the offChain names, since you have to iterate over them anyway in `AddReadBinding`

pkg/solana/chainreader/chain_reader.go
```diff
@@ -392,18 +392,21 @@ func (s *ContractReaderService) addEventRead(
 	readDefinition config.ReadDefinition,
 	events EventsReader,
 ) error {
-	subKeys := [][]string{}
-	for _, onChain := range readDefinition.IndexedFields {
-		subKeys = append(subKeys, strings.Split(onChain, "."))
-	}
+	mappedTuples := make(map[string]subkeyTuple)
```
reductionista: What is the purpose of the `string` field of `subkeyTuples`? It's not used anywhere. I think this can just be:

```
mappedTuples := make( map[string]uint64, 4)
```

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -17,20 +17,25 @@ import (
...
+}
+
 type eventReadBinding struct {
 	namespace, genericName string
 	codec                  types.RemoteCodec
 	modifier               commoncodec.Modifier
 	key                    solana.PublicKey
 	remapper               remapHelper
-	indexedSubkeys         map[string]string
+	indexedSubkeys         map[string]subkeyTuple
```
reductionista: `indexedSubKeys  map[string]uint64`

pkg/solana/config/chain_reader.go
```diff
@@ -54,6 +57,12 @@ func (r ReadType) String() string {
 	}
 }
 
+type IndexedField struct {
+	OffChainPath  string `json:"offChainPath"`
+	OnChainPath   string `json:"onChainPath"`
+	IndexPosition int    `json:"-"`
+}
+
```
reductionista, line 64: Isn't having `IndexPosition` a bit redundant here since it's already in the name of the config field? This opens the possibility for us ending up with, for example, an `IndexedFiled3` that has `IndexPosition` set to 2. Then we'd have to validate that they match, and/or guess whether we should use 2 or 3 for the index position, right?

I see at least 3 approaches for simplifying here:
1. Have a single`IndexedFields` defined as [4]IndexedFiled, with no `IndexPosition` field inside `IndexedField` type.
2. Keep it explicitly named `IndexedFiled0`, `IndexedField1`, but remove `IndexPosition`
3. Define IndexedFields as a map from chain agnostic key names to (on chain)) subKeyPaths:
```
type IndexedFields map[string]IndexedField
type IndexedField struct{
    subKeyPath string
    IndexPosition uint64
}
```
Also, regarding the naming of these fields:
OffChainPath doesn't seem right, these aren't paths, as I understand it.  They're just generic chain-agnostic subKey names... each should just be a simple string label, not a dot-separated path into anything (unless a product wants to use that to differentiate different keys somehow?)

For `OnChainPath`, that's probably alright but `subKeyPath` seems better to me since that's what it's named everywhere else.

reductionista (commented):
Looks mostly good, just a couple things look a bit more complex than they need to be

pkg/solana/chainreader/event_read_binding.go
reductionista: Oh yeah, regarding the capitalization: I recently updated all of the variables named `Subkey` to `SubKey` based on @ilija42's suggestion they be made consistent... so we should do that here as well.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+type eventReadBinding struct {
+	namespace, genericName string
+	codec                  types.RemoteCodec
+	modifier               commoncodec.Modifier
+	key                    solana.PublicKey
+	remapper               remapHelper
+	indexedSubkeys         map[string]subkeyTuple
+	events                 EventsReader
```
reductionista: nit: I got confused a few times thinking `events` was a list of events or something, rather than a pointer to logpoller.Service  Maybe we can name it `reader` or `eventReader`?

pkg/solana/config/chain_reader.go
reductionista: As far as names, I'd suggest `subKeyName` for the chain agnostic names of the subkeys, and for the on chain paths either `subKeyPath` or if you want I guess either `onChainPath` or `onChainSubKeyPath`?

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: The string value in the map is the `offChain` path.

pkg/solana/relay.go
EasterTheBunny: I just tried this and the interface doesn't have `FilteredLogs` which is necessary. Should we add this to the `LogPoller` interface expected by the relayer?

pkg/solana/chainreader/chain_reader.go
EasterTheBunny: The key is used in `pkg/solana/chainreader/event_read_binding.go` on line 143 to access the tuple using the `offChain` name.

pkg/solana/config/chain_reader.go
EasterTheBunny: I prefer the names in the config of `IndexedField0` because it is explicit about positioned indexes. I still think using positioned indexes like this will be a bit brittle because queries will be sensitive to position changes in the config.

Using explicit positioning in the config conveys the message that indexing is limited and index positions matter.

EasterTheBunny: You're right that `IndexPosition` is not needed as the position is set on the tuple.

pkg/solana/chainreader/chain_reader.go
reductionista: The string key is fine, that needs to stay. What should be removed is the `subkeyTuple` type... this should be replaced with `uint64`. (Re-read my comment above more carefully)

pkg/solana/relay.go
reductionista: Yes, I must have forgotten to add it when I added it to the `Service` struct in my latest PR--should be added

pkg/solana/chainreader/chain_reader.go
```diff
@@ -15,44 +16,59 @@ import (
...
+	cfg config.ContractReader,
+	events EventsReader,
+) (*ContractReaderService, error) {
+	svc := &ContractReaderService{
 		lggr:     logger.Named(lggr, ServiceName),
 		client:   dataReader,
 		bindings: namespaceBindings{},
 		lookup:   newLookup(),
 		parsed:   &codec.ParsedTypes{EncoderDefs: map[string]codec.Entry{}, DecoderDefs: map[string]codec.Entry{}},
+		filters:  []logpoller.Filter{},
```
reductionista, line 71: Just noticed the initialization of `events` field is missing here:
```suggestion
		filters:  []logpoller.Filter{},
		events: events,
```
Also, reminder... weren't we going to rename this?

reductionista (dismissed):

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if filter, err = b.remapper.remap(filter); err != nil {
+		return nil, err
+	}
+
+	// filter should always use the address and event sig
+	filter.Expressions = append([]query.Expression{
+		logpoller.NewAddressFilter(pubKey),
+		logpoller.NewEventSigFilter(b.eventSig[:]), // TODO: genericName is wrong; need the event sig hash
```
ilija42: Is this TODO outdated?

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+	)
+
+	switch primitive := expression.Primitive.(type) {
+	case *primitives.Comparator:
+		if comp, err = b.encodeComparator(primitive); err != nil {
+			return query.Expression{}, fmt.Errorf("failed to encode comparator %q: %w", primitive.Name, err)
+		}
+	case *primitives.Confidence:
+		// TODO: maybe use an ignore filter?
+		// confidence is ignored in solana
```
ilija42: What would an ignore filter do? just ignoring should be fine

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	return sequences, nil
+}
+
+func (b *eventReadBinding) decodeLog(ctx context.Context, log *logpoller.Log, into any) error {
+	itemType := codec.WrapItemType(false, b.namespace, b.genericName)
+
+	// decode non indexed topics and apply output modifiers
```
ilija42: evm comment

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Yes

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: It would be something that does nothing. I was thinking ignoring should be fine too.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Oops

ilija42 (approved):

reductionista (approved):

toblich (approved):

PR #1009, opened 2025-01-10
Added ArgsTransform and Updated CCIP Config Example to be up-to-date

silaslenihan:
### Description

- Added `ArgsTransform` configuration to CW config which allows the args passed in to be arbitrarily altered. This will enable the CCIP use case of calculating the TokenIndexes. 

- Updated CCIP Config example to be up-to-date with on-chain code and include example `ArgsTransform` for CCIP execute. 

 

### Requires Dependencies

- https://github.com/smartcontractkit/chainlink-solana/pull/1017
- https://github.com/smartcontractkit/chainlink-ccip/pull/487
- https://github.com/smartcontractkit/chainlink-ccip/pull/505

### Resolves Dependencies

- https://github.com/smartcontractkit/chainlink/pull/15848

PR #1017, opened 2025-01-15
Added bitmap location parsing for IsWritable and IsSignable on AccountLookups

silaslenihan:
### Description

This PR updates `AccountLookup`s to be able to search for `AccountMeta` values based on a provided Bitmap. This is needed for CCIP messaging `ExtraArgs` which will contain a bitmap for which accounts are writable.
 

### Resolves Dependencies

- https://github.com/smartcontractkit/chainlink-solana/pull/1009

### DIFF
--- a/integration-tests/relayinterface/lookups_test.go
+++ b/integration-tests/relayinterface/lookups_test.go
@@ -23,6 +23,7 @@ import (
 
 type InnerAccountArgs struct {
 	Accounts []*solana.AccountMeta
+	Bitmap   uint64
 }
 
 type TestAccountArgs struct {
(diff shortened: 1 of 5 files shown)


integration-tests/relayinterface/lookups_test.go
```diff
@@ -117,12 +125,201 @@ func TestAccountLookups(t *testing.T) {
...
+			Location:   "Inner.Accounts.PublicKey",
+			IsSigner:   chainwriter.MetaBool{Location: "ExternalBool"},
+			IsWritable: chainwriter.MetaBool{Location: "ExternalBool"},
+		}
+
+		args := TestAccountArgsExtended{
+			Inner: InnerAccountArgs{
+				Accounts: accounts[:],
+			},
+			ExternalBool: true,
```
amit-momin: Not really feedback but curious if you're handling this scenario just to be flexible or if this is an expected use case where we refer to the same value for all of the provided addresses. I would have expected if the `IsSigner` and `IsWritable` fields were provided, they would be given for each account.

integration-tests/relayinterface/lookups_test.go
```diff
@@ -117,12 +125,201 @@ func TestAccountLookups(t *testing.T) {
...
+
+		result, err := lookupConfig.Resolve(ctx, args, nil, nil)
+		require.NoError(t, err)
+
+		for i, meta := range result {
+			require.Equal(t, accounts[i], meta)
+		}
+	})
+
+	t.Run("AccountLookup fails with MetaBool an invalid number of Meta lookups", func(t *testing.T) {
```
amit-momin: nit: think the description is missing something? Also could you add the success scenario for this case too just so we can make sure the correct index bool is applied to the correct account? 
```suggestion
	t.Run("AccountLookup fails with MetaBool due to an invalid number of Meta lookups", func(t *testing.T) {
```

integration-tests/relayinterface/lookups_test.go
silaslenihan: This is mostly just to be flexible. The only concrete use-case we have of this is when the address to lookup and the boolean are in the same parent struct, but theoretically there is a case where you would want to look up the Meta based on a single field in the args.

integration-tests/relayinterface/lookups_test.go
silaslenihan: Fixed the description, but isn't that basically what some of the above tests are checking already? I added [this](https://github.com/smartcontractkit/chainlink-solana/blob/47f38136b823777243a6d5d849f2c33ed259561e/integration-tests/relayinterface/lookups_test.go#L141-L142) to the first test to ensure the order is being checked

amit-momin: Oh right, I was thinking the first test was different since it was looking back into the parent struct and this was an external field.

amit-momin (dismissed):

amit-momin (approved):

aalu1418 (approved):

PR #1027, opened 2025-01-25
Updated interface tests to use unique PDA per test

amit-momin:


pkg/solana/codec/byte_string_modifier.go
```diff
@@ -27,9 +27,9 @@ func (s SolanaAddressModifier) DecodeAddress(str string) ([]byte, error) {
 		return nil, fmt.Errorf("%w: failed to decode Base58 address: %s", commontypes.ErrInvalidType, err)
 	}
 
-	if pubkey.IsZero() {
-		return nil, fmt.Errorf("%w: zero-value address", commontypes.ErrInvalidType)
-	}
+	// if pubkey.IsZero() {
```
amit-momin: Wanted to call this out if it was problematic. This interfered with interface tests when initializing accounts with an empty pubkey field. Don't think this is necessarily needed for the CCIP use case but is this a validation we need?

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -100,171 +118,350 @@ func DisableTests(it *SolanaChainComponentsInterfaceTester[*testing.T]) {
...
+	CreateAccount(t T, it SolanaChainComponentsInterfaceTester[T], contractName string, value uint64, testStruct TestStruct) solana.PublicKey
 	TXM() *txm.TxManager
 	SolanaClient() *client.Client
 }
 
-type SolanaChainComponentsInterfaceTester[T TestingT[T]] struct {
+type WrappedTestingT[T any] interface {
+	TestingT[T]
+	Name() string
+}
```
amit-momin, line 143: This is probably something we could add to `chainlink-common` but added here to avoid changes in another repo.

PR #1033, opened 2025-01-29
Enable Automatic ATA Creation in ChainWriter

silaslenihan:
core ref: 8b048022c936ee2ca405707b4664081a44b9a26d
### Description
Enables ATA Creation when necessary based on the ChainWriter config. The code is based on [this design](https://docs.google.com/document/d/1HeQ-6JeOO4h8naQ5AghmgmwhzjpKTsZxBR41eCOS9Nk/edit?pli=1&tab=t.0), but has been altered slightly. 

Also, this PR refactors the ChainWriter to use the MultiClient rather than a single RPC client set at initialization.


pkg/solana/chainwriter/chain_writer.go
```diff
@@ -310,10 +389,13 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
 	discriminator := GetDiscriminator(methodConfig.ChainSpecificName)
 	encodedPayload = append(discriminator[:], encodedPayload...)
 
+	// Combine the two sets of instructions into one slice
+	var instructions []solana.Instruction
+	instructions = append(instructions, createATAinstructions...)
```
silaslenihan: Probably makes sense to split this into 2 separate TXs to save space for the main TX. @amit-momin does the TXM guarantee order of execution natively, or do we have to poll for finality on the first TX before sending the second one?

amit-momin: Just mentioning this here for visibility before I make the change. I think we may need to poll for finality if we split it. We would broadcast the transactions sequentially but don't think there's any guarantees that they are included in that order. Since they interact with the same account, I think Solana's localized gas markets would mean the tx with the higher price out of the 2 would get included first. Since they would be individual tx in the TXM, I don't think we can make the distinction to ensure the ATA tx always stays higher priced.

silaslenihan: Ok. Regardless I think it makes sense to split them up since tx sizes are so small on Solana. Should probably check with the on-chain team to confirm they agree

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -214,6 +218,76 @@ func (s *SolanaChainWriterService) FilterLookupTableAddresses(
...
+
+		mints, err := GetAddresses(ctx, args, []Lookup{lookup.MintAddress}, derivedTableMap, reader, idl)
+		if err != nil {
+			return nil, fmt.Errorf("error resolving mint address: %w", err)
+		}
+		if len(tokenPrograms) != len(mints) {
+			return nil, fmt.Errorf("expected equal number of token programs and mints, got %d tokenPrograms and %d mints", len(tokenPrograms), len(mints))
+		}
+
+		for i := range tokenPrograms {
```
huangzhen1997, line 270: Is it guaranteed that `tokenPrograms` and  `mints` have same size ? If not we might want to check

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -220,6 +225,76 @@ func (s *SolanaChainWriterService) FilterLookupTableAddresses(
...
 }
 
+// CreateATAs first checks if a specified location exists, then checks if the accounts derived from the
+// ATALookups in the ChainWriter's configuration exist on-chain and creates them if they do not.
+func CreateATAs(ctx context.Context, args any, lookups []ATALookup, derivedTableMap map[string]map[string][]*solana.AccountMeta, client client.MultiClient, idl string, feePayer solana.PublicKey) ([]solana.Instruction, error) {
+	createATAInstructions := []solana.Instruction{}
+	for _, lookup := range lookups {
+		// Check if location exists
+		if lookup.Location != "" {
+			// TODO refactor GetValuesAtLocation to not return an error if the field doesn't exist
```
amit-momin: Do we still need to address this?

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -316,10 +390,41 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
...
 
+	// Fetch latest blockhash
+	blockhash, err := s.client.LatestBlockhash(ctx)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error fetching latest blockhash: %w", err), debugID)
+	}
+
+	if len(createATAinstructions) > 0 {
+		ataTx, ataErr := solana.NewTransaction(
+			createATAinstructions,
```
amit-momin: Not sure about the worst case scenario but can we can safely just pass all the instructions into 1 transaction and not worry about tx size? If we're not sure and we know the avg size of an ATA instruction, we can maybe set a batch limit

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -316,10 +390,41 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
...
+			blockhash.Value.Blockhash,
+			solana.TransactionPayer(feePayer),
+		)
+		if ataErr != nil {
+			return errorWithDebugID(fmt.Errorf("error constructing ATA transaction: %w", err), debugID)
+		}
+
+		// Enqueue ATA transaction
+		if err = s.txm.Enqueue(ctx, methodConfig.FromAddress, ataTx, &transactionID, blockhash.Value.LastValidBlockHeight); err != nil {
```
amit-momin: I think we should create a custom transactionID for the ATA transactions. Maybe something like `ATA-{uuid}`? Otherwise, the TXM will collide on this ID.

pkg/solana/chainwriter/chain_writer.go
prashantkumar1982: Yes, we can remove this TODO.
Field not existing should be an error, and the way this code currently handles it, is correct.

pkg/solana/chainwriter/chain_writer.go
prashantkumar1982: Good catch!
If we reuse the transactionID, the TXM will always return ErrIDAlreadyExists when you actually send the main Tx.
I would recommend creating a new uuid here itself, for the ATA tx. This will allow us to track its status.

Using ATA-{uuid} can be problem if the caller hadn't set uuid. Because then we'd use "ATA-" as the id for all ATA txs, and they will then collide with each other.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -220,6 +225,76 @@ func (s *SolanaChainWriterService) FilterLookupTableAddresses(
...
+			if err != nil {
+				return nil, fmt.Errorf("error deriving ATA: %w", err)
+			}
+
+			_, err = client.GetAccountInfoWithOpts(ctx, ataAddress, &rpc.GetAccountInfoOpts{
+				Encoding:   "base64",
+				Commitment: rpc.CommitmentFinalized,
+			})
+			if err == nil {
+				continue
```
prashantkumar1982, line 285: Info log here, saying this ataAddress was already existing, so not creating.

And then I realized there is no logging anywhere in this file.
Perhaps have a separate ticket, to later on add logs in this file. Will help a lot during debugging.

pkg/solana/chainwriter/lookups.go
```diff
@@ -98,7 +98,17 @@ type AccountsFromLookupTable struct {
 	LookupOpts
 }
 
-func (ac AccountConstant) Resolve(_ context.Context, _ any, _ map[string]map[string][]*solana.AccountMeta, _ client.Reader) ([]*solana.AccountMeta, error) {
+type ATALookup struct {
+	// Field that determines whether the ATA lookup is necessary. Basically
+	// just need to check this field exists. Dot separated location.
+	Location string
+	// If the field exists, initialize a ATA account using the Wallet, Token Program, and Mint addresses below
```
prashantkumar1982, line 95: I'm getting confused by these 3 fields.
Ideally, when we do a Token Transfer, aren't these fields already derivable from the CCIP message, right?

For example Any2SVMRampMessage.TokenReceiver, and Any2SVMTokenTransfer.DestTokenAddress are the fields.
I think 
WalletAddress = Any2SVMRampMessage.TokenReceiver
TokenProgram = Any2SVMTokenTransfer.DestTokenAddress
MintAddress = ?

So is it right to assume that these 3 fields you added will correspond to the location in the message, for these fields?

Another thing to note is that in CCIP, the TokenReceiver is just 1 entity, but multiple TokenPrograms.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -220,6 +225,76 @@ func (s *SolanaChainWriterService) FilterLookupTableAddresses(
...
+		walletAddresses, err := GetAddresses(ctx, args, []Lookup{lookup.WalletAddress}, derivedTableMap, client)
+		if err != nil {
+			return nil, fmt.Errorf("error resolving wallet address: %w", err)
+		}
+		if len(walletAddresses) != 1 {
+			return nil, fmt.Errorf("expected exactly one wallet address, got %d", len(walletAddresses))
+		}
+		wallet := walletAddresses[0].PublicKey
+
+		tokenPrograms, err := GetAddresses(ctx, args, []Lookup{lookup.TokenProgram}, derivedTableMap, client)
```
prashantkumar1982, line 257: In CCIP, we know the location of tokenPrograms will be an array.
I recall, you already added the support to flatten the array into multiple values right?
Note that in CCIP case, the leaf node isn't the array field, but an intermediate path is.

Any2SVMRampMessage.TokenAmounts.DestTokenAddress
Here, the TokenAmounts is the array field.

I wasn't sure if current code will enumerate all the TokenAmounts here.

prashantkumar1982 (commented):
I assume we will also need to update the CW configs to reflect this.
When/who will do that?

pkg/solana/chainwriter/chain_writer.go
amit-momin: Oh sorry I sort of wrote that confusingly. I meant we can just generate a uuid on the spot since this txID is transient. We only need it for the next step when we poll the status so something like `ATA-{uuid.NewString()}` is what I meant

pkg/solana/chainwriter/chain_writer.go
silaslenihan: +1 above, I had originally considered doing it a different way and forgot to removed this TODO

pkg/solana/chainwriter/chain_writer.go
silaslenihan: After some testing, I've concluded that 12 ATAs is the maximum number of batch create ATA instructions that can be included in a single transaction. I've added a check that ensures the length of the `createATAinstructions` doens't exceed this.

pkg/solana/chainwriter/chain_writer.go
silaslenihan: added a custom txID in the form `fmt.Sprintf("ATA-%s", uuid.NewString())`

pkg/solana/chainwriter/lookups.go
silaslenihan: I think there is a bit of confusion on what these fields are.
1. WalletAddress above is correct.
2. TokenProgram isn't the DestTokenAddress, rather it is the program the token program that the dest token implements. I believe on solana this can either be the pre-2022 Token Program or the Token 2022 program. There may be more added in the future. I believe this is analagous to the ERC20, ERC667, etc. interfaces on EVM. Therefore we have to look this value up with the token address.
3. The MintAddress is the token address itself, i.e. the DestTokenAddress

silaslenihan: > So is it right to assume that these 3 fields you added will correspond to the location in the message, for these fields?

No, not always. For CCIP the TokenProgram needs to be looked up in the `TokenAdminRegistry` lookup table

> Another thing to note is that in CCIP, the TokenReceiver is just 1 entity, but multiple TokenPrograms.

Yes this is a valid and expected case. As you can see in the code [here](https://github.com/smartcontractkit/chainlink-solana/blob/2df78b7c9b27939849643f9555136e7d8639780e/pkg/solana/chainwriter/chain_writer.go#L253-L266) we check that there is a single wallet yet there can be multiple tokenPrograms and mints

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Yes I added in this functionality it will work as expected

silaslenihan:
> I assume we will also need to update the CW configs to reflect this. When/who will do that?

It should be a relatively simple addition. I can create a PR to add it in after this one goes through

pkg/solana/chainwriter/chain_writer.go
silaslenihan: I added some logging to the ChainWriter, including in this place

pkg/solana/chainwriter/chain_writer.go
amit-momin: Maybe we should double check with on-chain if it's a valid scenario to need more than 12 ATA instructions. If it is, we may want to allow this but just batch them into different transactions. These we could probably fire all at the same time. Unless we want to keep a limit anyways to cap costs but regardless think we should double check with on-chain.

pkg/solana/chainwriter/chain_writer.go
amit-momin: Worried Info might get too noisy since they would log on every transactions. Could we get away with Debug instead? Except maybe for the initialization log and the `Sending main transaction` one?

pkg/solana/chainwriter/lookups.go
prashantkumar1982: So if TokenProgram can be looked up from the wallet address, then is it possible to remove from here?

pkg/solana/chainwriter/chain_writer.go
prashantkumar1982: Yes makes sense about noise.
We could do the other way around.
Log an info log if we are creating an ATA. Since we will less frequently create ATAs, it wouldn't be too much noise.

pkg/solana/chainwriter/chain_writer.go
prashantkumar1982: I see that check is made lines above

pkg/solana/chainwriter/chain_writer.go
prashantkumar1982: Yes, batching is preferred.
If you don't want to do that right away, atleast instead of failing this check 
"if len(createATAinstructions) > maxAtas",
Tt is better to cap ATAs to maxAtas.
This way, we create 12 ATAs only, and ognore others. This will mean the original Tx will too fail, but then we have other nodes in the DON, which will pick this up, and create the remaining ATAs and the CCIP message won't get blocked.

Another workaround is to get confirmation from CCIP team that 12 token transfers in 1 message isn't ever going to happen.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -316,10 +402,47 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
...
+		ataUUID := fmt.Sprintf("ATA-%s", uuid.NewString())
+
+		s.lggr.Info("Sending create ATA transaction", "contract", contractName, "method", method)
+
+		// Enqueue ATA transaction
+		if err = s.txm.Enqueue(ctx, methodConfig.FromAddress, ataTx, &ataUUID, blockhash.Value.LastValidBlockHeight); err != nil {
+			return errorWithDebugID(fmt.Errorf("error enqueuing transaction: %w", err), debugID)
+		}
+
+		err = s.waitForTxFinality(ctx, transactionID)
```
prashantkumar1982: transactionID is wrong.
Should be ataUUID.

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -316,10 +402,47 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
 	discriminator := GetDiscriminator(methodConfig.ChainSpecificName)
 	encodedPayload = append(discriminator[:], encodedPayload...)
 
+	// Fetch latest blockhash
+	blockhash, err := s.client.LatestBlockhash(ctx)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("error fetching latest blockhash: %w", err), debugID)
+	}
+
+	if len(createATAinstructions) > 0 {
```
prashantkumar1982, line 402: nit: Optional suggestion.
This function is becoming too big to read. Maybe same for chain_writer file?
Maybe move this if block out of this function, to another function handlePossibleATACreation()
Maybe even create a new go file ata_creation.go

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Changed all logs to Debug besides the initialization and the ATA creation ones

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Broke this out into a separate function in a different file

pkg/solana/chainwriter/lookups.go
silaslenihan: I don't quite understand what you're suggesting. How would we define the lookup to TokenProgram if we remove it here?

pkg/solana/chainwriter/chain_writer.go
silaslenihan: From Agustina's comment in the slack thread, it seems that just failing here is the best option. The execution report will not go through with so many accounts, so there's no point wasting the gas to create the ATAs.

amit-momin: Based on offline discussions, sounds like >12 ATA instructions isn't possible so I'm ok with keeping this as an error.

amit-momin (dismissed):

amit-momin (dismissed):

prashantkumar1982 (approved):

toblich (approved):

PR #1036, opened 2025-01-31
If localnet, be less strict on the expected chainID

archseer:
In deployment tests we use a bunch of random identifiers to allow us to run multiple different solana chains at once. The current relayer implementation is quite rigid which causes a mismatch. This PR makes the check more permissive (but still disallows using official IDs on localnet)

### DIFF
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math/big"
 	"math/rand"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/solana/chain.go
```diff
@@ -150,7 +151,9 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
 
 	// check chainID matches expected chainID
 	expectedChainID := strings.ToLower(v.expectedChainID)
-	if v.chainID != expectedChainID {
+	// if this is localnet, allow any chain ID as long as it's not spoofing an official network
+	ignore := v.chainID == "localnet" && !slices.Contains([]string{"mainnet", "testnet", "devnet"}, expectedChainID)
```
tt-cll, line 155: do we have these consts defined anywhere?

tt-cll (dismissed):

tt-cll (approved):

PR #1041, opened 2025-02-04
NONEVM-1292 fix orm & filters

dhaidashenko:
core ref: 001c3c46470a8498080a00f56cc257d0d1cb112a


pkg/solana/logpoller/orm.go
```diff
@@ -53,7 +53,7 @@ func (o *DSORM) InsertFilter(ctx context.Context, filter Filter) (id int64, err
 		withEventSig(filter.EventSig).
 		withStartingBlock(filter.StartingBlock).
 		withEventIDL(filter.EventIdl).
-		withSubKeyPaths(filter.SubKeyPaths).
+		withSubKeyPaths(filter.SubkeyPaths).
```
ilija42, line 56:
```suggestion
		withSubKeyPaths(filter.SubKeyPaths).
```

ilija42, line 56: Is subKey or subkey more correct? Either way `withSubKeyPaths` are not aligned `filter.SubkeyPaths`

reductionista: In a recent commit I changed everything to be `SubKey` and `SubKeyPaths`, based on a suggestion from @ilija42 since there was a mix of some of one and some of the other capitalization.

reductionista: Oh, I see... I guess my change introduced a bug,in this case it has to match the db column name, which is `subkey_paths`.  Hmmm, not sure what we should do to make this consistent.

reductionista: Same issue with `subkey_values`

pkg/solana/logpoller/types.go
```diff
@@ -126,19 +127,14 @@ type Decoder interface {
...
 func (e *EventIdl) Scan(src interface{}) error {
 	return scanJSON("EventIdl", e, src)
 }
 
 func (e EventIdl) Value() (driver.Value, error) {
-	return json.Marshal(map[string]any{
-		"IdlEvent":        e.EventIDLTypes.Event,
-		"IdlTypeDefSlice": e.EventIDLTypes.Types,
-	})
+	return json.Marshal(e)
```
reductionista, line 137: Is there a reason why we need to save the rest of the junk in the idl? AFAIK we should only need these two fields the rest just takes up unnecessary space. Ideally, ChainReader should remove the extra fields before it's passed to LogPoller, but I added this so at least we're not persisting them.

pkg/solana/logpoller/orm.go
reductionista: This exposes a limitation with our decision to only test the ORM with a live db, and to mock the ORM for testing higher layers of the api. In this case, doing a more e2e test that included both the DSL parser and writing to the actual db would have caught this bug

reductionista: Actually the dsl parser seems like a fairly unique exception. Because it generates actual sql we ought to be testing the validity of that sql, and its consistency with the sql migration scripts in the `chainlink` repo. Using a mocked ORM shouldn't be a problem for other components.

pkg/solana/codec/anchoridl.go
```diff
@@ -195,6 +212,29 @@ type IdlTypeArray struct {
...
+		array := env.GetArray()
+		result = map[string]interface{}{
+			"array": []interface{}{array.Thing, array.Num},
+		}
+	default:
+		return nil, fmt.Errorf("nil envelope is not supported in IdlType")
+	}
+
+	return json.Marshal(result)
+}
```
reductionista, line 236: Nice!

pkg/solana/logpoller/types.go
reductionista: Oh, is it because this was messing up the `(e EventIdl) Equal` method? If that's the only reason... we could change that to compare those two fields instead of the whole thing.

pkg/solana/logpoller/types.go
```diff
@@ -191,6 +187,32 @@ func (v *IndexedValue) FromFloat64(f float64) {
...
+}
+
+func (v IndexedValues) Value() (driver.Value, error) {
+	byteArray := make(pq.ByteaArray, len(v))
+	for i, b := range v {
+		byteArray[i] = b
+	}
+
+	return byteArray.Value()
+}
```
reductionista, line 214: Thanks, forgot to add these

ilija42 (approved):

pkg/solana/logpoller/orm.go
dhaidashenko: This issue exposes limitation of not running orm test in the CI. 
I agree that we need to have consistent naming and SubKey is correct one.
At the same time I'm against doing this right now. I still struggling to make LP fetch the data from chain and store it to DB. It's higher priority to make LP work.

pkg/solana/logpoller/types.go
dhaidashenko: `codec.EventIDLTypes` has only two fileds. Ilija changed it in another PR.
```
type EventIDLTypes struct {
	Event IdlEvent
	Types IdlTypeDefSlice
}
```

reductionista: Oh, nice!

reductionista (dismissed):

pkg/solana/logpoller/orm_test.go
```diff
@@ -183,16 +202,17 @@ func TestLogPollerFilters(t *testing.T) {
 func TestLogPollerLogs(t *testing.T) {
```
reductionista, line 202: Does this test pass for you? When I run it locally, it fails on line 227 while comparing `CreatedAt` timestamps:
```
require.Equal(t, log, dbLogs[0]
```

reductionista: Don't worry about this. There are several issues here, and I ran into the same issues in a test I'm writing for my current PR. I added a helper function to santize logs, I can fix this test in my PR along with it.

reductionista (approved):

jadepark-dev (approved):

PR #1042, opened 2025-02-04
Implement multi read

ilija42:
core ref: 461d94af9431de5ee79447ab6972e3d2f64775fe

### Description

Helps solve Offramp contract discovery by  aggregating mulitple PDA reads from the same program into one return value


amit-momin (dismissed):

amit-momin (dismissed):

amit-momin (approved):

pkg/solana/chainreader/chain_reader.go
```diff
@@ -154,13 +153,21 @@ func (s *ContractReaderService) GetLatestValue(ctx context.Context, readIdentifi
 
 	values, ok := s.lookup.getContractForReadIdentifiers(readIdentifier)
```
reductionista, line 154: I guess this is a pre-existing issue, but this method name seems very misleading. The name sounds like the input would be a slice of read identifiers and the output would be a contract. Shouldn't it be called `getReadValuesForReadIdentifier`?

chris-de-leon-cll (approved):

PR #1044, opened 2025-02-05
Nonevm 1065/events for standard and batch reads

EasterTheBunny:
### Description

Connect events to GetLatestValue and BatchGetLatestValue.

[NONEVM-1065](https://smartcontract-it.atlassian.net/browse/NONEVM-1065)

[NONEVM-1065]: https://smartcontract-it.atlassian.net/browse/NONEVM-1065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Requires Dependencies

- https://github.com/smartcontractkit/chainlink-solana/pull/1016
- https://github.com/smartcontractkit/chainlink-solana/pull/1061

### DIFF
--- a/pkg/solana/chainreader/batch.go
+++ b/pkg/solana/chainreader/batch.go
@@ -70,18 +70,43 @@ func doMultiRead(ctx context.Context, client MultipleAccountGetter, bdRegistry *
 }
 
 func doMethodBatchCall(ctx context.Context, client MultipleAccountGetter, bdRegistry *bindingsRegistry, batch []call) ([]batchResultWithErr, error) {
-	// Create the list of public keys to fetch
-	keys := make([]solana.PublicKey, len(batch))
+	results := make([]batchResultWithErr, len(batch))
+
+	// create the list of public keys to fetch
+	keys := []solana.PublicKey{}
+
+	// map batch call index to key index (some calls are event reads and will be handled by a different binding)
+	dataMap := make(map[int]int)
+
 	for idx, batchCall := range batch {
 		rBinding, err := bdRegistry.GetReader(batchCall.Namespace, batchCall.ReadName)
 		if err != nil {
 			return nil, fmt.Errorf("%w: read binding not found for contract: %q read: %q: %w", types.ErrInvalidConfig, batchCall.Namespace, batchCall.ReadName, err)
 		}
 
-		keys[idx], err = rBinding.GetAddress(ctx, batchCall.Params)
+		key, err := rBinding.GetAddress(ctx, batchCall.Params)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get address for contract: %q read: %q: %w", batchCall.Namespace, batchCall.ReadName, err)
 		}
+
+		eBinding, ok := rBinding.(eventBinding)
+		if ok {
+			results[idx] = batchResultWithErr{
+				address:   key.String(),
+				namespace: batchCall.Namespace,
+				readName:  batchCall.ReadName,
+				returnVal: batchCall.ReturnVal,
+			}
+
+			results[idx].err = eBinding.GetLatestValue(ctx, batchCall.Params, results[idx].returnVal)
+
+			continue
+		}
+
+		// map the idx to the key idx
+		dataMap[idx] = len(keys)
+
+		keys = append(keys, key)
 	}
 
 	// Fetch the account data
@@ -90,18 +115,21 @@ func doMethodBatchCall(ctx context.Context, client MultipleAccountGetter, bdRegi
 		return nil, err
 	}
 
-	results := make([]batchResultWithErr, len(batch))
-
 	// decode batch call results
 	for idx, batchCall := range batch {
+		dataIdx, ok := dataMap[idx]
+		if !ok {
+			return nil, fmt.Errorf("%w: unexpected data index state", types.ErrInternal)
+		}
+
 		results[idx] = batchResultWithErr{
-			address:   keys[idx].String(),
+			address:   keys[dataIdx].String(),
 			namespace: batchCall.Namespace,
 			readName:  batchCall.ReadName,
 			returnVal: batchCall.ReturnVal,
 		}
 
-		if data[idx] == nil || len(data[idx]) == 0 {
+		if data[dataIdx] == nil || len(data[dataIdx]) == 0 {
 			results[idx].err = ErrMissingAccountData
 
 			continue
@@ -114,30 +142,35 @@ func doMethodBatchCall(ctx context.Context, client MultipleAccountGetter, bdRegi
 			continue
 		}
 
-		results[idx].err = errors.Join(
-			decodeReturnVal(ctx, rBinding, data[idx], results[idx].returnVal),
-			results[idx].err)
+		results[idx].err = asValueDotValue(
+			ctx,
+			rBinding,
+			results[dataIdx].returnVal,
+			wrapDecodeValuer(rBinding, data[dataIdx]),
+		)
 	}
 
 	return results, nil
 }
 
-// decodeReturnVal checks if returnVal is a *values.Value vs. a normal struct pointer, and decodes accordingly.
-func decodeReturnVal(ctx context.Context, binding readBinding, raw []byte, returnVal any) error {
-	// If we are not dealing with a `*values.Value`, just decode directly.
+// asValueDotValue checks if returnVal is a *values.Value vs. a normal struct pointer, and decodes accordingly.
+func asValueDotValue(
+	ctx context.Context,
+	binding readBinding,
+	returnVal any,
+	op func(context.Context, any) error,
+) error {
 	ptrToValue, isValue := returnVal.(*values.Value)
 	if !isValue {
-		return binding.Decode(ctx, raw, returnVal)
+		return op(ctx, returnVal)
 	}
 
-	// Otherwise, we need to create an intermediate type, decode into it,
-	// wrap it, and set it back into *values.Value
 	contractType, err := binding.CreateType(false)
 	if err != nil {
 		return err
 	}
 
-	if err = binding.Decode(ctx, raw, contractType); err != nil {
+	if err = op(ctx, contractType); err != nil {
 		return err
 	}
 
@@ -149,4 +182,10 @@ func decodeReturnVal(ctx context.Context, binding readBinding, raw []byte, retur
 	*ptrToValue = value
 
 	return nil
+}
+
+func wrapDecodeValuer(binding readBinding, data []byte) func(context.Context, any) error {
+	return func(ctx context.Context, returnVal any) error {
+		return binding.Decode(ctx, data, returnVal)
+	}
 }
--- a/pkg/solana/chainreader/event_read_binding.go
+++ b/pkg/solana/chainreader/event_read_binding.go
@@ -208,12 +208,18 @@ func (b *eventReadBinding) GetLatestValue(ctx context.Context, params, returnVal
 		return err
 	}
 
-	limiter := query.NewLimitAndSort(query.CountLimit(1), query.NewSortBySequence(query.Desc))
-	filter, err := logpoller.Where(
+	allFilters := []query.Expression{
 		logpoller.NewAddressFilter(pubKey),
 		logpoller.NewEventSigFilter(b.eventSig[:]),
-		subkeyFilters,
-	)
+	}
+
+	if len(subkeyFilters) > 0 {
+		allFilters = append(allFilters, query.And(subkeyFilters...))
+	}
+
+	limiter := query.NewLimitAndSort(query.CountLimit(1), query.NewSortBySequence(query.Desc))
+
+	filter, err := logpoller.Where(allFilters...)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


pkg/solana/logpoller/log_poller.go
```diff
@@ -382,3 +382,22 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+	if err != nil {
+		return []query.Expression{}, err
+	}
+
+	if filter.Expressions == nil {
+		return []query.Expression{}, nil
+	}
+
+	return filter.Expressions, nil
+}
```
reductionista: Shouldn't this Where helper go in `query.go` or `parser.go`, rather than `log_poller.go`?

reductionista: Looks like `query` would make the most sense, since all of the types it uses start with `query.`

reductionista: Wait, actually why is this even in the chainlink-solana repo? Looks like a chain-agnostic helper that belongs in `chainlink-common`

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -75,6 +78,46 @@ func (b *eventReadBinding) Decode(ctx context.Context, bts []byte, outVal any) e
...
+	filter, err := logpoller.Where(
+		logpoller.NewAddressFilter(pubKey),
+		logpoller.NewEventSigFilter(b.eventSig[:]),
+		subkeyFilters,
+	)
+	if err != nil {
+		return err
+	}
+
+	logs, err := b.reader.FilteredLogs(ctx, filter, limiter, b.namespace+"-"+pubKey.String()+"-"+b.genericName)
```
reductionista, line 227: Not sure how many other places there are where we do this `b.namespace+"-"+pubKey.String()+"-"+b.genericName", but assuming this isn't the only place a helper for it might make sense

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -115,6 +158,56 @@ func (b *eventReadBinding) QueryKey(
...
+		)
+		if err != nil {
+			return query.Expression{}, err
+		}
+
+		expressions = append(expressions, expression)
+	}
+
+	if len(expressions) == 0 {
+		return query.Expression{}, fmt.Errorf("%w: no topic filters found during query creation", types.ErrInternal)
```
reductionista: evm-specific error message

pkg/solana/chainreader/batch.go
```diff
@@ -31,19 +32,44 @@ type MultipleAccountGetter interface {
...
 	for idx, call := range batch {
 		binding, err := bindings.GetReadBinding(call.ContractName, call.ReadName)
 		if err != nil {
 			return nil, err
 		}
 
+		lvBinding, ok := binding.(latestValueBinding)
+		if ok {
+			results[idx] = batchResultWithErr{
+				address:      "", // TODO: get contract public key
```
reductionista: Is this needed for e2e testing? If not, maybe there's a JIRA # we could include?

amit-momin: Just double checking if this TODO is going to be addresses in a follow up PR? Don't we have the contract public key available here through the binding?

pkg/solana/logpoller/log_poller.go
EasterTheBunny: It probably does belong in common, but that will take longer to add.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Probably another thing that might go well in common since evm is very similar.

pkg/solana/chainreader/batch.go
EasterTheBunny: I made a goof. The address is the contract address which can come from the binding here.

EasterTheBunny: Fixed

amit-momin (dismissed):

pkg/solana/logpoller/log_poller.go
reductionista: Alright, we can skip it for now.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -115,6 +158,56 @@ func (b *eventReadBinding) QueryKey(
...
+
+		onChainValue, err := b.modifier.TransformToOnChain(offChainValue, itemType)
+		if err != nil {
+			return query.Expression{}, fmt.Errorf("%w: failed to apply on-chain transformation for key %s", types.ErrInternal, offChainKey)
+		}
+
+		// check that onChainValue is not zero value for type
+		if reflect.ValueOf(onChainValue).IsZero() {
+			continue
+		}
```
reductionista, line 316: I'm not sure I follow. Why are we treating zero values as special?

pkg/solana/chainreader/batch.go
```diff
@@ -43,7 +50,26 @@ func doMethodBatchCall(ctx context.Context, client MultipleAccountGetter, bindin
...
+		if ok {
+			results[idx] = batchResultWithErr{
+				address:      key.String(),
+				contractName: call.ContractName,
+				readName:     call.ReadName,
+				returnVal:    call.ReturnVal,
+			}
+
+			// is an event read
+			results[idx].err = lvBinding.GetLatestValue(ctx, call.Params, results[idx].returnVal)
```
reductionista: Is this "is an event read" comment right? How do we know it's an event read as opposed to an account read?

reductionista: if that's determined by whether the `latestValueBinding` interface is satisfied, maybe that and/or `lvBinding` should be called `eventBinding` instead?

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -115,6 +158,56 @@ func (b *eventReadBinding) QueryKey(
...
+	// in the case that the input params are presented as a map of values, apply the values to the off-chain type
+	// with solana hooks
+	if err = codec.MapstructureDecode(value, offChain); err != nil {
+		return nil, fmt.Errorf("%w: failed to decode offChain value: %s", types.ErrInternal, err.Error())
+	}
+
+	return offChain, nil
+}
+
+func (b *eventReadBinding) extractFilterSubkeys(offChainValue any) (query.Expression, error) {
```
reductionista: If I understand this correctly,`offChainValue` here is a map of key-value params right?  If so, would be easier to read if we renamed it to something like `offChainParams`.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Because a zero value could make it's way to the filter without the caller having set it to be included. It's the consequence of using struct types for filtering. A very common Go ORM, `gorm`, handles filtering this way too. Zero values are ignored when passing structs.

pkg/solana/chainreader/batch.go
EasterTheBunny: I agree it's confusing, but the account reads don't use `GetLatestValue`. They do batch data reads instead. I could rename these to offer better clarity.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: I can change it if it adds some clarity.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: I'm also fairly certain this is how the EVM implementation works too.

EasterTheBunny: Coming back to this, I can expand the explanation of what is happening under the hood to show why this works.

The type returned from the transform or create type contains pointer values for the fields even if the final returned struct doesn't:

```
struct { A: *int, B: *string }
```

So the zero value of the above fields is nil, which is what the value should be after `mapstructure.Decode` runs if the value isn't provided in a `map[string]any`.

reductionista: Ah, ok if they're always converted to pointers that makes sense.  I just wanted to make sure that it's possible for someone to search for, say, subKeyValue1 = 0 and get back the right results

reductionista (dismissed):

amit-momin (dismissed):

ilija42 (changes requested):
Lets wait on merging this because of merge conflicts till E2E work is done

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -228,6 +270,56 @@ func (b *eventReadBinding) QueryKey(
...
+		)
+		if err != nil {
+			return query.Expression{}, err
+		}
+
+		expressions = append(expressions, expression)
+	}
+
+	if len(expressions) == 0 {
+		return query.Expression{}, fmt.Errorf("%w: no subkey filters found during query creation", types.ErrInternal)
```
reductionista: Something doesn't sound right here. The only required field should be the EventSig (aka discriminator).

As I understand it, the EventSig is the "key" and the fields inside the event struct are the "subkeys". This comment makes it sound as if the plugin is required to narrow the search down by subkey. But one of the most common use cases of QueryKey is where you only search by Address & EventSig.

Hopefully this is just a semantic difference, and not an issue with the code?

reductionista: I do see that you have a separate `EventSig` filter here:
```
		logpoller.NewAddressFilter(pubKey),
		logpoller.NewEventSigFilter(b.eventSig[:]),
		subkeyFilters,
```
So I guess what you're calling a subkey is the same as what I think of as a subkey. But that means, you can't error out here when there are no subkey filters... right? I think that's a common case, even for CCIP

reductionista (changes requested):
Unless I'm missing something, we need to remove the check for len(expressions) = 0 in `extractSubkeys` so that it can handle calls to QueryKey which filter only on Address & EventSig

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Well this is for `GetLatestValue`, but you have a point. If no params exist, that is still a valid event call. I'll make that change.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -504,3 +513,36 @@ func (k *indexedSubkeys) indexForKey(key string) (uint64, bool) {
...
+
+		if tail == "" {
+			return field.Interface(), nil
+		}
+
+		return valueForPath(field, tail)
+	default:
+		return nil, fmt.Errorf("%w: cannot extract a field from kind %s", types.ErrInvalidType, from.Kind())
+	}
+}
```
reductionista, line 548: Don't we need to handle fields that are maps also?

This looks pretty similar to this function.  Is there a way we could combine them, maybe putting it in the codec?
https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/logpoller/filters.go#L508-L560

There's also a third version of it implemented in ChainWriter somewhere, but that one I think is aimed at a slightly different use case. Ideally, we'll eventually combine all of these together in a more general helper.

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -286,38 +292,41 @@ func (b *eventReadBinding) normalizeParams(value any, itemType string) (any, err
...
 
-func (b *eventReadBinding) extractFilterSubkeys(offChainParams any) (query.Expression, error) {
+func (b *eventReadBinding) extractFilterSubkeys(offChainParams any) ([]query.Expression, error) {
 	var expressions []query.Expression
 
 	for offChainKey, idx := range b.indexedSubKeys.lookup {
 		itemType := codec.WrapItemType(true, b.namespace, b.genericName+"."+offChainKey)
 
-		onChainValue, err := b.modifier.TransformToOnChain(offChainParams, itemType)
+		fieldVal, err := valueForPath(reflect.ValueOf(offChainParams), offChainKey)
```
reductionista, line 301: Oh... I thought an offchain key was just supposed to be some generic key name, is it actually a path sometimes?

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: This is a copy of an unexported function in common. We should use that one and add appropriate type handling there.

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Yes, it is a path to a value in a struct.

reductionista (approved):

jadepark-dev (approved):

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -188,6 +190,52 @@ func (b *eventReadBinding) Decode(ctx context.Context, bts []byte, outVal any) e
 	return b.codec.Decode(ctx, bts, outVal, itemType)
 }
 
+func (b *eventReadBinding) GetLatestValue(ctx context.Context, params, returnVal any) error {
+	itemType := codec.WrapItemType(true, b.namespace, b.genericName)
+
+	pubKey, err := b.GetAddress(ctx, nil)
+	if err != nil {
+		return err
+	}
```
ilija42, line 199:
```suggestion
	pubKey := b.key
```

pkg/solana/chainreader/event_read_binding.go
ilija42: Can you create a ticket that references all similar use cases and the common code?

pkg/solana/logpoller/parser.go
```diff
@@ -496,3 +496,22 @@ func makeComp(comp IndexedValueComparator, args *queryArgs, field, subfield, pat
 		args.withIndexedField(field, comp.Value),
 	), nil
 }
+
+// Where is a query.Where wrapper that ignores the Key and returns a slice of query.Expression rather than
+// query.KeyFilter. If no expressions are provided, or an error occurs, an empty slice is returned.
+func Where(expressions ...query.Expression) ([]query.Expression, error) {
```
ilija42, line 502: Can you create a ticket for cleaning this up?

ilija42 (commented):
Looks good! Do we have a follow up PR with interface tests for this and QueryKey?

EasterTheBunny:
> Looks good! Do we have a follow up PR with interface tests for this and QueryKey?

Yes, that is [NONEVM-1064](https://github.com/smartcontractkit/chainlink-solana/pull/1084)

[NONEVM-1064]: https://smartcontract-it.atlassian.net/browse/NONEVM-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: `GetAddress` offers a read lock on the key so it is safer to call the public method than to access the value directly.

pkg/solana/logpoller/parser.go
EasterTheBunny: This was copied from EVM directly. By cleaning up, do you think this should go in common or have different behavior?

ilija42 (approved):

pkg/solana/logpoller/parser.go
ilija42: Go into common

PR #1049, opened 2025-02-07
Upgraded chainlink-ccip solana dependency

amit-momin:
core ref: 41da9bafca488a87b23a7847e014122c31a74d91

- Pinned `chainlink-ccip/chains/solana` to the latest commit
- Fixed breaking changes in the ChainWriter code
  - CCIP args transform method still relies on the router's IDL for the `TokenAdminRegistry`

### DIFF
--- a/pkg/solana/chainwriter/transform_registry.go
+++ b/pkg/solana/chainwriter/transform_registry.go
@@ -44,7 +44,7 @@ func CCIPArgsTransform(ctx context.Context, cw *SolanaChainWriterService, args a
 	}
 	// PDA lookup to fetch router address
 	routerAddrLookup := PDALookups{
-		Name:      "ReferenceAddresses",
+		Name: "ReferenceAddresses",
 		PublicKey: AccountConstant{
 			Address: toAddress,
 		},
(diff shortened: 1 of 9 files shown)


pkg/solana/chainwriter/transform_registry.go
```diff
@@ -37,14 +37,43 @@ func FindTransform(id string) (func(context.Context, *SolanaChainWriterService,
 // This Transform function looks up the token pool addresses in the accounts slice and augments the args
 // with the indexes of the token pool addresses in the accounts slice.
 func CCIPArgsTransform(ctx context.Context, cw *SolanaChainWriterService, args any, accounts solana.AccountMetaSlice, toAddress string) (any, error) {
+	// Fetch offramp config to use to fetch the router address
```
prashantkumar1982, line 40: This function gets called on each execute() invocation.
And it has InternalField set, meaning it will make an RPC call each time?

I don't think the router address changes that often.
We might want to use a cache, or perhaps initialize the router address in CR, and pass it here?

If too problematic, we could leave it as it is for now, to be optimized later.

amit-momin: That's right, we would have to make an RPC call to fetch the router address. This is sort of the quick and dirty solution because of timelines but agree this probably could be improved. We probably could add an `ArgsTransformAddress` that pipes into this function and can be statically set to the router address but wanted to avoid a config change for now.

archseer (approved):

PR #1051, opened 2025-02-07
Lazy-start LogPoller from ContractReader, stop if started on shutdown

reductionista:
### Description

- Start LogPoller from ChainReader, if there are events defined in ChainReaderConfig, and it's not already started yet
- Adds a lot more testing coverage to ChainReader's Start() method

JIRA:  [NONEVM-1286](https://smartcontract-it.atlassian.net/browse/NONEVM-1286)

[NONEVM-1286]: https://smartcontract-it.atlassian.net/browse/NONEVM-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/chain.go
```diff
@@ -547,6 +548,9 @@ func (c *chain) Close() error {
 			c.lggr.Debug("Stopping multinode")
 			closeAll = append(closeAll, c.multiNode, c.txSender)
 		}
+		if c.lp.Ready() == nil {
```
ilija42, line 551: What happens in ready throws an error

ilija42 (dismissed):

pkg/solana/chainreader/chain_reader.go
```diff
@@ -109,13 +111,24 @@ func (s *ContractReaderService) Name() string {
 // and error.
 func (s *ContractReaderService) Start(ctx context.Context) error {
 	return s.StartOnce(ServiceName, func() error {
+		if len(s.filters) == 0 {
+			// No dependency on EventReader
+			return nil
+		}
+		if s.reader.Ready() != nil {
```
dhaidashenko, line 124: This seems to be racy. Ready returns an error if the reader is starting. 
In case of LogPoller, it's easy to decide if it needs to do work or not. Can't we always start and check if there are any filters inside the run loop?

pkg/solana/chainreader/chain_reader.go
```diff
@@ -457,7 +470,7 @@ func toLPFilter(
 		Address:     logpoller.PublicKey(address),
 		EventName:   f.EventName,
 		EventSig:    logpoller.EventSignature([]byte(f.EventName)[:logpoller.EventSignatureLength]),
```
dhaidashenko:
```suggestion
		EventSig:    logpoller.EventSignature([]byte(f.EventName)[:logpoller.EventSignatureLength]),
		EventSig: logpoller.NewEventSignatureFromName(f.EventName),
```

dhaidashenko: Also eventIDL is not set. Seems like we need our own e2e tests as soon as possible.

pkg/solana/chain.go
reductionista: The only two things it can return are Unstarted or nil, so there are no other errors to handle.

pkg/solana/chainreader/chain_reader.go
reductionista: Good catch! Although I think that doesn't fully fix it.  The EventName can't simply be cast to an EventSignature. I think it should be:
```
EventSig:    logpoller.NewEventSignatureFromName(f.EventName),
```
Does that look right @ilija42 ?

pkg/solana/chainreader/chain_reader.go
reductionista: Ugh... I think you're right. Maybe there isn't a way of guaranteeing this isn't racy. Maybe what you're suggesting would be the simplest way of doing this without introducing too much risk.

pkg/solana/chain.go
reductionista: Actually I guess that's not quite right. It only returns one error, which means it's not in state==stateStarted, but there are a number of other possible states besides stateStarted and stateUnstarted it could be in:
```
	stateUnstarted state = iota
	stateStarted
	stateStarting
	stateStartFailed
	stateStopping
	stateStopped
	stateStopFailed
```
I guess the only problematic one is stateStarting... which is sort of another way of saying what @dhaidashenko said, which is that it could be racy. We don't want to start it if it's already starting... even if it's not fully started yet.

pkg/solana/chainreader/chain_reader.go
reductionista: You're right, this could fail if two ContractReader services both tried to start LogPoller at the same time.
The StateMachine.StartOnce()  method called by Start() does handle it atomically though, so I think all we need to do is ignore the error if it returns  an "already started" error.

> In case of LogPoller, it's easy to decide if it needs to do work or not. Can't we always start and check if there are any filters inside the run loop?

Before implementing the above, I looked over the LogPoller startup code to see how feasible this would be. I think handling things there is a bit more messy and error prone. In part because there is a cascade of sub services started automatically by the engine. And in part because the first thing it does is load filters from the db. If it doesn't allow the db read to happen due to no filters being present, it will be deadlocked and unable to start.  But if it does allow reading from the db, then suddenly we're introducing a lot of extra risk for the existing production OCR2 node ops, who are currently running without the Solana relay ever opening any db connections. The present version of the Solana relay they're running doesn't have access to a database at all, since it's not needed for DataFeeds. At least until we've fully e2e tested all this code I think it's safer for them to not have to access the database or start LogPoller.

dhaidashenko (approved):

ilija42 (approved):

jadepark-dev (approved):

PR #1053, opened 2025-02-09
Handle GetTokenPrices

ilija42:


pkg/solana/chainreader/chain_reader.go
```diff
@@ -164,6 +168,10 @@ func (s *ContractReaderService) GetLatestValue(ctx context.Context, readIdentifi
 		return doMultiRead(ctx, s.client, s.bdRegistry, values, returnVal)
 	}
 
+	if values.multiRead[0] == "GetTokenPrices" {
```
EasterTheBunny: If this hard-coded solution is temporary, we should make a note of that somewhere here. It also looks like this will essentially override anything added to the config for this read name. I think for this it's ok, but we might put some checks in the `init` to return an error for overlaps and consider this read name to be _reserved_. 

Also what happens if this is called through `BatchGetLatestValue` or with a return val of `value.Value`?

ilija42: > Also what happens if this is called through BatchGetLatestValue or with a return val of value.Value?

won't work

EasterTheBunny: I'm guessing it doesn't need to work?

pkg/solana/chainreader/chain_reader.go
```diff
@@ -495,3 +512,141 @@ func applyIndexedFieldTuple(lookup map[string]uint64, subKeys [4][]string, conf
...
+			"for contract %q read %q: %w (could not parse program address %q)",
+			values.contract, values.multiRead[0], types.ErrInvalidConfig, values.address,
+		)
+	}
+
+	// Build the PDA addresses for all tokens.
+	var pdaAddresses []solana.PublicKey
+	for _, token := range *tokens {
+		tokenAddr := solana.PublicKeyFromBytes(token[:])
+		if tokenAddr.IsOnCurve() && !tokenAddr.IsZero() {
```
amit-momin: Based on the error message below, shouldn't this be 
```suggestion
		if tokenAddr.IsOnCurve() || tokenAddr.IsZero() {
```

ilija42: it should be on curve and not zero, the err msg is bad

amit-momin: Just to leave a note here too for visibility. A valid token address is on the curve and is non-zero. Updated the error case accordingly.

amit-momin (dismissed):

EasterTheBunny (approved):

amit-momin (approved):

jadepark-dev (approved):

PR #1061, opened 2025-02-10
Make LP Filter Registration like EVM

EasterTheBunny:
### Description

Align LogPoller filter registration logic to be similar to EVM. Filters are checked and registered on `ContractReader` startup and unregistered on close.

### DIFF
--- a/pkg/solana/chainreader/bindings.go
+++ b/pkg/solana/chainreader/bindings.go
@@ -42,8 +42,22 @@ type readBinding interface {
 
 type addressShareGroup struct {
 	address solana.PublicKey
-	mux     sync.Mutex
+	mux     sync.RWMutex
 	group   []string
+}
+
+func (g *addressShareGroup) getAddress() solana.PublicKey {
+	g.mux.RLock()
+	defer g.mux.RUnlock()
+
+	return g.address
+}
+
+func (g *addressShareGroup) setAddress(addr solana.PublicKey) {
+	g.mux.Lock()
+	defer g.mux.Unlock()
+
+	g.address = addr
 }
 
 type bindingsRegistry struct {
@@ -136,7 +150,14 @@ func (r *bindingsRegistry) GetReaders(namespace string) ([]readBinding, error) {
 	return rBindings.GetReaders()
 }
 
-func (r *bindingsRegistry) Bind(ctx context.Context, reg filterRegistrar, binding types.BoundContract) error {
+// Bind has a side-effect of updating the bound address to a group shared address.
+//
+// DO NOT CHANGE binding from pointer type.
+func (r *bindingsRegistry) Bind(ctx context.Context, reg filterRegistrar, binding *types.BoundContract) error {
+	if binding == nil {
+		return fmt.Errorf("%w: bound contract is nil", types.ErrInvalidType)
+	}
+
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
@@ -217,28 +238,26 @@ func (r *bindingsRegistry) getShareGroup(nameSpace string) (*addressShareGroup, 
 	return shareGroup, sharesAddress
 }
 
-func (r *bindingsRegistry) handleAddressSharing(boundContract types.BoundContract) error {
+func (r *bindingsRegistry) handleAddressSharing(boundContract *types.BoundContract) error {
 	shareGroup, isInAGroup := r.getShareGroup(boundContract.Name)
 	if !isInAGroup {
 		return nil
 	}
 
-	shareGroup.mux.Lock()
-	defer shareGroup.mux.Unlock()
-
 	// set shared address to the binding address
-	if shareGroup.address.IsZero() {
+	if shareGroup.getAddress().IsZero() {
 		key, err := solana.PublicKeyFromBase58(boundContract.Address)
 		if err != nil {
 			return err
 		}
 
-		r.addressShareGroups[boundContract.Name].address, shareGroup.address = key, key
-	} else if boundContract.Address != shareGroup.address.String() && boundContract.Address != "" {
-		return fmt.Errorf("namespace: %q shares address: %q with namespaceBindings: %v and cannot be bound with a new address: %s", boundContract.Name, shareGroup.address, shareGroup.group, boundContract.Address)
+		shareGroup.setAddress(key)
+	} else if boundContract.Address != shareGroup.getAddress().String() && boundContract.Address != "" {
+		return fmt.Errorf("namespace: %q shares address: %q with namespaceBindings: %v and cannot be bound with a new address: %s", boundContract.Name, shareGroup.getAddress(), shareGroup.group, boundContract.Address)
 	}
 
-	boundContract.Address = shareGroup.address.String()
+	// side-effect of updating the bound contract address to group-shared address
+	boundContract.Address = shareGroup.getAddress().String()
 
 	return nil
 }
--- a/pkg/solana/chainreader/bindings.go
+++ b/pkg/solana/chainreader/bindings.go
@@ -137,8 +137,12 @@ func (r *bindingsRegistry) GetReaders(namespace string) ([]readBinding, error) {
 }
 
 func (r *bindingsRegistry) Bind(ctx context.Context, reg filterRegistrar, binding types.BoundContract) error {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if err := r.handleAddressSharing(binding); err != nil {
+		return err
+	}
 
 	namespace, nbsExist := r.namespaceBindings[binding.Name]
 	if !nbsExist {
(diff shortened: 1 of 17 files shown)


pkg/solana/chainreader/bindings.go
```diff
@@ -13,167 +14,420 @@ import (
...
+
+			r.addressShareGroups[namespace] = shareGroup
 		}
 	}
+
+	return nil
+}
+
+func (r *bindingsRegistry) getShareGroup(nameSpace string) (*addressShareGroup, bool) {
+	shareGroup, sharesAddress := r.addressShareGroups[nameSpace]
```
amit-momin, line 247: Wouldn't we need a read lock here?

pkg/solana/chainreader/bindings.go
```diff
@@ -13,167 +14,420 @@ import (
...
+func newNamespaceBinding(namespace string) *namespaceBinding {
+	return &namespaceBinding{
+		name:    namespace,
+		readers: make(map[string]readBinding),
+		bound:   make(map[solana.PublicKey]bool),
 	}
+}
+
+func (b *namespaceBinding) SetCodecs(codec types.RemoteCodec) {
+	b.mu.RLock()
```
amit-momin, line 298: Shouldn't we use a write lock for the set method?

pkg/solana/chainreader/bindings.go
```diff
@@ -13,167 +14,420 @@ import (
...
+	defer b.mu.RUnlock()
+
+	for _, rb := range b.readers {
+		rb.SetCodec(codec)
+	}
+}
+
+func (b *namespaceBinding) SetModifiers(modifier commoncodec.Modifier) {
+	b.mu.RLock()
```
amit-momin, line 307: Same here for the write lock

pkg/solana/chainreader/bindings.go
```diff
@@ -13,167 +14,420 @@ import (
...
-	boundContract.Address = shareGroup.address.String()
 	return nil
 }
 
-func (b *bindingsRegistry) getShareGroup(nameSpace string) (*addressShareGroup, bool) {
-	shareGroup, sharesAddress := b.addressShareGroups[nameSpace]
-	if !sharesAddress {
-		return nil, false
+func (b *namespaceBinding) BindReaders(ctx context.Context, address solana.PublicKey) error {
+	b.mu.RLock()
```
amit-momin, line 326: And also here for write locks?

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -262,3 +384,25 @@ func (r remapHelper) remapExpression(key string, expression query.Expression) (q
...
+
+func (k *indexedSubkeys) addForIndex(offChainPath, onChainPath string, idx uint64) {
+	k.lookup[offChainPath] = idx
+	k.subKeys[idx] = strings.Split(onChainPath, ".")
+}
+
+func (k *indexedSubkeys) indexForKey(key string) (uint64, bool) {
+	idx, ok := k.lookup[key]
+
+	return idx, ok
```
DylanTinianov, line 407: nit
```suggestion
	return k.lookup[key]
```

pkg/solana/chainreader/bindings.go
EasterTheBunny: It would be safe, but the groups don't change after initialization.

pkg/solana/chainreader/bindings.go
EasterTheBunny: The `*namespaceBinding` is only doing a read on `b.readers`.

pkg/solana/chainreader/bindings.go
EasterTheBunny: The `*namespaceBinding` is only doing a read on `b.readers`.

pkg/solana/chainreader/bindings.go
EasterTheBunny: The `*namespaceBinding` is only doing a read on `b.readers`.

pkg/solana/chainreader/bindings.go
amit-momin: Never mind, realized the set is on the reader so you're right only needing a read lock here and below

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: This hasn't worked in the past. Does it work now?

pkg/solana/chainreader/bindings.go
```diff
@@ -13,167 +14,420 @@ import (
...
+	rbs, rbsExists := b.readers[genericName]
+	if !rbsExists {
+		return nil, fmt.Errorf("%w: no read binding exists for %s", types.ErrInvalidConfig, genericName)
+	}
+
+	return rbs, nil
+}
+
+func (b *namespaceBinding) GetReaders() ([]readBinding, error) {
+	allBindings := make([]readBinding, len(b.readers))
```
amit-momin, line 384: Wouldn't we need a read lock here?

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -5,54 +5,142 @@ import (
...
+	}
+
+	b.registerCalled = true
+
+	// can't be true before filters params are set so there is no race with a bad filter outcome
+	if !b.bound {
+		return nil
+	}
+
+	return b.filter.Register(ctx, b.reader)
```
DylanTinianov, line 122: Do we still want to set `b.registerCalled = true` if we return before calling `b.filter.Register(ctx, b.reader)`?
```suggestion
	// can't be true before filters params are set so there is no race with a bad filter outcome
	if !b.bound {
		return nil
	}
	b.registerCalled = true
	return b.filter.Register(ctx, b.reader)
```

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: Nope. Doing a direct return like that assumes the value without the boolean check. :(

pkg/solana/chainreader/bindings.go
EasterTheBunny: Ahh shoot yes

pkg/solana/chainreader/event_read_binding.go
EasterTheBunny: It's the same in EVM. I'm not quite sure why, but I think a lot of this on that side was done by @ilija42.

DylanTinianov (dismissed):

pkg/solana/chainreader/bindings.go
EasterTheBunny: fixed

pkg/solana/chainreader/bindings.go
EasterTheBunny: fixed

amit-momin (dismissed):

pkg/solana/config/chain_reader.go
```diff
@@ -5,14 +5,38 @@ import (
...
-
 	commoncodec "github.com/smartcontractkit/chainlink-common/pkg/codec"
 	commontypes "github.com/smartcontractkit/chainlink-common/pkg/types"
 
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codec"
 )
 
+const (
+	DefaultLogPollerRetention = 24 * time.Hour
+	DefaultMaxLogsKept        = 0
```
reductionista, line 16: I think 24 hours is too short of a default for this

As I understood, the plan was to have retention for CCIP jobs set to 30 days... which is what it's set to on EVM.  In LogPoller, the default retention of "0" represents infinite (permanent retention) on both EVM & Solana... although we haven't implemented pruning on Solana yet, so it will currently be infinite no matter what this value is set to.

Ideally, we should just have the default here be 0, which LogPoller will interpret as "permanent retention".  But if we're planning to re-use this field to also compute StartingBlock that may pose problems. In that case, using 30 days for the default would seem reasonable.

DylanTinianov (dismissed):

pkg/solana/chainreader/filter.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (r *syncedFilter) AddressSet() bool {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	return r.addressSet
+}
+
+func (r *syncedFilter) HasIndexedSubkeys() bool {
```
reductionista: Is this used anywhere?

pkg/solana/config/chain_reader.go
reductionista: I just looked at `toLPFilter` and `StartingBlock` isn't filled in at all. That's more important than `Retention` since `Retention` and `MaxLogsKept` are presently unused

The two main options I'd recommend would be either:
1. Use Retention field of `PollingFilter` for both purposes. Set default to 30 days.
2. Add a separate `Lookback` field of `PollingFilter` to specify how far in the past we need to start fetching logs when the filter first gets registered, or when it's fallen far enough behind that it is allowed to skip ahead to catch up. Default for this can be 0, interpreted as "current time / starting block" and default for retention can be 0 as well (but in that case, will represent "permanent retention").

Either is fine, option 1 is probably easier for now.  But whichever way it's done, we need to compute the StartingBlock somehow, to pass that to RegisterFilter. This would presumably be based on taking the time interval and dividing by the block time for the chain, then subtracting that from the current block.

amit-momin (dismissed):

pkg/solana/chainreader/filter.go
EasterTheBunny: I'll remove it

amit-momin (approved):

chris-de-leon-cll (approved):

reductionista (approved):

PR #1062, opened 2025-02-11
Implement multi read parent read params usage

ilija42:
### Description
Enables CR to multi read same PDA account multiple times, this is useful for handling very complex chain agnostic <> chain specific data restructuring without having to immediately develop new codec modifiers.


EasterTheBunny (approved):

amit-momin (approved):

pkg/solana/chainreader/lookup.go
```diff
@@ -6,12 +6,19 @@ import (
...
+}
+
 type readValues struct {
 	address  string
 	contract string
-	// First read in multi read has type info that other sequential reads are filling out.
-	// this works by having hard coder codec modifier define fields that are filled out by subsequent reads.
-	multiRead []string
+	// reads of size one is a regular Account read.
+	// When multiple read reads are present, first read has type info that other sequential reads are filling out.
```
reductionista, line 19: should "multiple read reads" say "multiple reads"?

contracts/programs/contract-reader-interface/src/lib.rs
```diff
@@ -125,6 +141,36 @@ pub struct InitializeMultiReadOnce<'info> {
...
 
+#[derive(Accounts)]
+pub struct InitializeMultiReadWithParamsOnce<'info> {
+    #[account(mut)]
+    pub signer: Signer<'info>,
+
+    #[account(
+        init_if_needed,
+        payer = signer,
+        space = size_of::<MultiRead3>() + 8,
```
toblich, line 152: Suggestion: instead of doing this `size_of::`, the Anchor way of doing this is to:
- Add `#[derive(InitSpace)]` to the struct
- Use `MultRead3::INIT_SPACE` here
See [the docs](https://www.anchor-lang.com/docs/references/space#the-initspace-macro). That way it will also handle some other types, such as vectors (where you can specify a `max_len` to it for example).

chris-de-leon-cll (approved):

toblich (approved):

PR #1075, opened 2025-02-13
Fix get token prices read handling

ilija42:
### Description

Fix GetTokenPrices special handling so that its cfg can be json serialised and deserialised  without type loss


pkg/solana/chainreader/chain_reader.go
```diff
@@ -635,72 +637,84 @@ func (s *ContractReaderService) handleGetTokenPricesGetLatestValue(
...
 		return err
 	}
 
 	data, err := s.client.GetMultipleAccountData(ctx, pdaAddresses...)
 	if err != nil {
-		return fmt.Errorf(
-			"for contract %q read %q: failed to get multiple account data: %w",
-			values.contract, values.reads[0].readName, err,
-		)
+		return err
```
amit-momin, line 656: Is there a reason we didn't want to keep more verbose errors? Unless this is redundant in the call stack, wouldn't we want to keep this?

pkg/solana/chainreader/chain_reader.go
```diff
@@ -187,7 +187,9 @@ func (s *ContractReaderService) GetLatestValue(ctx context.Context, readIdentifi
 
 	// TODO this is a temporary edge case - NONEVM-1320
 	if values.reads[0].readName == GetTokenPrices {
-		return s.handleGetTokenPricesGetLatestValue(ctx, params, values, returnVal)
+		if err := s.handleGetTokenPricesGetLatestValue(ctx, params, values, returnVal); err != nil {
```
amit-momin, line 190: Wouldn't this returnVal get overwritten by the batch call? I could be missing something though since the context wouldn't be in the PR.

amit-momin (dismissed):

pkg/solana/chainreader/chain_reader.go
ilija42: it would be redundant, the caller attaches the details by wrapping the error

pkg/solana/chainreader/chain_reader.go
ilija42: Not completely sure what you mean by this, this won't be used in a batch call

chris-de-leon-cll (dismissed):

pkg/solana/chainreader/chain_reader.go
amit-momin: I probably shouldn't have referred to it as a batch call but I meant just the `doMethodBatchCall` call below. But think I just answered my own question looking at this again.

amit-momin (approved):

chris-de-leon-cll (approved):

chris-de-leon-cll (approved):

PR #1076, opened 2025-02-13
Add global replay feature to LogPoller

reductionista:
### Description

  [NONEVM-1260](https://smartcontract-it.atlassian.net/browse/NONEVM-1260)

Add global replay feature to LogPoller

This adds the functionality to LogPoller. Later, we can add cli commands for starting a cli command for starting a replay and checking whether there is one in progress.

There are two new exported methods of logpoller.Service:

```
// Replay submits a new replay request. If there was already a new replay request
// submitted since the last replay completed, it will be updated to the earlier of the
// two requested fromBlock's. The expectation is that, on the next timer tick of the
// LogPoller run loop it will backfill all filters starting from fromBlock. If there
// are new filters in the backfill queue, with an earlier StartingBlock, then they
// will get backfilled from there instead.
func (lp *Service) Replay(ctx context.Context, fromBlock int64) error
```
and
```
// ReplayStatus returns the current replay status of LogPoller:
//
// NoRequests - there have not been any replay requests yet since node startup
// Requested - a replay has been requested, but has not started yet
// Pending - a replay is currently in progress
// Complete - there was at least one replay executed since startup, but all have since completed
func (lp *Service) ReplayStatus() ReplayStatus {
```
The  actual behavior of the `run` method remains unchanged, and the `backfillFilters()` method it calls is nearly unchanged, aside from what's been added for monitoring the replay.  It checks whether there is a new replay request before backfilling, setting the replay.pending flag if appropriate so the status can be monitored. It clears the replay flag when the backfilling is complete. The requestBlock is also reset on completion to NoNewReplayRequests, unless there has been a new request for an earlier block while the backfill was happening.

[NONEVM-1260]: https://smartcontract-it.atlassian.net/browse/NONEVM-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -419,10 +419,6 @@ func (fl *filters) UpdateStartingBlocks(ctx context.Context, startingBlock int64
 		}
 		startingBlocks[id] = newStartingBlock
 	}
-	err := fl.orm.UpdateStartingBlocks(ctx, startingBlocks)
-	if err != nil {
-		return err
-	}
 
 	for id, blk := range startingBlocks {
 		fl.filtersByID[id].IsBackfilled = false
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -402,6 +402,36 @@ func (fl *filters) MarkFilterBackfilled(ctx context.Context, filterID int64) err
 	return nil
 }
 
+// UpdateStartingBlocks will update the starting blocks of all backfilled filters to startingBlock
+// and set IsBackfilled=false for them. It will also update the starting blocks of any filters which
+// are already scheduled for backfilling, but only if the new startingBlock is less than the one they
+// currently have set. All filters will then be added to the filtersToBackfill index, so their logs will
+// be (re-)fetched from the specified starting point on the next iteration of the main LogPoller run loop.
+func (fl *filters) UpdateStartingBlocks(ctx context.Context, startingBlock int64) error {
+	fl.filtersMutex.Lock()
+	defer fl.filtersMutex.Unlock()
+
+	startingBlocks := make(map[int64]int64, len(fl.filtersByID))
+	for id, filter := range fl.filtersByID {
+		newStartingBlock := filter.StartingBlock
+		if filter.IsBackfilled || startingBlock < newStartingBlock {
+			newStartingBlock = startingBlock
+		}
+		startingBlocks[id] = newStartingBlock
+	}
+	err := fl.orm.UpdateStartingBlocks(ctx, startingBlocks)
+	if err != nil {
+		return err
+	}
+
+	for id, blk := range startingBlocks {
+		fl.filtersByID[id].IsBackfilled = false
+		fl.filtersByID[id].StartingBlock = blk
+		fl.filtersToBackfill[id] = struct{}{}
+	}
+	return nil
+}
+
 // LoadFilters - loads filters from database. Can be called multiple times without side effects.
 func (fl *filters) LoadFilters(ctx context.Context) error {
 	if fl.loadedFilters.Load() {
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -407,7 +407,7 @@ func (fl *filters) MarkFilterBackfilled(ctx context.Context, filterID int64) err
 // are already scheduled for backfilling, but only if the new startingBlock is less than the one they
 // currently have set. All filters will then be added to the filtersToBackfill index, so their logs will
 // be (re-)fetched from the specified starting point on the next iteration of the main LogPoller run loop.
-func (fl *filters) UpdateStartingBlocks(ctx context.Context, startingBlock int64) error {
+func (fl *filters) UpdateStartingBlocks(startingBlock int64) {
 	fl.filtersMutex.Lock()
 	defer fl.filtersMutex.Unlock()
 
@@ -425,7 +425,6 @@ func (fl *filters) UpdateStartingBlocks(ctx context.Context, startingBlock int64
 		fl.filtersByID[id].StartingBlock = blk
 		fl.filtersToBackfill[id] = struct{}{}
 	}
-	return nil
 }
 
 // LoadFilters - loads filters from database. Can be called multiple times without side effects.
--- a/pkg/solana/logpoller/log_poller.go
+++ b/pkg/solana/logpoller/log_poller.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"iter"
 	"math"
+	"sync"
 	"time"
 
 	"github.com/gagliardetto/solana-go/rpc"
@@ -17,9 +18,10 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
 )
 
+const NoNewReplayRequests = -1
+
 var (
-	ErrFilterNameConflict   = errors.New("filter with such name already exists")
-	ErrMissingDiscriminator = errors.New("Solana log is missing discriminator")
+	ErrFilterNameConflict = errors.New("filter with such name already exists")
 )
 
 type ORM interface {
@@ -30,6 +32,7 @@ type ORM interface {
 	DeleteFilters(ctx context.Context, filters map[int64]Filter) error
 	MarkFilterDeleted(ctx context.Context, id int64) (err error)
 	MarkFilterBackfilled(ctx context.Context, id int64) (err error)
+	UpdateStartingBlocks(ctx context.Context, startingBlock map[int64]int64) (err error)
 	GetLatestBlock(ctx context.Context) (int64, error)
 	InsertLogs(context.Context, []Log) (err error)
 	SelectSeqNums(ctx context.Context) (map[int64]int64, error)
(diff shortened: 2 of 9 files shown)


pkg/solana/logpoller/filters.go
```diff
@@ -393,6 +393,36 @@ func (fl *filters) MarkFilterBackfilled(ctx context.Context, filterID int64) err
 	return nil
 }
 
+// UpdateStartingBlocks will update the starting blocks of all backfilled filters to startingBlock
+// and set IsBackfilled=false for them. It will also update the starting blocks of any filters which
+// are already scheduled for backfilling, but only if the new startingBlock is less than the one they
+// currently have set. All filters will then be added to the filtersToBackfill index, so their logs will
+// be (re-)fetched from the specified starting point on the next iteration of the main LogPoller run loop.
+func (fl *filters) UpdateStartingBlocks(ctx context.Context, startingBlock int64) error {
```
dhaidashenko: ChainReader registers filters with `startingBlock=0`. On filter registration (that occurs on every restart), we set IsBackfilled to false if the new starting block is less than the persisted. So, on restart after replay, we'll trigger another backfill through the whole available history.
Do we need to persist a replay request? Can't we keep it in memory?

pkg/solana/logpoller/log_poller.go
```diff
@@ -201,6 +213,44 @@ func (lp *Service) UnregisterFilter(ctx context.Context, name string) error {
...
+// will get backfilled from there instead.
+func (lp *Service) Replay(ctx context.Context, fromBlock int64) error {
+	ctx, cancel := lp.eng.Ctx(ctx)
+	defer cancel()
+
+	lp.replay.mut.Lock()
+	defer lp.replay.mut.Unlock()
+
+	if lp.replay.requestBlock != NoNewReplayRequests && lp.replay.requestBlock <= fromBlock {
+		// Already requested, no further action required
```
dhaidashenko, line 239: Should we log warning that replay request was ignored?

pkg/solana/logpoller/log_poller.go
```diff
@@ -201,6 +213,44 @@ func (lp *Service) UnregisterFilter(ctx context.Context, name string) error {
...
+}
+
+// ReplayPending returns the current replay status of LogPoller. true indicates there is a replay currently in progress.
+// False means there is no replay in progress. Some subtleties to bear in mind:
+//  1. if a new request has been submitted since the last replay finished, but the backfilling hasn't been started yet,
+//     this will still return false.
+//  2. It may also return false if the node is restarted after a replay request was submitted but before it completes,
+//     even though it will still know to finish backfilling everything needed in order to satisfy the replay request.
+func (lp *Service) ReplayPending() bool {
+	lp.replay.mut.RLock()
```
dhaidashenko, line 260: What's the use case for this method? If it's used by a caller to determine whether replay was done, we need to ensure that it returns true right after the `Replay` was called and until it is fully complete.

pkg/solana/logpoller/log_poller.go
```diff
@@ -248,6 +316,10 @@ func (lp *Service) backfillFilters(ctx context.Context, filters []Filter, to int
 	}
 
 	lp.lggr.Infow("Done backfilling filters", "filters", len(filters), "from", minSlot, "to", to)
+	if replayBlock > 0 {
```
dhaidashenko:
```suggestion
	if replayBlock != NoNewReplayRequests {
```

pkg/solana/logpoller/filters.go
reductionista: > ChainReader registers filters with startingBlock=0

Hmm... this definitely goes against what I thought the plan was for how ChainReader was supposed to behave.

In the ChainReader design doc, I was asked what this parameter should be set to in the ChainReader config and I suggested we make it time-based and set it to 30 days for CCIP, so it matches the behavior on EVM.  I thought I remembered there being general agreement around this in the comments, but I guess at some point that plan must have been discarded in favor of a different one.

I think I'd need to understand more about why we're doing it this way... is the config parameter not time based then? I think the default should be to start at the current block, as it is on EVM, unless that's overridden with a time to have ChainReader subtract from that. Are we putting block numbers directly in the config now?

reductionista: Tagging @ilija42 & @EasterTheBunny as they may have more be able to give more insight into this.

pkg/solana/logpoller/log_poller.go
reductionista: Sure, I can add that

pkg/solana/logpoller/log_poller.go
reductionista: The use case I had in mind was for a cli command that could be used to check whether the node is currently in replay mode, or if it's back to normal operation. So that if a node op triggers a replay via cli, they don't have to go look at grafana logs to understand why they're not seeing progress on chain... if they run the status command, it could tell them it's still replaying so they know not to expect any new transactions to appear yet. This could be particularly helpful for replays that take hours or more.

Come to think of it though... we might want to expand that to include backfilling in general, not just replays.  When a node first starts up, there will be some period where it's backfilling.  Would be good to have an easy way to know when it's done without sifting through logs.

If we wanted to use this in a more programmatic way, rather than just manually on the command line, I agree that it should have more synchronization added. In particularly, I was thinking... the caller could pass a channel to Replay() that gets notified when the replay is done.  But at least for now, it doesn't seem like we have a use case for that so I lean towards saying it's not worth introducing the complexity.

reductionista: Some background context: on evm, lp.Replay was originally a blocking operation which would run until the context it was passed was cancelled... and at that point, the replay itself got cancelled.  This became a problem because there were both client and server timeouts for the http requests used to submit the replay command. With the default timeouts, it would run for a short while and then stop.  You could override the timeouts to be longer, which worked better, but then if anything happened to the terminal where the command was being run, or the connection got broken, that would also cancel the context and the whole operation.  After that we changed it to work in a more asynchronous way, but now it's a bit annoying that there isn't an easy way to know when it's done. So I was thinking for Solana we should have something there... especially because replays might take a lot longer than on evm.

pkg/solana/logpoller/filters.go
reductionista: Regarding replays, I guess we don't necessarily need to persist them since the node op will be more likely to be around to restart the replay if needed.  But letting it do a partial replay that gets aborted by a restart seems a bit messy. And from a UI standpoint, it seems more consistent to have backfills of one filter or all filters work in the same way.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Makes sense. My main concern is that it returns false if the replay was scheduled but not started.

reductionista: > My main concern is that it returns false if the replay was scheduled but not started.

I guess instead of returning a bool, it could return one of 3 values representing REQUESTED / PENDING / COMPLETE ?

reductionista: Might be a better word for COMPLETE, since that state would include "nothing requested yet"

reductionista: How about `ReplayStatus()`, and it returns NONE/REQUESTED/PENDING ?

pkg/solana/logpoller/filters.go
reductionista: Thinking about this more, starting to agree with @dhaidashenko that not persisting replays might be better.  (I still think the ChainReader design should be based on time, and not be setting anything to block 0 but that's a separate issue... which we can fix later if we're all in agreement on it.)

There is a missing feature in my current implementation I'm realizing: there needs to be a way for node ops to cancel a replay if it's running too long. This could be a separate cli command, but the simplest is just to allow the node op to issue a new replay command with a more recent start block.  However, if IsBackfilled is set to false it will think these are new filters that need to be backfilled and avoid skipping over blocks to go to a more recent block.  So I think dropping the persistence feature would be one way of fixing this, so we don't have to add a separate command.

pkg/solana/logpoller/log_poller.go
reductionista: I decided to change it from `ReplayPending()` to `ReplayStatus()`, it can now return one of 4 states in an enum:  NoRequest, Requested, Pending, and Complete

Replacing the pending field of `ReplayInfo` with this `status` field also allowed me to get rid of the `NoNewRequests = -1` constant, which I think is a bit cleaner. Ready for another review!

pkg/solana/logpoller/log_poller_test.go
```diff
@@ -350,3 +351,80 @@ func TestProcess(t *testing.T) {
...
+	t.Parallel()
+	fromBlock := int64(5)
+
+	lp := newMockedLP(t)
+	assertReplayInfo := func(requestBlock int64, status ReplayStatus) {
+		assert.Equal(t, requestBlock, lp.LogPoller.replay.requestBlock)
+		assert.Equal(t, status, lp.LogPoller.replay.status)
+	}
+
+	t.Run("ReplayInfo state initialized propery", func(t *testing.T) {
```
EasterTheBunny: properly

pkg/solana/logpoller/log_poller.go
```diff
@@ -210,6 +225,43 @@ func (lp *Service) UnregisterFilter(ctx context.Context, name string) error {
...
 
+// Replay submits a new replay request. If there was already a new replay request
+// submitted since the last replay completed, it will be updated to the earlier of the
+// two requested fromBlock's. The expectation is that, on the next timer tick of the
+// LogPoller run loop it will backfill all filters starting from fromBlock. If there
+// are new filters in the backfill queue, with an earlier StartingBlock, then they
+// will get backfilled from there instead.
+func (lp *Service) Replay(fromBlock int64) error {
+	lp.replay.mut.Lock()
+	defer lp.replay.mut.Unlock()
```
EasterTheBunny, line 236: I'm not usually a fan of using locks in this way. I tend to think getter and setter methods are safer and consolidates lock access instead of spreading them around a larger surface area. However, if it works for now it's good by me.

dhaidashenko (dismissed):

EasterTheBunny (approved):

dhaidashenko (approved):

jadepark-dev (approved):

PR #1078, opened 2025-02-15
Change account reading err handling and fix get token prices for a non loop call #1079

ilija42:
### Description

PDAs are often not initialised immediately, on EVM this corresponded do method calls that returned empty data. Mimick this behaviour by not returning an error when an account address is not found.

Also fix get token prices for a non loop call

### DIFF
--- a/pkg/solana/chainreader/batch.go
+++ b/pkg/solana/chainreader/batch.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/gagliardetto/solana-go"
-	"github.com/gagliardetto/solana-go/rpc"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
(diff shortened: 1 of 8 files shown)


pkg/solana/chainreader/batch.go
```diff
@@ -85,13 +87,27 @@ func doMethodBatchCall(ctx context.Context, client MultipleAccountGetter, bdRegi
 	}
 
 	// Fetch the account data
+	results := make([]batchResultWithErr, len(batch))
 	data, err := client.GetMultipleAccountData(ctx, keys...)
 	if err != nil {
+		if errors.Is(err, rpc.ErrNotFound) {
```
amit-momin: Would the entire batch call fail if a single account isn't found? I would have though we would need to handle this on a per call basis somewhere below like where we set `ErrMissingAccountData`

ilija42: true, fixed

ilija42: actually the whole call will fail, so the first solution was right, we don't get results for any of the calls and don't know which one failed

ilija42: reverted the change

amit-momin (dismissed):

pkg/solana/chainreader/batch.go
archseer: This has issues: GetMultipleAccountData will return the error if any of the accounts is not found and no account data, so we will get back zeroed data for even valid accounts. I'll implement a fix shortly

archseer (dismissed):

pkg/solana/chainreader/batch.go
ilija42: I just thought that the RPC was being unreasonable with err handling, forgot that  its a wrapper

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1083, opened 2025-02-17
Type fix for ChainSelector and ArgsTransform for commit method in ChainWriter

prashantkumar1982:

### DIFF
--- a/pkg/solana/chainwriter/helpers.go
+++ b/pkg/solana/chainwriter/helpers.go
@@ -81,6 +81,10 @@ func GetValuesAtLocation(args any, location string) ([][]byte, error) {
 			buf := make([]byte, 8)
 			binary.LittleEndian.PutUint64(buf, value)
 			vals = append(vals, buf)
+		case ccipocr3.ChainSelector:
+			buf := make([]byte, 8)
+			binary.LittleEndian.PutUint64(buf, uint64(value))
+			vals = append(vals, buf)
 		case [32]uint8:
 			vals = append(vals, value[:])
 		default:
--- a/pkg/solana/chainwriter/helpers.go
+++ b/pkg/solana/chainwriter/helpers.go
@@ -85,6 +85,8 @@ func GetValuesAtLocation(args any, location string) ([][]byte, error) {
 			buf := make([]byte, 8)
 			binary.LittleEndian.PutUint64(buf, uint64(value))
 			vals = append(vals, buf)
+		case ccipocr3.Bytes32:
+			vals = append(vals, value[:])
 		case [32]uint8:
 			vals = append(vals, value[:])
 		default:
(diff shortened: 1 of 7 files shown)


pkg/solana/chainwriter/transform_registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/chainwriter"
+)
+
+func Test_CCIPCommitTransform(t *testing.T) {
```
silaslenihan: Should we move the [CCIP execute ArgsTransform test](https://github.com/smartcontractkit/chainlink-solana/blob/e15b1333a14ad5d59aa6cc5407570d373cb3604e/pkg/solana/chainwriter/chain_writer_test.go#L861-L923) here or add this test to `chain_writer_test`?

silaslenihan (dismissed):
LGTM

pkg/solana/chainwriter/transform_registry_test.go
amit-momin: I should maybe do both tbh. The test in `chain_writer_test.go` tests this in the context of the entire CW with the TXM. My test here is just a unit test on the specific function. I should probably do both for both functions

pkg/solana/chainwriter/helpers.go
```diff
@@ -77,10 +77,22 @@ func GetValuesAtLocation(args any, location string) ([][]byte, error) {
 			vals = append(vals, value.Bytes())
 		case ccipocr3.UnknownAddress:
 			vals = append(vals, value)
+		case ccipocr3.UnknownEncodedAddress:
+			decoded, err := solana.PublicKeyFromBase58(string(value))
```
prashantkumar1982, line 81: Could we add unit-tests for each new type added to this switch statement?

pkg/solana/chainwriter/transform_registry_test.go
silaslenihan: Yeah that's probably the safest option

amit-momin (dismissed):

pkg/solana/chainwriter/transform_registry_test.go
```diff
@@ -1,59 +1,109 @@
...
 	"github.com/stretchr/testify/require"
 
+	"github.com/smartcontractkit/chainlink-solana/pkg/monitoring/testutils"
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/chainwriter"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
+	clientmocks "github.com/smartcontractkit/chainlink-solana/pkg/solana/client/mocks"
 )
 
-func Test_CCIPCommitTransform(t *testing.T) {
+func Test_CCIPExecuteTransform(t *testing.T) {
```
prashantkumar1982: nit: make the Test_[name] pattern actually match the real function.
So Test_CCIPExecuteArgsTransform and Test_CCIPCommitAccountTransform

silaslenihan (approved):

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1084, opened 2025-02-17
Nonevm 1064/querykey integration tests

EasterTheBunny:
### Description

[NONEVM-1064](https://smartcontract-it.atlassian.net/browse/NONEVM-1064)
 

### Requires Dependencies
- https://github.com/smartcontractkit/chainlink-solana/pull/1044
- https://github.com/smartcontractkit/chainlink-solana/pull/1088
- https://github.com/smartcontractkit/chainlink-solana/pull/1099
- https://github.com/smartcontractkit/chainlink-common/pull/1040

[NONEVM-1064]: https://smartcontract-it.atlassian.net/browse/NONEVM-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/chainreader/event_read_binding.go
```diff
@@ -521,6 +521,10 @@ func valueForPath(from reflect.Value, itemType string) (any, error) {
 
 	switch from.Kind() {
 	case reflect.Pointer:
+		if from.IsNil() {
+			from = reflect.New(from.Type().Elem())
+		}
+
```
reductionista: Ah, yes... this is one thing I was thinking would be necessary for these tests to pass!

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -1394,6 +1467,45 @@ func (it *SolanaChainComponentsInterfaceTester[T]) buildContractWriterConfig(t T
 						},
 						DebugIDLocation: "",
 					},
+					MethodTriggeringEvent: {
+						FromAddress: fromAddress,
+						InputModifications: []commoncodec.ModifierConfig{
+							&commoncodec.PropertyExtractorConfig{FieldName: "Data"},
+							&commoncodec.AddressBytesToStringModifierConfig{
+								Fields: []string{"AccountStruct.AccountStr"},
+							},
```
reductionista, line 1578: As I understand it, the purpose of `InputModifications` is to convert offchain data structures into onchain data structures, right? Since there are no offchain data structures with a "Data" field in them, I don't see how this could make sense.

This looks to me like a workaround for the bug in `normalizeParams`, where we pass:
```
offChain, err := b.codec.CreateType(itemType, true)
```
instead of:
```
offChain, err := b.codec.CreateType(itemType, false)
```

I'm not sure how these types of modifications going from onchain to offchain would be able to co-exist with modifications needed in order to go from offchain to onchain in the same set of `InputModifications`. But even if they can, it seems like an overly complicated and confusing way of doing something and then undoing it... when we never wanted to do it in the first place. Why not just create the right type in the first place, instead of creating the wrong type and then making changes to get it back to the right type?

reductionista: Also, putting them both here and in OutputModifications is redundant. If we're going to apply the OutputModifications to the input every time before we apply any InputModificaitons, then we may as well just apply them directly rather than effectively specifying the OutputModifications twice.

reductionista: Not sure if you found another way of getting the ValueComparator to pass, but removing the InputModifications and changing those two encoding params to false worked for me:
https://github.com/smartcontractkit/chainlink-solana/pull/1096

pkg/solana/logpoller/parser.go
```diff
@@ -387,17 +386,21 @@
...
 	if err != nil {
 		return 0, 0, nil, fmt.Errorf("%w: invalid transaction hash: %s", ErrInvalidCursorFormat, err.Error())
 	}
 
-	return block, int(logIdx), txHash.Bytes(), nil
+	var txHashBytes []byte
+
+	copy(txHashBytes[:], txHash[:])
+
+	return block, int(logIdx), txHashBytes, nil
```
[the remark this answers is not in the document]
EasterTheBunny: Resolved by not casting since we want to surface an `int64` instead of an `int`.

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -1024,20 +1041,83 @@ func (it *SolanaChainComponentsInterfaceTester[T]) buildContractReaderConfig(t T
...
+							&commoncodec.AddressBytesToStringModifierConfig{
+								Fields:             []string{"AccountStruct.AccountStr"},
+								EnablePathTraverse: true,
+							},
+							&commoncodec.ConstrainedBytesToStringModifierConfig{
+								Fields:             []string{"DifferentField", "NestedDynamicStruct.Inner.S"},
+								EnablePathTraverse: true,
+								MaxLen:             32,
 							},
+						},
```
reductionista, line 1070: See comments here about why we should fix the bug rather than requiring the user to work around it with complex InputModificaitons:
https://github.com/smartcontractkit/chainlink-solana/pull/1084/files#r1964301442

It's a simple fix, we're just passing true when it should be false... so it gives back the wrong type.  Intentionally requesting the wrong type and then asking the user to modify it to be the right type is bad UI... doing this in the interface test obscures the problem.

integration-tests/relayinterface/chain_components_test.go
reductionista: To summarize the problem, as I see it:  we're calling `CreateType` with the wrong encoding param, so it gives us the wrong type. Then requiring the user to work around that by adding InputModifications to change it to the type we actually want. Instead of placing the burden on the user to add extra things to the config, we can avoid that by simply requesting the right type to begin with.

Am I missing something?

reductionista: Oh... I just noticed I left these comments at the wrong place on the config, apologies if that has created confusion.  I meant to leave them under the InputModifications for `EventName` and `EventWithFilterName`.

Now that I notice they are also under MethodTriggeringEvent, I'm a little confused about how that's working... maybe I am missing something about what these are supposed to do 🤔

reductionista (dismissed):
LGTM, aside from the tests not being able to run in parallel

PR #1088, opened 2025-02-18
Chainreader/writer bug fixes

EasterTheBunny:
### Description

This PR isolates bug fixes from integration test work.

### DIFF
--- a/pkg/solana/chainwriter/chain_writer.go
+++ b/pkg/solana/chainwriter/chain_writer.go
@@ -113,7 +113,7 @@ func (s *SolanaChainWriterService) parsePrograms(config ChainWriterConfig) error
 				return fmt.Errorf("failed to create codec entry for method %s.%s, error: %w", program, method, err)
 			}
 
-			s.parsed.EncoderDefs[codec.WrapItemType(true, program, method)] = input
+			s.parsed.EncoderDefs[codec.WrapItemType(true, program, methodConfig.ChainSpecificName)] = input
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/solana/chainwriter/chain_writer.go
```diff
@@ -113,7 +113,7 @@ func (s *SolanaChainWriterService) parsePrograms(config ChainWriterConfig) error
 				return fmt.Errorf("failed to create codec entry for method %s.%s, error: %w", program, method, err)
 			}
 
-			s.parsed.EncoderDefs[codec.WrapItemType(true, program, method)] = input
+			s.parsed.EncoderDefs[codec.WrapItemType(true, program, methodConfig.ChainSpecificName)] = input
```
amit-momin: @silaslenihan why did we have to change this to `ChainSpecificName`? As long as it matches with the Encode call below in SubmitTransaction, I thought it was ok to choose whichever name here

silaslenihan: Oh yeah, you're right actually. As long as the chain specific name is being used for `FindDefinitionFromIDL` (which it is) we can use the generic name here

ilija42: In CR chainSpecific name is only used to find things in IDL, everywhere else we use the generic name. I think that makes debugging easier

silaslenihan: Yeah I think that convention makes sense

EasterTheBunny: Done.

reductionista (dismissed):

ilija42 (approved):

chris-de-leon-cll (approved):

PR #1091, opened 2025-02-19
ChainWriter: Specify Ignorable Errors for Optional Lookups

silaslenihan:
### Description
Rather than ignoring any error on an optional lookup, only ignore anticipated ones such as an argument not being found or a lookup table being missing.

### DIFF
--- a/pkg/solana/chainwriter/chain_writer.go
+++ b/pkg/solana/chainwriter/chain_writer.go
@@ -155,7 +155,7 @@ func GetAddresses(ctx context.Context, args any, accounts []Lookup, derivedTable
 	var addresses []*solana.AccountMeta
 	for _, accountConfig := range accounts {
 		meta, err := accountConfig.Resolve(ctx, args, derivedTableMap, client)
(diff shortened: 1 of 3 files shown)


pkg/solana/chainwriter/chain_writer.go
```diff
@@ -167,6 +167,14 @@ func GetAddresses(ctx context.Context, args any, accounts []Lookup, derivedTable
 	return addresses, nil
 }
 
+// These errors are ignorable if the lookup is optional.
+func isIgnorableError(err error) bool {
+	return err == nil ||
```
amit-momin: Wouldn't this always be false because of the condition above? Since we're checking err != nil before?`accountConfig.Optional && err != nil && isIgnorableError(err)`

silaslenihan: Yea true, I'll remove

amit-momin (approved):

prashantkumar1982 (approved):

jadepark-dev (approved):

PR #1093, opened 2025-02-20
Logpoller Observed ORM

DylanTinianov:
### Description
Implemented Observed ORM for Solana Logpoller to enable reusing metrics and some Grafana dashboards between EVM and Solana.

Ticket: https://smartcontract-it.atlassian.net/browse/NONEVM-746


pkg/solana/logpoller/observability.go
```diff
@@ -0,0 +1,179 @@
...
+	})
+}
+
+func (o *ObservedORM) GetLatestBlock(ctx context.Context) (int64, error) {
+	return withObservedQuery(o, "GetLatestBlack", func() (int64, error) {
+		return o.ORM.GetLatestBlock(ctx)
+	})
+}
+
+func withObservedQueryAndResults[T any](o *ObservedORM, queryName string, query func() ([]T, error)) ([]T, error) {
```
reductionista, line 142: I guess this is the same as evm LogPoller, but just noticing it's kind of strange that these with helper functions accept `o *ObservedORM` as their first param instead of being implemented as methods of ObservedORM.

I'm okay with leaving it as-is though, or changing it

reductionista: Oh I see, this is to be able to test them without exporting them

reductionista: (although I think that would still be possible either way?)

reductionista (approved):

reductionista:
I notice we don't have metrics for `HasFilter`, but I think that's fine since that one is simple enough it should never be an issue.

pkg/solana/logpoller/observability.go
DylanTinianov: Let's leave this out of scope for this PR. But I'll keep this in mind!

jadepark-dev (approved):

PR #1094, opened 2025-02-20
Ata support integration test

silaslenihan:


amit-momin (dismissed):

amit-momin (dismissed):

amit-momin (approved):

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -60,7 +149,9 @@ func (s *SolanaChainWriterService) waitForTxFinality(ctx context.Context, transa
 			return fmt.Errorf("context ended while waiting for finality of transaction %s", transactionID)
 		case <-ticker.C:
 			status, err := s.txm.GetTransactionStatus(waitCtx, transactionID)
-			if err != nil {
+			// Allow poll to continue if status not found for transaction since transactions are not immediately broadcasted
+			// Status check could have started before there is a status to report
+			if err != nil && !strings.Contains(err.Error(), "failed to find transaction") {
```
prashantkumar1982, line 154: This behavior from TXM seems wrong.
If a Tx was accepted by the TXM, then it should keep track of it. If it is not yet broadcasted, then the tXM should return the right status, instead of saying failed to find transaction, and let the caller worry about whether it was a broadcast delay, or that TXM doesn't recognize this tx.

Maybe not in this PR, but we should fix this TXM behavior.
cc @amit-momin

prashantkumar1982 (approved):

toblich (approved):

PR #1104, opened 2025-02-25
Custom workaround for OffRampLatestConfigDetails

archseer:
We required a special implementation to be able to do `ocrConfigs[pluginType]` (where pluginType is `0` or `1`).

I included an additional commit that just enables GetTransaction to support the latest transaction format like the rest of the methods.


silaslenihan (dismissed):

pkg/solana/chainreader/batch.go
```diff
@@ -154,6 +158,35 @@ func doMethodBatchCall(ctx context.Context, lggr logger.Logger, client MultipleA
...
+			if err != nil {
+				results[idx].err = err
+				continue
+			}
+
+			config := reflect.Indirect(reflect.ValueOf(output["Ocr3"])).Index(int(ocrPluginType)).Interface()
+			output["OCRConfig"] = config
+
+			// weakdecode so the uint8 field is cast to bool
+			err = mapstructure.WeakDecode(output, results[dataIdx].returnVal)
```
EasterTheBunny: This should probably be the following:

```
err = mapstructure.WeakDecode(output, results[idx].returnVal)
```

`results` should be the same length as `batch` therefore the indexes should match.

EasterTheBunny (changes requested):

pkg/solana/chainreader/batch.go
amit-momin: Just pushed a fix

EasterTheBunny (approved):

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1106, opened 2025-02-25
Added ExtraArgs ComputeLimit to CCIP Execute transform func

silaslenihan:


pkg/solana/chainwriter/transform_registry.go
```diff
@@ -7,19 +7,17 @@ import (
...
 	"github.com/mitchellh/mapstructure"
-	"github.com/smartcontractkit/chainlink-ccip/chains/solana/gobindings/ccip_offramp"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
 type ReportPostTransform struct {
-	ReportContext  [2][32]byte
-	Report         []byte
-	Info           ccipocr3.ExecuteReportInfo
-	AbstractReport ccip_offramp.ExecutionReportSingleChain
```
silaslenihan, line 29: Removed the AbstractReport field here since that field isn't in the off-chain or on-chain report objects, not sure how it got in here in the first place

pkg/solana/chainwriter/transform_registry_test.go
```diff
@@ -76,16 +81,58 @@ func Test_CCIPExecuteArgsTransform(t *testing.T) {
 		args := struct {
 			ReportContext [2][32]uint8
 			Report        []uint8
+			Info          ccipocr3.ExecuteReportInfo
 		}{
 			ReportContext: [2][32]uint8{},
 			Report:        []uint8{},
+			Info: ccipocr3.ExecuteReportInfo{
```
silaslenihan, line 188: @amit-momin I was struggling to understand the purpose of this test, and I'm not sure if my additions interfere with what it's supposed to be testing...

amit-momin: Ya I think this might conflict with what this was testing before. I was purposely trying to exclude a field to test the mapstructure decode we're using now. Before that would fail the `ReportPreTransform` type check we had. I think you can just remove `Report`. That should still test the same thing

silaslenihan: ok, done

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -32,19 +34,43 @@ func FindTransform(id string) (func(context.Context, any, solana.AccountMetaSlic
 
 // This Transform function looks up the token pool addresses in the accounts slice and augments the args
 // with the indexes of the token pool addresses in the accounts slice.
-func CCIPExecuteArgsTransform(ctx context.Context, args any, accounts solana.AccountMetaSlice, tableMap map[string]map[string][]*solana.AccountMeta) (any, solana.AccountMetaSlice, error) {
+func CCIPExecuteArgsTransform(ctx context.Context, args any, accounts solana.AccountMetaSlice, tableMap map[string]map[string][]*solana.AccountMeta) (any, solana.AccountMetaSlice, []txmutils.SetTxConfig, error) {
```
amit-momin: nit: Since these methods have grown recently, maybe we should just call them `CCIPExecuteTransform` and `CCIPCommitTransform` now to keep it a little more generic

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -7,44 +7,72 @@ import (
...
-		return nil, nil, err
+		return nil, nil, []txmutils.SetTxConfig{}, err
+	}
+
+	if len(argsTransformed.Info.AbstractReports) != 1 || len(argsTransformed.Info.AbstractReports[0].Messages) != 1 {
+		return nil, nil, []txmutils.SetTxConfig{}, fmt.Errorf("Expected 1 report with 1 message")
+	}
+
+	// Compute Units: static cu overhead + svmExtraArgsV1.computeUnits + sum_per_token(tokenDestGasAmount)
+	cu := argsTransformed.Info.AbstractReports[0].Messages[0].ExtraArgsDecoded["ComputeUnits"]
```
matYang: Note: need to align with where the extra args are in the data structure when encoded/decoded with family specific transmitter

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -7,44 +7,72 @@ import (
...
+	if cu == nil {
+		return nil, nil, []txmutils.SetTxConfig{}, errors.New("missing ComputeUnits in ExtraArgs")
+	}
+	computeUnits := staticCuOverhead + cu.(uint32)
+
+	for _, token := range argsTransformed.Info.AbstractReports[0].Messages[0].TokenAmounts {
```
matYang: nit: could consider looping over messages, not important tho, happy to keep as is if this becomes asymmetrical with other impl details

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -7,44 +7,72 @@ import (
...
-	TokenIndexes   []byte
+	ReportContext [2][32]byte
+	Report        []byte
+	Info          ccipocr3.ExecuteReportInfo
+	TokenIndexes  []byte
 }
 
-func FindTransform(id string) (func(context.Context, any, solana.AccountMetaSlice, map[string]map[string][]*solana.AccountMeta) (any, solana.AccountMetaSlice, error), error) {
+// TODO: replace with actual value from CCIP on-chain
+const staticCuOverhead = 1000
```
matYang: for a more reasonable placeholder, let's set it to 120,000

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -311,14 +311,15 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
...
+	options := []txmutils.SetTxConfig{}
 	// Transform args if necessary
 	if methodConfig.ArgsTransform != "" {
 		transformFunc, tfErr := FindTransform(methodConfig.ArgsTransform)
 		if tfErr != nil {
 			return errorWithDebugID(fmt.Errorf("error finding transform function: %w", tfErr), debugID)
 		}
 		s.lggr.Debugw("Applying args transformation", "contract", contractName, "method", method)
-		args, accounts, err = transformFunc(ctx, args, accounts, derivedTableMap)
+		args, accounts, options, err = transformFunc(ctx, args, accounts, derivedTableMap)
```
amit-momin: Think we'll need to move this call further up before `GetAddresses` so `ExtraArgsDecoded` is available to resolve the UserAccounts addresses

amit-momin: Maybe worth adding an account to one of the test configs for this too?

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -2,49 +2,105 @@ package chainwriter
...
+		return nil, nil, []txmutils.SetTxConfig{}, err
+	}
+
+	argsTransformed := ReportPostTransform{
+		ReportContext: argsDecoded.ReportContext,
+		Report:        argsDecoded.Report,
+		Info:          argsDecoded.Info,
+	}
+
+	if len(argsTransformed.Info.AbstractReports) != 1 || len(argsTransformed.Info.AbstractReports[0].Messages) != 1 {
```
amit-momin: We didn't want to enforce this right? In case we decide to support multiple messages in the future?

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -2,49 +2,105 @@ package chainwriter
...
-	err := mapstructure.Decode(args, &argsTransformed)
+func CCIPExecuteTransform(ctx context.Context, args any, accounts solana.AccountMetaSlice, tableMap map[string]map[string][]*solana.AccountMeta) (any, solana.AccountMetaSlice, []txmutils.SetTxConfig, error) {
+	var argsDecoded SVMExecCallArgs
+	err := mapstructure.Decode(args, &argsDecoded)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, []txmutils.SetTxConfig{}, err
+	}
+
+	argsTransformed := ReportPostTransform{
```
amit-momin: Looks like the field names are the same, couldn't we keep this like it was before and just decode directly into `ReportPostTransform` rather than `SVMExecCallArgs`?

pkg/solana/chainwriter/transform_registry.go
silaslenihan: yep good catch forgot to remove this

pkg/solana/chainwriter/transform_registry.go
silaslenihan: I think `SVMExecCallArgs` is a better name, because we will be importing this type directly into the ccip contract transmitter, so it doesn't really make sense to have the transform name in there, plus this is more than just a report. We could potentially combine the two types. `SVMExecCallArgs` is currently missing `TokenIndexes` but it could just be not set in the transmitter and sent over to be populated, then `ExtraData` will be discarded when the args are encoded for chain.

pkg/solana/chainwriter/chain_writer.go
silaslenihan: But we need to resolve the accounts first to pass into the transform func no?

pkg/solana/chainwriter/transform_registry.go
silaslenihan: Went ahead and made those changes

pkg/solana/chainwriter/chain_writer.go
amit-momin: Oh duh right. Should we just manually add UserAccounts in args transform also? That way we're just adding all of the remaining accounts in the transform

amit-momin: If that's the only way forward, it might make more sense to change that in my PR

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya agree `SVMExecCallArgs` is probably the better name. Just didn't want to mapdecode into a struct then parse the info into another. The changes look good!

pkg/solana/chainwriter/chain_writer.go
silaslenihan: Don't we have user accounts in the config already? https://github.com/smartcontractkit/chainlink/blob/e97537621be083ea457818a693fb9affc76e3475/core/capabilities/ccip/configs/solana/chain_writer.go#L238-L246 We have to update the URI but this will still work I think?

amit-momin: Sorry think I confused myself with something else. You can ignore this.

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -2,49 +2,89 @@ package chainwriter
...
+	ReportContext [2][32]byte               `mapstructure:"ReportContext"`
+	Report        []byte                    `mapstructure:"Report"`
+	Rs            [][32]byte                `mapstructure:"Rs"`
+	Ss            [][32]byte                `mapstructure:"Ss"`
+	RawVs         [32]byte                  `mapstructure:"RawVs"`
+	Info          ccipocr3.CommitReportInfo `mapstructure:"Info"`
 }
 
-func FindTransform(id string) (func(context.Context, any, solana.AccountMetaSlice, map[string]map[string][]*solana.AccountMeta) (any, solana.AccountMetaSlice, error), error) {
+// TODO: replace with actual value from CCIP on-chain
```
amit-momin: Do we know when we'd get the actual value for this? Or are we moving forward with this overhead

amit-momin (dismissed):

pkg/solana/chainwriter/transform_registry.go
silaslenihan: @matYang mentioend 120,000 would be a reasonable estimate. I am not sure whether we are expecting a more exact value cc: @agusaldasoro

amit-momin: Ok if we end up moving forward with this, think we can drop the TODO.

matYang (dismissed):

pkg/solana/chainwriter/transform_registry.go
matYang: latest number confirmed to be `50,000`

matYang (dismissed):

amit-momin (dismissed):

amit-momin (dismissed):

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -43,25 +61,45 @@ func FindTransform(id string) (func(context.Context, client.MultiClient, any, so
 
 // This Transform function looks up the token pool addresses in the accounts slice and augments the args
```
prashantkumar1982: nit: update comment with latest changes

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -12,25 +11,44 @@ import (
...
+	// DestExecDataDecoded contain token transfer specific extra args.
+	DestExecDataDecoded []map[string]any
+}
+
+type SVMExecCallArgs struct {
+	ReportContext [2][32]byte                `mapstructure:"ReportContext"`
+	Report        []byte                     `mapstructure:"Report"`
+	Info          ccipocr3.ExecuteReportInfo `mapstructure:"Info"`
+	ExtraData     ExtraDataDecoded           `mapstructure:"ExtraData"`
+	TokenIndexes  []byte                     `mapstructure:"TokenIndexes"`
```
prashantkumar1982: In the chainlink repo, I see this defined as:

type SVMExecCallArgs struct {
	ReportContext [2][32]byte                 `mapstructure:"ReportContext"`
	Report        []byte                      `mapstructure:"Report"`
	Info          ccipocr3.ExecuteReportInfo  `mapstructure:"Info"`
	ExtraData     ccipcommon.ExtraDataDecoded `mapstructure:"ExtraData"`
}
https://github.com/smartcontractkit/chainlink/blob/fe50f9203b4d4ce2767e8994992ab645b0649db8/core/capabilities/ccip/ocrimpls/svm_contract_transmitter_factory.go#L35-L40


Why doesn't the TokenIndexes reflect there?

Could you export this type from here or somewhere in chainlink-ccip repo, and make both the chainlink-solana and chainlink repos use the same?

Same for the SVMCommitCallArgs

silaslenihan: We were planning on importing from `chainlink-solana` into `chainlink`  after this goes through. `TokenIndexes` would just be left empty there. I previously suggested having them in `chainlink-ccip` but Will said it would be better to have it in `chainlink-solana` since the types are Solana specific.

pkg/solana/chainwriter/transform_registry.go
silaslenihan: done

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1108, opened 2025-02-26
Fix interface tests for passing params to GetLatestValue

reductionista:
### Description

  [NONEVM-1312](https://smartcontract-it.atlassian.net/browse/NONEVM-1312)

### Requires Dependencies

  - https://github.com/smartcontractkit/chainlink-common/pull/1049

[NONEVM-1312]: https://smartcontract-it.atlassian.net/browse/NONEVM-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


integration-tests/relayinterface/db_setup.sql
```diff
@@ -0,0 +1,36 @@
+-- logpoller db table initialization for tests. loosely based on what's in chainlnk repo
```
EasterTheBunny, line 1: Is there a strategy for keeping these in sync?

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -923,7 +937,13 @@ func (it *SolanaChainComponentsInterfaceTester[T]) buildContractReaderConfig(t T
...
 	pdaDataPrefix = binary.LittleEndian.AppendUint64(pdaDataPrefix, idx)
 	pdaStructDataPrefix := []byte("struct_data")
-	pdaStructDataPrefix = binary.LittleEndian.AppendUint64(pdaStructDataPrefix, idx)
+	pdaStructData1Prefix := make([]byte, len(pdaStructDataPrefix))
+	copy(pdaStructData1Prefix, pdaStructDataPrefix)
+	pdaStructData1Prefix = binary.LittleEndian.AppendUint64(pdaStructData1Prefix, 1)
+	pdaStructData1Prefix = binary.LittleEndian.AppendUint64(pdaStructData1Prefix, idx)
+	pdaStructData2Prefix := make([]byte, len(pdaStructData1Prefix))
+	copy(pdaStructData2Prefix, pdaStructData1Prefix)
+	binary.LittleEndian.PutUint64(pdaStructData2Prefix[11:19], 2)
```
amit-momin, line 946: nit: couldn't we just do the same thing as `pdaStructData1Prefix` here so you don't have to replace specific indexes?
```suggestion
	pdaStructData2Prefix := make([]byte, len(pdaStructDataPrefix))
	copy(pdaStructData2Prefix, pdaStructDataPrefix)
	pdaStructData2Prefix = binary.LittleEndian.AppendUint64(pdaStructData2Prefix, 2)
	pdaStructData2Prefix = binary.LittleEndian.AppendUint64(pdaStructData2Prefix, idx)
```

integration-tests/relayinterface/db_setup.sql
reductionista: My understanding is that as part of the EVM Extraction epic (probably the last step?) the sql migrations for solana will get moved into chainlink-solana. At that point, I'm assuming we would combine them here.

In the meantime, I guess the ideal strategy would be to have chainlink CI run these Solana interface tests, or at least this one. (Not sure if that's something we already do? I know there are some Solana integration tests that get run.)  If that happens, then any time we update the schema in the chainlink repo in a way that breaks this test, we'll immediately be prompted to fix it by updating it here.

reductionista: > In the meantime, I guess the ideal strategy would be to have chainlink CI run these Solana interface tests, or at least this one.

It looks like the only Solana integration test that chainlink CI runs is `TestSolanaOCRV2Smoke`  (via `make test_smoke`):
https://github.com/smartcontractkit/chainlink-solana/blob/develop/Makefile#L93

At present, the purpose of this test schema is just to make sure that LogPoller doesn't error out when the interface tests are run without an external db, as soon as it starts up. Otherwise even the GetLatestValue interface tests wouldn't pass in CI.  We still have to disable any that have a non-trivial dependence on the external db (ie, most of the ones where we search for events).

The only time this schema needs to be updated is if there are code changes to LogPoller in this repo which break the interface tests in this repo's CI (due to the new LogPoller code not being compatible with the test schema), so that will be caught immediately anyway, even if we keep the chainlink CI as-is.

But the real issue is that if someone makes a change to the db schema in chainlink repo, we would like any incompatibility with the LogPoller ORM that uses it to be caught immediately.  That means we should be running something besides the OCRV2 smoke test anyway.  This is already a problem on develop branch. It's not directly related to this test schema but a side effect of breaking the correspondence between Solana LogPoller code and chainlink schema would be the two schemas would also be out of sync.

Ideally, there would be a CCIP smoke test that gets run for this purpose. Until we have that, I think it does make sense to have it run the Solana relay interface tests instead. But there is one problem with that I hadn't thought of in my previous comment: we've completely disabled any of them which depend on an external db, due to them not being able to run in Solana CI.

I guess what we need is a way of disabling them if they're run from this CI, but not if they're run from chainlink CI--I'll look into that.

PR #1109, opened 2025-02-26
codec: fix parsing of `optional` values, add a special type codec for OnRampAddress

archseer:


pkg/solana/codec/solana_test.go
```diff
@@ -30,7 +30,7 @@ func TestNewIDLAccountCodec(t *testing.T) {
 	bts, err := entry.Encode(ctx, expected, testutils.TestStructWithNestedStruct)
 
 	// length of fields + discriminator
-	require.Equal(t, 262, len(bts))
+	require.Equal(t, 263, len(bts))
```
archseer, line 33: This is because we were actually encoding the option types incorrectly

EasterTheBunny (approved):

amit-momin (approved):
Approving to keep this moving but we should add some unit tests for sanity as a follow up

PR #1110, opened 2025-02-26
Track transaction status in TXM prior to broadcast

amit-momin:
- Added pending state to internal transaction states in TXM
- Created new state map to store pending transactions
- Add transactions to in-memory store on `Enqueue`

### DIFF
--- a/pkg/solana/txm/pendingtx.go
+++ b/pkg/solana/txm/pendingtx.go
@@ -22,13 +22,13 @@ var (
 )
 
 type PendingTxContext interface {
-	// New adds a new transaction in Pending state to the storage
+	// New adds a new transaction in AwaitingBroadcast state to the storage
 	New(msg pendingTx) error
 	// AddSignature adds a new signature to a broadcasted transaction in the pending transaction context.
 	// It associates the provided context and cancel function with the signature to manage retry and bumping cycles.
 	AddSignature(cancel context.CancelFunc, id string, sig solana.Signature) error
-	// RevertToPending reverts the transaction's status to Pending. It also removes context and related signatures from storage associated to given tx id if not in finalized or errored state
-	RevertToPending(id string) error
+	// RevertToAwaitingBroadcast reverts the transaction's status to AwaitingBroadcast. It also removes context and related signatures from storage associated to given tx id if not in finalized or errored state
+	RevertToAwaitingBroadcast(id string) error
 	// ListAllSigs returns all of the signatures being tracked for all transactions not yet finalized or errored
 	ListAllSigs() []solana.Signature
 	// ListAllExpiredBroadcastedTxs returns all the txes that are in broadcasted state and have expired for given block number compared against lastValidBlockHeight (last valid block number)
@@ -58,8 +58,8 @@ type PendingTxContext interface {
 	// 	- Processed -> Broadcasted || Not Found
 	// The function returns the transaction ID associated with the signature and a boolean indicating whether a re-org has occurred.
 	IsTxReorged(sig solana.Signature, currentState utils.TxState) (string, bool)
-	// GetReorgTx returns the pendingTx for the given ID if it exists
-	GetReorgTx(id string) (pendingTx, error)
+	// GetPendingTx returns the pendingTx for the given ID if it exists
+	GetPendingTx(id string) (pendingTx, error)
 }
 
 // finishedTx is used to store info required to track transactions to finality or error
@@ -92,7 +92,7 @@ type pendingTxContext struct {
 	cancelBy    map[string]context.CancelFunc
 	sigToTxInfo map[solana.Signature]txInfo
 
-	pendingTxs              map[string]pendingTx  // pending transactions awaiting broadcast
+	queuedTxs               map[string]pendingTx  // pending transactions awaiting broadcast
 	broadcastedProcessedTxs map[string]pendingTx  // broadcasted and processed transactions that may require retry and bumping
 	confirmedTxs            map[string]pendingTx  // transactions that require monitoring for re-org
 	finalizedErroredTxs     map[string]finishedTx // finalized and errored transactions held onto for status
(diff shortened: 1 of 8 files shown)


pkg/solana/txm/pendingtx.go
```diff
@@ -56,33 +57,33 @@ type PendingTxContext interface {
 	// 	- Confirmed -> Processed || Broadcasted || Not Found
 	// 	- Processed -> Broadcasted || Not Found
 	// The function returns the transaction ID associated with the signature and a boolean indicating whether a re-org has occurred.
-	IsTxReorged(sig solana.Signature, currentState txmutils.TxState) (string, bool)
-	// GetPendingTx returns the pendingTx for the given ID if it exists
-	GetPendingTx(id string) (pendingTx, error)
+	IsTxReorged(sig solana.Signature, currentState utils.TxState) (string, bool)
+	// GetReorgTx returns the pendingTx for the given ID if it exists
```
Farber98: Nit: It feels like we're using the word `pending` to refer to two different things. Returning a `pendingTx` (which is not necessarily in a pending state) seems a bit confusing.

Also, the `GetReorgTx` method, without the context of where it's being used, might be misleading since it simply returns a `pendingTx` (the type, not the state 😛) if it exists in the `broadcastedProcessed` or `confirmed` maps.

Do you think it would make sense to find a better name for the `pendingTx` type? It would be a big change, so maybe we could just stick with clear comments instead

pkg/solana/txm/txm.go
```diff
@@ -720,6 +720,20 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
 		return errors.New("error in soltxm.Enqueue: tx is nil pointer")
 	}
+
+	// Use transaction ID provided by caller if set
+	id := uuid.NewString()
+	if txID != nil && *txID != "" {
+		id = *txID
+	}
+
+	_, err := txm.txs.GetTxState(id)
```
Farber98: smart 🤓

pkg/solana/txm/pendingtx_test.go
```diff
@@ -1409,11 +1357,18 @@ func TestPendingTxContext_ListAllExpiredBroadcastedTxs(t *testing.T) {
 	}
 }
 
+func addBroadcastedTxWithSigAndCancel(t *testing.T, txs *pendingTxContext, tx pendingTx, sig solana.Signature, cancel context.CancelFunc) {
```
Farber98, line 1360: 🙌

Farber98 (commented):
Which was the rationale of this change? I like it but want more context to understand better

It improves the overall flow but introduces a "naming collision" that could cause confusions

amit-momin:
> Which was the rationale of this change? I like it but want more context to understand better
> 
> It improves the overall flow but introduces a "naming collision" that could cause confusions

Think we had considered this before but can't remember why we decided against it. The worry is that if someone submits a transaction and immediately starts polling for statuses, they could get a transaction not found error briefly. The TXM usually broadcasts the tx immediately so this probably wasn't an issue in normal testing. But accounting for network congestion and any slow downs, we could see delays extending this window that we return not found which could lead to unexpected behavior upstream.

pkg/solana/txm/pendingtx.go
amit-momin: Ya good point, this sort of bothered me from when we started making in-memory storage changes. I could use this as an opportunity to rename it but agreed it's probably a lot. I also want to minimize line changes for the auditors. What if I just rename the new state from `Pending` to `AwaitingBroadcast`?

Also good point about `GetReorgTx`, I'll revert that name change.

Farber98: Sounds good!

Farber98 (dismissed):

pkg/solana/txm/pendingtx.go
```diff
@@ -536,6 +591,9 @@ func (c *pendingTxContext) OnError(sig solana.Signature, retentionTimeout time.D
 func (c *pendingTxContext) GetTxState(id string) (utils.TxState, error) {
```
prashantkumar1982: The error return type here is a misnomer.
It just means that Tx not found. The calling code too just ingores the error field if error != nil.

Can we use same pattern that a map uses, when you try to access a key that doesn't exist?
Instead of error, have a boolean ok, which will be set to false when not found.

if tx, exists := GetTxState(id); exists {
  // tx already exists
  return tx
}

pkg/solana/txm/txm.go
prashantkumar1982: Ahh, this check is something that didn't exist before!

pkg/solana/txm/txm.go
```diff
@@ -765,6 +773,14 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 		lastValidBlockHeight: txLastValidBlockHeight,
 	}
 
+	err = txm.txs.New(msg)
+	// No-op if transaction already exists for provided ID
+	if err != nil && errors.Is(err, ErrIDAlreadyExists) {
```
prashantkumar1982, line 778: This is just sanity checking right? We already checked if Tx exists above in this function. So this condition likely won't be true.

pkg/solana/txm/txm.go
amit-momin: It did but it was happening further downstream when we were adding the tx to storage after broadcast. I just had to add this here to prevent `Enqueue` from erroring since we're storing it here now.

pkg/solana/txm/txm.go
amit-momin: That's correct. We probably never hit this error since much can't change between our check above and this

prashantkumar1982 (approved):

pkg/solana/txm/pendingtx.go
Farber98: like this

Farber98 (approved):

silaslenihan (approved):

chris-de-leon-cll (approved):

PR #1112, opened 2025-02-27
Updated compute unit limit estimation to use provided limit as a cap

amit-momin:
### Context
The existing compute unit limit estimation would ignore the provided limit and allow any limit to be set to ensure transaction success. This behavior is problematic for some products where the compute unit limits need to be capped to ensure our cost does not exceed what the user has paid for

### Changes
Updated the compute unit limit estimation to compare the estimated limit against the provided limit. If the estimated value (+buffer) exceeds the provided limit, broadcast is prevented to avoid wasted gas and the transaction is marked as fatal to avoid future rebroadcast.

### DIFF
--- a/pkg/solana/txm/txm.go
+++ b/pkg/solana/txm/txm.go
@@ -748,14 +748,11 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	// Perform compute unit limit estimation after storing transaction
 	// If error found during simulation, transaction should be in storage to mark accordingly
 	if cfg.EstimateComputeUnitLimit {
-		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id)
+		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id, cfg.ComputeUnitLimit)
 		if err != nil {
 			return fmt.Errorf("transaction failed simulation: %w", err)
 		}
-		// If estimation returns 0 compute unit limit without error, fallback to original config
-		if computeUnitLimit != 0 {
-			cfg.ComputeUnitLimit = computeUnitLimit
-		}
+		cfg.ComputeUnitLimit = computeUnitLimit
 	}
 
 	msg := pendingTx{
@@ -799,7 +796,7 @@ func (txm *Txm) GetTransactionStatus(ctx context.Context, transactionID string) 
 
 // EstimateComputeUnitLimit estimates the compute unit limit needed for a transaction.
 // It simulates the provided transaction to determine the used compute and applies a buffer to it.
-func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Transaction, id string) (uint32, error) {
+func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Transaction, id string, providedComputeUnitLimit uint32) (uint32, error) {
 	txCopy := *tx
 
 	// Set max compute unit limit when simulating a transaction to avoid getting an error for exceeding the default 200k compute unit limit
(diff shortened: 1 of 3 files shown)


pkg/solana/txm/txm.go
```diff
@@ -840,25 +837,35 @@ func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Trans
...
 	// Add buffer to the used compute estimate
-	computeUnitLimit := bigmath.AddPercentage(new(big.Int).SetUint64(unitsConsumed), EstimateComputeUnitLimitBuffer).Uint64()
+	estimatedComputeUnitLimit := bigmath.AddPercentage(new(big.Int).SetUint64(unitsConsumed), EstimateComputeUnitLimitBuffer).Uint64()
 	// Ensure computeUnitLimit does not exceed the max compute unit limit for a transaction after adding buffer
-	computeUnitLimit = mathutil.Min(computeUnitLimit, MaxComputeUnitLimit)
+	estimatedComputeUnitLimit = mathutil.Min(estimatedComputeUnitLimit, MaxComputeUnitLimit)
+
+	// Validate that the estimated compute unit limit does not exceed the provided limit
+	// If it does, mark the transaction as fatally errored to avoid broadcasting and wasting funds
+	if estimatedComputeUnitLimit > uint64(providedComputeUnitLimit) {
```
amit-momin, line 859: Compute unit limit estimation adds a 10% buffer to the limit to account for any changes between simulation and broadcast. We're validating if the estimated limit + buffer exceeds the provided limit. Is this preferred or should this comparison happen before the buffer is added?

pkg/solana/txm/txm.go
```diff
@@ -748,14 +748,11 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
-		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id)
+		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id, cfg.ComputeUnitLimit)
 		if err != nil {
 			return fmt.Errorf("transaction failed simulation: %w", err)
 		}
-		// If estimation returns 0 compute unit limit without error, fallback to original config
-		if computeUnitLimit != 0 {
-			cfg.ComputeUnitLimit = computeUnitLimit
-		}
+		cfg.ComputeUnitLimit = computeUnitLimit
```
prashantkumar1982, line 755: This cfg  var naming is so confusing.
I got confused this is the Txm.cfg because of same name.
Could we rename to txCfg?

pkg/solana/txm/txm.go
```diff
@@ -748,14 +748,11 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	// Perform compute unit limit estimation after storing transaction
 	// If error found during simulation, transaction should be in storage to mark accordingly
 	if cfg.EstimateComputeUnitLimit {
-		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id)
+		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id, cfg.ComputeUnitLimit)
 		if err != nil {
 			return fmt.Errorf("transaction failed simulation: %w", err)
 		}
-		// If estimation returns 0 compute unit limit without error, fallback to original config
-		if computeUnitLimit != 0 {
```
prashantkumar1982, line 756: was this previous code wrong? Comment says do it if computeUnitLimit is 0, but then code is doing opposite

pkg/solana/txm/txm.go
```diff
@@ -748,14 +748,11 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	// Perform compute unit limit estimation after storing transaction
 	// If error found during simulation, transaction should be in storage to mark accordingly
 	if cfg.EstimateComputeUnitLimit {
-		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id)
+		computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id, cfg.ComputeUnitLimit)
```
prashantkumar1982, line 751: We could directly store it in cfg.ComputeUnitLimit.
cfg.computeUnitLimit, err := txm.EstimateComputeUnitLimit(ctx, tx, id, cfg.ComputeUnitLimit)

And then remove the below assignment

pkg/solana/txm/txm.go
prashantkumar1982: If the caller has provided a limit, we should check the estimation with that limit, without adding buffer.
Buffer should only be added by us in the actual Tx, iff the estimated cu+buffer is less than the limit provided.

pkg/solana/txm/txm.go
```diff
@@ -840,25 +837,35 @@ func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Trans
...
-	computeUnitLimit := bigmath.AddPercentage(new(big.Int).SetUint64(unitsConsumed), EstimateComputeUnitLimitBuffer).Uint64()
+	estimatedComputeUnitLimit := bigmath.AddPercentage(new(big.Int).SetUint64(unitsConsumed), EstimateComputeUnitLimitBuffer).Uint64()
 	// Ensure computeUnitLimit does not exceed the max compute unit limit for a transaction after adding buffer
-	computeUnitLimit = mathutil.Min(computeUnitLimit, MaxComputeUnitLimit)
+	estimatedComputeUnitLimit = mathutil.Min(estimatedComputeUnitLimit, MaxComputeUnitLimit)
+
+	// Validate that the estimated compute unit limit does not exceed the provided limit
+	// If it does, mark the transaction as fatally errored to avoid broadcasting and wasting funds
+	if estimatedComputeUnitLimit > uint64(providedComputeUnitLimit) {
+		// NoFailure provided as the TxErrType since there isn't a metric to increment for this type of error
```
prashantkumar1982, line 860: Didn't follow.
Why can't we create a new TxErrType TxFailComputeLimitExceeded

pkg/solana/txm/txm.go
```diff
@@ -840,25 +837,35 @@ func (txm *Txm) EstimateComputeUnitLimit(ctx context.Context, tx *solanaGo.Trans
...
+		err := txm.txs.OnPrebroadcastError(id, txm.cfg.TxRetentionTimeout(), txmutils.FatallyErrored, NoFailure)
+		if err != nil {
+			return 0, fmt.Errorf("failed to mark tx ID %s as %s: %w", id, txmutils.FatallyErrored, err)
+		}
+		return 0, fmt.Errorf("estimated compute unit limit %d for tx ID %s exceeds the provided limit %d", estimatedComputeUnitLimit, id, providedComputeUnitLimit)
+	}
 
-	return uint32(computeUnitLimit), nil //nolint // computeUnitLimit can only be a maximum of 1.4M
+	return uint32(estimatedComputeUnitLimit), nil //nolint // computeUnitLimit can only be a maximum of 1.4M
```
prashantkumar1982, line 868: Once all checks against providedComputeUnitLimit have passed without using buffer, then we add the buffer here, and ensure the total still remains less than equal to providedComputeUnitLimit.

pkg/solana/txm/txm.go
```diff
@@ -748,14 +748,11 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	// Perform compute unit limit estimation after storing transaction
 	// If error found during simulation, transaction should be in storage to mark accordingly
 	if cfg.EstimateComputeUnitLimit {
```
prashantkumar1982, line 750: So simulating and failing a Tx if it exceeds its ComputeUnitLimit is conditional on this field being set to true.
How do we ensure this field is set to true always for CCIP?

Farber98: I don’t think we currently have a way to do this, we rely on proper configuration. This value is set [here](https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/config/config.go#L36).

If we introduce a config option to specify the product and make conditional decisions based on that, we would still be dependent on the config being set correctly.

Maybe we could wrap the "default config" for each "product name"? For example, the user selects "CCIP" or "DF", and we automatically apply the standard default values we consider appropriate. If they need something different, they can manually override the defaults. Would that make configs easier for users?

pkg/solana/txm/txm.go
Farber98: There's no metric associated with this error [here](https://github.com/smartcontractkit/chainlink-solana/blob/8fcdaa6f597f0352c27cb7c377f8c8ef7567a538/pkg/solana/txm/pendingtx.go#L727-L728)

Why not tracking this type of error with metrics?

pkg/solana/txm/txm.go
Farber98: And if the 10% buffer goes beyond, we should use a % of the buffer that still goes through?

PR #1122, opened 2025-03-10
Ata audit fixes

silaslenihan:
### Description
Fixing the following issues that came up in audit: 
1. ATA support transaction causes the context to timeout because the parent submit transaction context is not expect to live for that long.
a. Change ATA to only wait for confirmation to help. Change [GetTransactionStatus](https://github.com/smartcontractkit/chainlink-solana/blob/c20246255e2b0e3b9e28917e76ab601c1204289a/pkg/solana/txm/txm.go#L785-L788) to return Pending for Processed so we can have a 1 to 1 mapping of Unconfirmed to Confirmed
b. Audit finding: Introduce an exponential backoff for the waitForTxFinality method since polling every second could get resource intensive.
c. Audit finding: Make the waitForTxFinality timeout configurable. We can just use this [config](https://github.com/smartcontractkit/chainlink-solana/blob/c20246255e2b0e3b9e28917e76ab601c1204289a/pkg/solana/config/config.go#L20) for the timeout.

### DIFF
--- a/pkg/solana/chainwriter/ata_creation.go
+++ b/pkg/solana/chainwriter/ata_creation.go
@@ -129,6 +129,7 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
 		return fmt.Errorf("error enqueuing transaction: %w", err)
 	}
 
+	// Wait for ATA transaction to be confirmed (common Unconfirmed maps to Confirmed on Solana)
 	err = s.waitForTxStatus(ctx, ataUUID, types.Unconfirmed)
 	if err != nil {
 		return fmt.Errorf("error waiting for ATA transaction finality: %w", err)
--- a/pkg/solana/chainwriter/ata_creation.go
+++ b/pkg/solana/chainwriter/ata_creation.go
@@ -137,7 +137,7 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
 }
 
 func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transactionID string, desiredStatus types.TransactionStatus) error {
-	waitCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	waitCtx, cancel := context.WithTimeout(ctx, s.config.TxStatusTimeout.Duration())
 	defer cancel()
 
 	backoff := 1 * time.Second
--- a/pkg/solana/chainwriter/ata_creation.go
+++ b/pkg/solana/chainwriter/ata_creation.go
@@ -141,6 +141,7 @@ func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transact
 	defer cancel()
 
 	backoff := 1 * time.Second
+	maxBackoff := 8 * time.Second
 
 	for {
 		select {
@@ -152,7 +153,7 @@ func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transact
 				return fmt.Errorf("error fetching transaction status: %w", err)
 			}
 			switch status {
-			case types.Finalized, types.Unconfirmed:
+			case types.Finalized, types.Unconfirmed, types.Pending:
 				if status >= desiredStatus {
 					// if status is equal to or greater than desired status, return
 					s.lggr.Debug("ATA transaction reached state", "status", status, "transactionID", transactionID)
@@ -166,5 +167,8 @@ func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transact
 			}
 		}
 		backoff *= 2
+		if backoff > maxBackoff {
+			backoff = maxBackoff
+		}
 	}
 }
--- a/pkg/solana/chainwriter/ata_creation.go
+++ b/pkg/solana/chainwriter/ata_creation.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/gagliardetto/solana-go"
 	"github.com/gagliardetto/solana-go/rpc"
@@ -14,7 +13,6 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/tokens"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
-	"github.com/smartcontractkit/chainlink-common/pkg/types"
 
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
 )
@@ -100,17 +98,17 @@ func CreateATAs(ctx context.Context, args any, lookups []ATALookup, derivedTable
 	return createATAInstructions, nil
 }
 
-func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, createATAinstructions []solana.Instruction, methodConfig MethodConfig, contractName, method string, feePayer solana.PublicKey) error {
+func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, createATAinstructions []solana.Instruction, methodConfig MethodConfig, contractName, method string, feePayer solana.PublicKey) (string, error) {
 	if len(createATAinstructions) == 0 {
-		return nil
+		return "", nil
 	}
 	blockhash, err := s.client.LatestBlockhash(ctx)
 	if err != nil {
-		return fmt.Errorf("error fetching latest blockhash: %w", err)
+		return "", fmt.Errorf("error fetching latest blockhash: %w", err)
 	}
 
 	if len(createATAinstructions) > maxAtas {
-		return fmt.Errorf("too many ATAs to create: %d, max allowed: %d", len(createATAinstructions), maxAtas)
+		return "", fmt.Errorf("too many ATAs to create: %d, max allowed: %d", len(createATAinstructions), maxAtas)
 	}
 	ataTx, ataErr := solana.NewTransaction(
 		createATAinstructions,
@@ -118,7 +116,7 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
 		solana.TransactionPayer(feePayer),
 	)
 	if ataErr != nil {
-		return fmt.Errorf("error constructing ATA transaction: %w", err)
+		return "", fmt.Errorf("error constructing ATA transaction: %w", err)
 	}
 	ataUUID := fmt.Sprintf("ATA-%s", uuid.NewString())
 
@@ -126,50 +124,8 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
 
 	// Enqueue ATA transaction
 	if err = s.txm.Enqueue(ctx, methodConfig.FromAddress, ataTx, &ataUUID, blockhash.Value.LastValidBlockHeight); err != nil {
-		return fmt.Errorf("error enqueuing transaction: %w", err)
+		return "", fmt.Errorf("error enqueuing transaction: %w", err)
 	}
 
-	// Wait for ATA transaction to be confirmed (common Unconfirmed maps to Confirmed on Solana)
-	err = s.waitForTxStatus(ctx, ataUUID, types.Unconfirmed)
-	if err != nil {
-		return fmt.Errorf("error waiting for ATA transaction finality: %w", err)
-	}
-	return nil
-}
-
-func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transactionID string, desiredStatus types.TransactionStatus) error {
-	waitCtx, cancel := context.WithTimeout(ctx, s.config.TxStatusTimeout.Duration())
-	defer cancel()
-
-	backoff := 1 * time.Second
-	maxBackoff := 8 * time.Second
-
-	for {
-		select {
-		case <-waitCtx.Done():
-			return fmt.Errorf("context ended while waiting for finality of transaction %s", transactionID)
-		case <-time.After(backoff):
-			status, err := s.txm.GetTransactionStatus(waitCtx, transactionID)
-			if err != nil {
-				return fmt.Errorf("error fetching transaction status: %w", err)
-			}
-			switch status {
-			case types.Finalized, types.Unconfirmed, types.Pending:
-				if status >= desiredStatus {
-					// if status is equal to or greater than desired status, return
-					s.lggr.Debugw("ATA transaction reached state", "status", status, "transactionID", transactionID)
-					return nil
-				}
-				// otherwise keep polling
-			case types.Failed, types.Fatal:
-				return fmt.Errorf("transaction %s failed", transactionID)
-			default:
-				// Keep polling
-			}
-		}
-		backoff *= 2
-		if backoff > maxBackoff {
-			backoff = maxBackoff
-		}
-	}
+	return ataUUID, nil
 }
--- a/pkg/solana/chainwriter/ata_creation.go
+++ b/pkg/solana/chainwriter/ata_creation.go
@@ -15,6 +15,7 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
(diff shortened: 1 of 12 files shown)


pkg/solana/chainwriter/ata_creation.go
```diff
@@ -129,40 +129,44 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
...
+					return nil
+				}
+				// otherwise keep polling
 			case types.Failed, types.Fatal:
 				return fmt.Errorf("transaction %s failed", transactionID)
 			default:
 				// Keep polling
 			}
 		}
+		backoff *= 2
```
amit-momin: Think we should cap the backoff so we don't get stuck behind a big interval since this has implications on the parent context's deadline. Maybe 4 or 8 seconds? What do you think?

silaslenihan: added max of 8 seconds

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -129,40 +129,44 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
...
 			// Allow poll to continue if status not found for transaction since transactions are not immediately broadcasted
 			// Status check could have started before there is a status to report
 			if err != nil && !strings.Contains(err.Error(), "failed to find transaction") {
 				return fmt.Errorf("error fetching transaction status: %w", err)
 			}
 			switch status {
-			case types.Finalized:
-				s.lggr.Debug("ATA transaction finalized", "transactionID", transactionID)
-				return nil
+			case types.Finalized, types.Unconfirmed:
```
amit-momin: Since this method is a little more generalized now, think we should also include `types.Pending` here. Otherwise, calling this method for that status would still wait till Unconfirmed at least

silaslenihan: good point, added that in as well

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -129,40 +129,48 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
...
 				return fmt.Errorf("error fetching transaction status: %w", err)
 			}
 			switch status {
-			case types.Finalized:
-				s.lggr.Debug("ATA transaction finalized", "transactionID", transactionID)
-				return nil
+			case types.Finalized, types.Unconfirmed, types.Pending:
+				if status >= desiredStatus {
+					// if status is equal to or greater than desired status, return
+					s.lggr.Debug("ATA transaction reached state", "status", status, "transactionID", transactionID)
```
amit-momin: nit: 
```suggestion
					s.lggr.Debugw("ATA transaction reached state", "status", status, "transactionID", transactionID)
```

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -129,40 +129,48 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
 		return fmt.Errorf("error enqueuing transaction: %w", err)
 	}
 
-	err = s.waitForTxFinality(ctx, ataUUID)
+	err = s.waitForTxStatus(ctx, ataUUID, types.Unconfirmed)
```
amit-momin: Think a comment here mentioning we're waiting for confirmation might help since the common status type is a misnomer on the CW/TXM side.

pkg/solana/txm/txm.go
```diff
@@ -782,9 +782,9 @@ func (txm *Txm) GetTransactionStatus(ctx context.Context, transactionID string)
 	}
 
 	switch state {
-	case txmutils.Broadcasted:
+	case txmutils.Broadcasted, txmutils.Processed:
 		return commontypes.Pending, nil
-	case txmutils.Processed, txmutils.Confirmed:
+	case txmutils.Confirmed:
 		return commontypes.Unconfirmed, nil
```
archseer, line 866: Maybe rename `commontypes.Unconfirmed` as well? `pending -> confirmed -> finalized`

amit-momin (dismissed):

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -129,40 +129,49 @@ func (s *SolanaChainWriterService) handleATACreation(ctx context.Context, create
...
-func (s *SolanaChainWriterService) waitForTxFinality(ctx context.Context, transactionID string) error {
-	waitCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+func (s *SolanaChainWriterService) waitForTxStatus(ctx context.Context, transactionID string, desiredStatus types.TransactionStatus) error {
+	waitCtx, cancel := context.WithTimeout(ctx, s.config.TxStatusTimeout.Duration())
 	defer cancel()
 
-	ticker := time.NewTicker(1 * time.Second)
-	defer ticker.Stop()
+	backoff := 1 * time.Second
+	maxBackoff := 8 * time.Second
```
archseer: It still seems to me that this should be taken out of `SubmitTransaction`. libocr expects transmit calls to be quick, and `8s` can cause the whole round to expire and be aborted.

silaslenihan: Oh wow, I wasn't aware of this constraint. This seems like a major issue... If we move this out of SubmitTransaction, where would it go? Some solana-specific hook before submit transaction in the CCIP plugin? CC: @amit-momin

pkg/solana/txm/txm.go
silaslenihan: I agree that this naming is confusing... I believe it stems from a misalignment on what the ChainReader defines as Unconfirmed and how the TXM defines it. I think aligning those changes is out of scope for this PR, but I've created [a ticket](https://smartcontract-it.atlassian.net/jira/software/c/projects/BCFR/boards/358?assignee=61a6f090ef18ca00717f84db&selectedIssue=BCFR-1111) for it to address in the future.

pkg/solana/chainwriter/ata_creation.go
prashantkumar1982: What about this?
We add a feature in solana TXM. Any Tx can take a dependency on another Tx, as a pre-requisite. Unless that pre-requisite tx gets to Finalized/Confirmed state, the other one must not be processed. 
Then, the moment we create and enqueue the ATA Tx, we use it's id and add that as a pre-requisite for the real Tx during creation.
Thus we can quickly return the async SubmitTransaction() call, and still ensure ATA is done before the actual Tx.

pkg/solana/txm/txm.go
```diff
@@ -781,6 +782,28 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 		return fmt.Errorf("failed to store new transaction with ID %s: %w", id, err)
 	}
 
+	// If a dependency transaction ID is provided, handle waiting and enqueuing asynchronously.
+	if cfg.DependencyTxID != "" {
+		go func(msg pendingTx, depID string) {
+			err := txm.waitForTxStatus(ctx, depID, types.Unconfirmed)
```
prashantkumar1982: Since we are not on critical path, maybe we could wait for Finalized state?

silaslenihan: Good point, updated.

prashantkumar1982 (dismissed):

pkg/solana/chainwriter/ata_creation.go
```diff
@@ -100,67 +98,34 @@ func CreateATAs(ctx context.Context, args any, lookups []ATALookup, derivedTable
...
 	if ataErr != nil {
-		return fmt.Errorf("error constructing ATA transaction: %w", err)
+		return "", fmt.Errorf("error constructing ATA transaction: %w", err)
 	}
 	ataUUID := fmt.Sprintf("ATA-%s", uuid.NewString())
 
 	s.lggr.Infow("Sending create ATA transaction", "contract", contractName, "method", method)
 
 	// Enqueue ATA transaction
 	if err = s.txm.Enqueue(ctx, methodConfig.FromAddress, ataTx, &ataUUID, blockhash.Value.LastValidBlockHeight); err != nil {
```
amit-momin: Maybe we can double check with CCIP but think we may want to enable compute unit limit estimation for ATA transactions. The default limit might be larger than what we need for the ATA tx.

pkg/solana/txm/prom.go
```diff
@@ -47,4 +47,8 @@ var (
 		Name: "solana_txm_tx_error_sim_other",
 		Help: "Number of transactions that failed simulation with an unrecognized error. Note: tx may still be included onchain",
 	}, []string{"chainID"})
+	promSolTxmDependencyFailTxs = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "solana_txm_tx_error_dependency",
+		Help: "Number of transactions that failed simulation with an unrecognized error. Note: tx may still be included onchain",
```
amit-momin: Good thinking adding this metric! Just a nit that we need to update the `Help` description

pkg/solana/txm/txm.go
```diff
@@ -790,6 +813,41 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 	return nil
 }
 
+func (txm *Txm) waitForTxStatus(ctx context.Context, transactionID string, desiredStatus types.TransactionStatus) error {
+	waitCtx, cancel := context.WithTimeout(ctx, txm.cfg.TxTimeout())
```
amit-momin: Since the transaction is already broadcasted and we're just waiting for confirmation here, I think we should use `TxConfirmTimeout` instead.

silaslenihan: I actually changed it to wait for Finalization not confirmation, I figured the longer timeout may be safer since we don't have one specifically for finalization.

pkg/solana/txm/txm.go
```diff
@@ -790,6 +813,41 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+			if err != nil {
+				return fmt.Errorf("error fetching transaction status: %w", err)
+			}
+			switch status {
+			case types.Failed, types.Fatal:
+				return fmt.Errorf("transaction %s failed", transactionID)
+			default:
+				if status >= desiredStatus {
+					// if status is equal to or greater than desired status, return
+					txm.lggr.Debugw("ATA transaction reached state", "status", status, "transactionID", transactionID)
```
amit-momin: nit: think the message got carried over from the ATA code
```suggestion
					txm.lggr.Debugw("transaction reached state", "status", status, "transactionID", transactionID)
```

pkg/solana/txm/txm.go
```diff
@@ -781,6 +782,28 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
 		return fmt.Errorf("failed to store new transaction with ID %s: %w", id, err)
 	}
 
+	// If a dependency transaction ID is provided, handle waiting and enqueuing asynchronously.
+	if cfg.DependencyTxID != "" {
+		go func(msg pendingTx, depID string) {
+			err := txm.waitForTxStatus(ctx, depID, types.Finalized)
+			if err != nil {
+				txm.lggr.Errorw("dependency transaction did not reach desired state", "dependencyTxID", depID, "error", err)
+				err = txm.txs.OnPrebroadcastError(msg.id, txm.cfg.TxRetentionTimeout(), txmutils.FatallyErrored, TxDependencyFail)
```
amit-momin: In the case of ATA transactions, we were thinking they're probably retryable. If we always mark as `FatallyErrored` here, we would never retry. Maybe it's better to move this state transition into `waitForTxStatus`? Then we can just have the dependent transaction inherit the status whether it's failed or fatal? wdyt?

pkg/solana/txm/txm_internal_test.go
amit-momin, line 1: Could you also add an integration test for this in `txm_integration_test.go` for sanity? More often than not, we've found some weird behavior after running the full flow with a test node.

pkg/solana/txm/txm.go
amit-momin: Even though we don't have to worry about the context anymore, I think it might still be better to only wait for confirmed here. It should help throughput and tbh talking to one of the auditors it seems like re-orgs on confirmed tx never really happen since a super majority validated it.

pkg/solana/chainwriter/ata_creation.go
silaslenihan: Hmm, how would you imagine this working? Would they need to provide a separate limit? I can start a thread with CCIP

pkg/solana/txm/txm.go
silaslenihan: Yeah good idea, but instead of moving the state transition into `waitForTxStatus`, I think it makes sense to just return a state there so it can be handled in any way by the caller.

pkg/solana/chainwriter/ata_creation.go
amit-momin: If they want to explicitly cap the cost of this, I guess that's something they would need to provide. Don't think that's feasible though so I was just thinking we do our best effort to keep costs low and just set this override for this tx. That way we're not overspending on the CU limit for the ATA tx.
```suggestion
	if err = s.txm.Enqueue(ctx, methodConfig.FromAddress, ataTx, &ataUUID, blockhash.Value.LastValidBlockHeight, txmutils.SetEstimateComputeUnitLimit(true)); err != nil {
```

pkg/solana/txm/txm_integration_test.go
```diff
@@ -306,6 +307,61 @@ func TestTxm_Integration_Reorg(t *testing.T) {
...
+		depTx, lastValidBlockHeight := createTransaction(ctx, t, client, senderPubKey, receiverPubKey, amount, true)
+		require.NoError(t, txmInstance.Enqueue(ctx, "", depTx, &depTxID, lastValidBlockHeight))
+
+		status, err := txmInstance.waitForTxStatus(ctx, depTxID, types.Finalized)
+		require.NoError(t, err)
+		require.Equal(t, types.Finalized, status)
+
+		txID := "main-tx-id-success"
+		tx, lastValidBlockHeight := createTransaction(ctx, t, client, senderPubKey, receiverPubKey, amount, true)
+		require.NoError(t, txmInstance.Enqueue(ctx, depTxID, tx, &txID, lastValidBlockHeight, []txmutils.SetTxConfig{txmutils.SetDependencyTxID(depTxID)}...))
```
amit-momin, line 326: I think we should call Enqueue on the second tx before waiting on the status of the first one. This probably simulates how these tx would queue in practice. Otherwise, I don't think this would really test the new wait loop if the dependency tx's status is already populated as Finalized. Same below for the tx fail

silaslenihan: good point, fixed.

amit-momin (dismissed):

amit-momin (dismissed):

pkg/solana/txm/txm.go
```diff
@@ -785,10 +785,14 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
 			if err != nil {
 				txm.lggr.Errorw("dependency transaction did not reach desired state", "dependencyTxID", depID, "error", err)
-				err = txm.txs.OnPrebroadcastError(msg.id, txm.cfg.TxRetentionTimeout(), txmutils.FatallyErrored, TxDependencyFail)
+				errorStatus := txmutils.Errored
+				if status == types.Fatal {
+					errorStatus = txmutils.FatallyErrored
+				}
+				err = txm.txs.OnPrebroadcastError(msg.id, txm.cfg.TxRetentionTimeout(), errorStatus, TxDependencyFail)
 				if err != nil {
 					txm.lggr.Errorw("failed to mark transaction as fatally errored", "id", msg.id, "error", err)
```
prashantkumar1982: nit: failed to mark transaction as errored

prashantkumar1982 (approved):

amit-momin (approved):

PR #1124, opened 2025-03-11
Change GetMultipleAccountData batch result handling to include missing accounts data

ilija42:

### DIFF
--- a/pkg/solana/chainreader/batch.go
+++ b/pkg/solana/chainreader/batch.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/go-viper/mapstructure/v2"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
@@ -35,7 +36,7 @@ var (
 )
 
 type MultipleAccountGetter interface {
-	GetMultipleAccountData(context.Context, ...solana.PublicKey) ([][]byte, error)
+	GetMultipleAccountData(context.Context, ...solana.PublicKey) ([]*rpc.Account, error)
 }
 
 // doMultiRead aggregate results from multiple PDAs from the same contract into one result.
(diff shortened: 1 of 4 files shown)


pkg/solana/chainreader/batch.go
```diff
@@ -143,7 +144,7 @@ func doMethodBatchCall(ctx context.Context, lggr logger.Logger, client MultipleA
 			returnVal: batchCall.ReturnVal,
 		}
 
-		if len(data[idx]) == 0 {
+		if data[idx] == nil || data[idx].Data == nil || data[idx].Data.GetBinary() == nil || len(data[idx].Data.GetBinary()) == 0 {
```
archseer, line 147: `data[idx].GetBinary()` is enough and will handle these nils internally

pkg/solana/chainreader/chain_reader.go
```diff
@@ -794,11 +799,16 @@ func (s *ContractReaderService) handleGetTokenPricesGetLatestValue(
...
 
-	for _, d := range data {
+	for _, accRes := range accountsRes {
 		var wrapper fee_quoter.BillingTokenConfigWrapper
-		// if we got back an empty account then the account must not exist yet, use zero value
-		if len(d) > 0 {
-			if err = wrapper.UnmarshalWithDecoder(bin.NewBorshDecoder(d)); err != nil {
+
+		var data []byte
+		if accRes != nil && accRes.Data != nil && accRes.Data.GetBinary() != nil {
```
archseer, line 806: same here

pkg/solana/chainreader/batch.go
ilija42: Calling data[idx].Data.GetBinary() without checking the first two conditions would panic. Seems like only `data[idx].Data.GetBinary() `== nil can be removed?

EasterTheBunny (approved):

pkg/solana/chainreader/client_wrapper.go
```diff
@@ -25,15 +25,14 @@ func (w *RPCClientWrapper) GetMultipleAccountData(ctx context.Context, keys ...s
 		return nil, err
 	}
 
-	bts := make([][]byte, len(result.Value))
-
-	for idx, res := range result.Value {
+	var accounts []*rpc.Account
+	for _, res := range result.Value {
 		if res == nil || res.Data == nil || res.Data.GetBinary() == nil {
```
amit-momin, line 30: Should we also len check the binary here?
```suggestion
		if res == nil || res.Data == nil || res.Data.GetBinary() == nil || len(res.Data.GetBinary()) == 0 {
```

ilija42: I would leave that up to the caller to decide, don't want to handle it at rpc call level

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1125, opened 2025-03-11
Add support for including logs from reverted transactions

reductionista:
### Description

JIRA: [NONEVM-1387](https://smartcontract-it.atlassian.net/browse/NONEVM-1387)

Logs from reverted transactions will now be included in LogPoller's logs table if `include_reverted` is set on the Filter object passed to `RegisterFilter`.

### Resolves Dependencies
- https://github.com/smartcontractkit/chainlink/pull/16731

[NONEVM-1387]: https://smartcontract-it.atlassian.net/browse/NONEVM-1387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -127,15 +127,15 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
 		}
 
 		fl.removeFilterFromIndexes(*existingFilter)
-	} else {
-		// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
-		if contractFilters, okAddr := fl.filtersByAddress[filter.Address]; okAddr {
-			if similarFilters, okEv := contractFilters[filter.EventSig]; okEv {
-				for id := range similarFilters {
-					if conflicting := fl.filtersByID[id]; conflicting.IncludeReverted != filter.IncludeReverted {
-						return fmt.Errorf("IncludeReverted=%v for filter %v conflicts with IncludeReverted=%v for filter %v",
-							conflicting.IncludeReverted, conflicting, filter.IncludeReverted, filter)
-					}
+	}
+
+	// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
+	if contractFilters, okAddr := fl.filtersByAddress[filter.Address]; okAddr {
+		if similarFilters, okEv := contractFilters[filter.EventSig]; okEv {
+			for id := range similarFilters {
+				if conflicting := fl.filtersByID[id]; conflicting.IncludeReverted != filter.IncludeReverted {
+					return fmt.Errorf("IncludeReverted=%v for filter %v conflicts with IncludeReverted=%v for filter %v",
+						conflicting.IncludeReverted, conflicting, filter.IncludeReverted, filter)
 				}
 			}
 		}
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -131,7 +131,7 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
 		// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
 		if contractFilters, okAddr := fl.filtersByAddress[filter.Address]; okAddr {
 			if similarFilters, okEv := contractFilters[filter.EventSig]; okEv {
-				for id, _ := range similarFilters {
+				for id := range similarFilters {
 					if conflicting := fl.filtersByID[id]; conflicting.IncludeReverted != filter.IncludeReverted {
 						return fmt.Errorf("IncludeReverted=%v for filter %v conflicts with IncludeReverted=%v for filter %v",
 							conflicting.IncludeReverted, conflicting, filter.IncludeReverted, filter)
--- a/pkg/solana/logpoller/filters.go
+++ b/pkg/solana/logpoller/filters.go
@@ -119,9 +119,26 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
 		if existingFilter.IsBackfilled {
 			// if existing filter was already backfilled, but starting block was higher we need to backfill filter again
 			filter.IsBackfilled = existingFilter.StartingBlock <= filter.StartingBlock
+
+			// also trigger a backfill if IncludeReverted is being updated from false to true
+			if !existingFilter.IncludeReverted && filter.IncludeReverted {
+				filter.IsBackfilled = false
+			}
 		}
 
 		fl.removeFilterFromIndexes(*existingFilter)
+	} else {
+		// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
+		if contractFilters, okAddr := fl.filtersByAddress[filter.Address]; okAddr {
+			if similarFilters, okEv := contractFilters[filter.EventSig]; okEv {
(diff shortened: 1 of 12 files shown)


pkg/solana/logpoller/models.go
```diff
@@ -5,18 +5,19 @@ import (
...
+	EventName       string
+	EventSig        EventSignature
+	StartingBlock   int64
+	EventIdl        EventIdl
+	SubkeyPaths     SubKeyPaths
+	Retention       time.Duration
+	MaxLogsKept     int64
+	IsDeleted       bool // only for internal usage. Values set externally are ignored.
+	IsBackfilled    bool // only for internal usage. Values set externally are ignored.
+	IncludeReverted bool
```
dhaidashenko, line 20: We should also take into account `IncludeReverted` in `MatchSameLogs` method to enforce explicit filter removal if this flag is changed.

dhaidashenko (commented):
Should we have some protection from cases when a caller registers two equal filters, but with different names and one of them allows reverted transactions?

pkg/solana/logpoller/log_poller.go
```diff
@@ -146,7 +147,24 @@ func (lp *Service) Process(ctx context.Context, programEvent ProgramEvent) (err
 	}
 
 	var logs []Log
+
 	for filter := range matchingFilters {
+		var revertErr error
+		if blockData.Error != nil {
```
EasterTheBunny, line 151: This block looks like all errors are treated as revert errors. Is there not a special RPC error that we need that captures the `reverted` state?

reductionista: Oh, that's a good point. I was under the impression that the terms "failed transaction" and "reverted transaction" were interchangeable. But apparently, that's only true for evm... in Solana it looks like things such as running out of gas are treated as failed transactions but not reverted transactions.

I guess I should rename reverted to failed everywhere?

pkg/solana/logpoller/log_poller.go
```diff
@@ -146,7 +147,24 @@ func (lp *Service) Process(ctx context.Context, programEvent ProgramEvent) (err
...
+			switch e := blockData.Error.(type) {
+			case *jsonrpc.RPCError:
+				revertErr = fmt.Errorf("JsonRpcError %d: %s", e.Code, e.Message)
+			case error:
+				revertErr = e
+			default:
+				revertErr = fmt.Errorf("unknown type: %v", e)
+				lp.lggr.Warnf("received unknown revert error type %T for log with blockData=%v", revertErr, blockData)
+			}
+			if !filter.IncludeReverted {
```
prashantkumar1982, line 152: nit: Probably better if this check is done at the start of the if block for Error.

if (blockData.Error != nil && filter.IncludeReverted) {

prashantkumar1982 (approved):

prashantkumar1982 (commented):
Sorry, clicked on approve by mistake

prashantkumar1982 (dismissed):
Removing approval

pkg/solana/logpoller/log_poller.go
reductionista: Yeah, I had it that way originally but I was thinking this might help us gather some information about what sorts of types and errors can be returned here (without having to wait for a Chainlink transaction to error out) by logging this message for each new error type we receive:
```
lp.lggr.Warnf("received unknown revert error type %T for log with blockData=%v", revertErr, blockData)
```
I spent a while looking through the solana-go code to figure out what the possiilities are here, but the type is `interface{}` and the comment next to it links to a 404 Not Found file within an archived Rust repository. I figured since we're not using this information yet for anything, at this stage we can just be saving and logging it:
https://github.com/gagliardetto/solana-go/blob/main/rpc/types.go#L185-L187

I guess I could try and run some analysis of actual on-chain data to see what the possibilities are, but I'm not confident we could categorize all of them easily.

pkg/solana/logpoller/log_poller.go
reductionista: Actually that's not quite right.  On evm, reverted transactions are transactions that fail while executing on chain. These are the only type of failed transactions which get included in blocks (although the logs themselves are just discarded on evm). We don't care about transactions that fail before they're even executed.

So "logs from reverted transaction" is an accurate description of what we want in evm terminology.  On Solana, there is a similar difference between transactions that fail before execution and ones that fail on chain. Apparently, there is no such thing at all as a reverted transaction in Solana, but... there doesn't appear to be a suitable substitute term for transactions that fail while executing.  Suggestions from Gemini AI were "runtime failure" or "execution failure" but these seem too verbose. I think anyone familiar with evm terminology should know what is meant here, so I'd lean towards not renaming it.

pkg/solana/logpoller/models.go
reductionista: Could you elaborate on what you mean by this?
> Should we have some protection from cases when a caller registers two equal filters, but with different names and one of them allows reverted transactions?

What's wrong with having two different filters where one matches reverted transactions and one doesn't? Isn't that a valid use case?

> We should also take into account IncludeReverted in MatchSameLogs method to enforce explicit filter removal if this flag is changed.

I think taking into account IncludeReverted in MatchSameLogs would have the opposite effect. Presently, changing the flag forces explicit filter removal (because they are treated as matching, since the flag is ignored), and properly updates the filter with the new value of IncludeReverted. (This is done in one of the tests.). If we changed it to take into account IncludeReverted then it would error out with ErrFilterNameConflict. 

I think the current behavior is what we want.

pkg/solana/logpoller/log_poller.go
reductionista: I pulled some sample blocks from an rpc server and looked over the errors in failed transactions. It appears that they don't look anything like what I was assuming, so I've replaced the error type with *string (so it can be saved as NULL in the error column of the db for success or include a string representation of the error for failure.

The errors I'm seeing all look something like this:
```
      "err": {
        "InstructionError": [
          2,
          {
            "Custom": 6003
          }
        ]
      }
```
Not particularly useful I guess, but we may as well save it. It looks like what happens is it just gets json unmarshalled into the interface{} type, which creates a map with "InstructionError" as the only key. I figure just reversing the marshalling and storing it as a json string makes the most sense.

Also, took your suggestion and moved the continue up to the top of the block; now that I'm more confident about what type of errors there are, I don't think we need it.

prashantkumar1982 (dismissed):

pkg/solana/logpoller/filters.go
```diff
@@ -119,9 +119,26 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter Filter) error {
...
+
+			// also trigger a backfill if IncludeReverted is being updated from false to true
+			if !existingFilter.IncludeReverted && filter.IncludeReverted {
+				filter.IsBackfilled = false
+			}
 		}
 
 		fl.removeFilterFromIndexes(*existingFilter)
+	} else {
+		// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
```
dhaidashenko: We'll end up with two similar filters and different IncludeReverted, if there were two filters and I've changed IncludeReverted for one of them.

dhaidashenko (approved):

jadepark-dev (approved):

PR #1126, opened 2025-03-12
Fixed token account ordering and token indexes

amit-momin:
### Issue
Accounts are expected to be provided in the following [order](https://github.com/smartcontractkit/chainlink-ccip/blob/28a24770d357a70b314f1ca02493bc39b27c7a66/chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs#L62-L75) for each token transfer. We expect ordering like below for multiple transfers:
```
user_token_account1
token_billing_config1
pool_chain_config1
...registry lookup table accounts...
user_token_account2
token_billing_config2
pool_chain_config2
...registry lookup table accounts...
```
but ChainWriter configs are only able to order them like the following:
```
user_token_account1
user_token_account2
token_billing_config1
token_billing_config2
pool_chain_config1
pool_chain_config2
...registry lookup table accounts...
```
### Solution
To be able to order the accounts properly, the logic to append these accounts to the transaction was added to the CCIP execute args transform. This also enables us to properly track the `TokenIndexes` required by our contracts.


pkg/solana/chainwriter/transform_registry.go
```diff
@@ -47,40 +56,71 @@ func CCIPExecuteArgsTransform(ctx context.Context, args any, accounts solana.Acc
...
-			if account.PublicKey == address {
-				if i > 255 {
-					return nil, nil, fmt.Errorf("index %d out of range for uint8", i)
-				}
-				tokenIndexes = append(tokenIndexes, uint8(i)) //nolint:gosec
-			}
-		}
+	var aggregatedMessages []ccipocr3.Message
+	for _, report := range argsTransformed.Info.AbstractReports {
+		aggregatedMessages = append(aggregatedMessages, report.Messages...)
```
silaslenihan, line 69: I believe we can assume there is only going to be one report and one message, but I guess it's fine to have this for flexibility

silaslenihan (commented):
I can envision a config-based solution to this but since were moving away from config anyways I think this is a fine temp solution

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -90,3 +130,81 @@ func CCIPCommitAccountTransform(ctx context.Context, args any, accounts solana.A
...
+		}
+		tokenAdminRegistry := ccip_router.TokenAdminRegistry{}
+		err = bin.NewBorshDecoder(resp.GetBinary()).Decode(&tokenAdminRegistry)
+		if err != nil {
+			return nil, fmt.Errorf("failed to borsh decode token admin registry account: %w", err)
+		}
+
+		// lookup tables can store 256 addresses
+		// token admin registry's WritableIndexes field is the binary representation of indexes that are writable stored in 2 separate uint128 on-chain
+		writableBytes := append(tokenAdminRegistry.WritableIndexes[0].Bytes(), tokenAdminRegistry.WritableIndexes[1].Bytes()...)
```
silaslenihan, line 161: why are there two separate writable index bitmaps?

amit-momin: Not sure why on-chain did it this way but they stored the writable bits in uint128's. Since lookup tables can only have 256 addresses, they have to use 2 [here](https://github.com/smartcontractkit/chainlink-ccip/blob/a8f5bd293fe5b6ecdd177124b29b198096ff2dc2/chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/messages.rs#L82).

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya did this in case we ever enable multiple messages to avoid one less thing to change on our side

pkg/solana/chainwriter/transform_registry.go
archseer: it's to save on space since we only need one bit per account, rather than a whole byte

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -47,40 +56,71 @@ func CCIPExecuteArgsTransform(ctx context.Context, args any, accounts solana.Acc
...
+			perChainTokenConfig, err := getPerChainTokenConfig(sourceChainSelector, destTokenAddress, feeQuoterAddress)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to calculate per chain per token config PDA: %w", err)
+			}
+			poolChainConfig, err := getPoolChainConfig(sourceChainSelector, destTokenAddress, poolProgram.PublicKey)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to calculate pool chain config PDA: %w", err)
+			}
+			// Add token index at which token accounts will be appended
+			tokenIndexes = append(tokenIndexes, uint8(len(accounts))) //nolint:gosec
```
archseer: This is incorrect, take a look at my original commit. The index is an offset from the start of "remaining accounts" (that includes the user specified accounts right before the tokens), so the value will be increasing.

archseer (changes requested):

pkg/solana/chainwriter/transform_registry.go
archseer: e.g. with a list like this:

```
... 14 mandatory accounts
<-- remaining_accounts starts here, this is `0`
0-3 | 4 user accounts for the program call // 0-3
4   | user token account 1 <- tokenIndexes for 1
5   | per chain token config 1
6   | pool chain config 1
7-10| ...4 pool accounts 1
11  | user token account 2 <- tokenIndexes for 2 
12  | per chain token config 2
13  | pool chain config 2
...3 pool accounts 2  
```

`tokenIndexes: would be [4, 11]`

amit-momin: Oh right, missed [this](https://github.com/smartcontractkit/chainlink-ccip/blob/794d0c14d1bba6fe204178dac3c425beb689479e/chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs#L159). Think just updating to this should do the trick then
```suggestion
			tokenIndexes = append(tokenIndexes, uint8(len(accounts) - 14)) //nolint:gosec
```

amit-momin: Pushed the fix in the latest commit

silaslenihan (dismissed):

archseer (approved):

PR #1127, opened 2025-03-12
Adding Priority fees to each command

chray-zhang:
### Description
- Added a mechanism to incorporate Priority fees based off of a block history helped
 - Block History logic taken from TXM (reference: https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/fees/block_history.go)
- Created a retry mechanism if it fails due to congestion

### Testing
Was able to confirm all the priority fees via Solana Explorer and logged them to to check that they existed
Screenshot 2025-03-24 at 11 56 38 AM

Screenshot 2025-03-24 at 11 53 24 AM


gauntlet/packages/gauntlet-solana/src/lib/feeUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // Decode the value (Little Endian)
+  return instructionType === ComputeBudgetInstruction.SetComputeUnitPrice
+    ? Number(data.readBigUInt64LE(1)) // u64 for SetComputeUnitPrice
+    : data.readUInt32LE(1) // u32 for SetComputeUnitLimit
+}
+
+export const parseComputeUnitPrice = (data: Buffer): number | null => {
+  if (data[0] !== ComputeBudgetInstruction.SetComputeUnitPrice) {
+    return
```
augustbleeds: i think it's safer to return null here since that's the return type is. kind of suprised typescript doesn't throw an error here bc I think it'd implicitly return `undefined` as is

gauntlet/packages/gauntlet-solana/src/lib/feeUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+  return parseComputeBudgetInstruction(ComputeBudgetInstruction.SetComputeUnitPrice, data)
+}
+
+export const parseComputeUnitLimit = (data: Buffer): number | null => {
```
augustbleeds, line 86: i think this may be useful, but what do you plan on using this for?

gauntlet/packages/gauntlet-solana/src/lib/feeUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+const COMPUTE_BUDGET_PROGRAM_ID = new PublicKey('ComputeBudget111111111111111111111111111111')
+
+// Instructions are serialized in order in rust
+// https://github.com/solana-labs/solana/blob/60858d043ca612334de300805d93ea3014e8ab37/sdk/src/compute_budget.rs#L25
+enum ComputeBudgetInstruction {
+  SetComputeUnitLimit = 0x02,
+  SetComputeUnitPrice = 0x03,
+}
+
+export const parseBlock = (block: BlockResponse): BlockData => {
```
augustbleeds: nit: rename to parseBlockFees

gauntlet/packages/gauntlet-solana-contracts/src/lib/rdd.ts
```diff
@@ -14,6 +14,7 @@ export enum CONTRACT_TYPES {
 
 function load(network = DEFAULT_NETWORK, path = `${RDD_DIR}/directory-solana-${network}.json`) {
   try {
+    console.log(path)
```
augustbleeds: should we remove this line

gauntlet/packages/gauntlet-solana-contracts/src/commands/contracts/store/acceptFeedOwnership.ts
```diff
@@ -47,8 +47,10 @@ export default class AcceptFeedOwnership extends SolanaCommand {
     const address = contract.programId.toString()
     const program = this.loadProgram(contract.idl, address)
     const state = this.args[0]
+    const signer = this.wallet.publicKey
 
-    const rawTx = await this.makeRawTransaction(this.wallet.publicKey)
+    const rawTx = await this.makeRawTransaction(signer)
+    await this.simulateTx(signer, rawTx)
```
augustbleeds, line 53: so you added `simulateTx` for wherever it was missing? Makes sense.

 I wonder if @aalu1418 or @archseer have context on why we didn't have it before

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -39,6 +42,19 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
 
   static lamportsToSol = (lamports: number) => lamports / LAMPORTS_PER_SOL
 
+  calculatePriceFromLatestBlock = async (): Promise<number> => {
```
augustbleeds: I'd like to see us have a couple of optional flags for all gauntlet commands. 

1. --priority-fee-historical
If the user specifies this flag then it'll use the nth percentile for the history of the block. It could default on 50 percentile (or median as currently defined already). 

Perhaps `--priority-fee-historical=<NUMBER_OF_BLOCKS,PERCENTILE_VALUE>` . You can query the past N blocks and the percentile value could be between 0 and 100. They're comma separated.

To make it usable for all commands you can pass in `this.flags.priority_fee_historical` to the signAndSendRawTx method (or do something similar). Similar to how this.flags.network is used by all commands.

#2 `--priority-fee-constant`
It'd also help to have an option to manually specify a raw value for the priority fee. Imagine that an operator doesn't care about cost but just getting a transaction through at any cost. Something like `--priority-fee-constant` and specify the cost in lamports. 

For the above two flags, only one should be enabled at a time, else you should error.

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -107,7 +124,35 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
     logger.loading('Sending tx...')
-    return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+
+    try {
+      return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+    } catch (error) {
+      // Retry mechanism with greater priority fees
+      console.log('Error type:', error)
+      if (error instanceof SendTransactionError && error.message.includes('congestion') && withPriorityFee) {
+        overrides.price += 1000
```
augustbleeds: I think an exponential increase is better. So perhaps bumping by 50% each time you fail. 

We should make the retry number configurable and the increase configurable as well.

Perhaps `--priority-retry=<BUMP_PERCENTAGE, RETRY_NUMBER>`

augustbleeds (commented):
In addition to the suggestions to create more flags to customize the priority fee functionality, I think we should create a command for easily querying the block history data. 

So you can do yarn gauntlet fee:estimate --priority-fee-historical=5,50 to see the median percentile over the last 5 blocks or something like that

gauntlet/packages/gauntlet-solana/src/lib/feeUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (programId.equals(COMPUTE_BUDGET_PROGRAM_ID)) {
+        const parsedPrice = parseComputeUnitPrice(Buffer.from(bs58.decode(ins.data)))
+        if (parsedPrice !== null) {
+          computeUnitPrice = parsedPrice
+          break // Only one SetComputeUnitPrice instruction per transaction
+        }
+      }
+    }
+
+    if (meta && computeUnitPrice) {
```
augustbleeds: I don't think we need this if statement, we always want to add computePrice to the list even if it is 0 in order to get an accourate sense of the percentile. The go code also defaults to 0 if it couldn't find a set price instruction: https://github.com/smartcontractkit/chainlink-solana/blob/priority-fee/pkg/solana/fees/utils.go#L73 in the current transaction

gauntlet/packages/gauntlet-solana/src/lib/feeUtils.ts
chray-zhang: The Compute Budget program in Solana has two main instructions: 
1. Price -> The actual priority fee you're willing to pay for a tx
2. Limit -> The maximum TOTAL priority fee you're willing to pay for a tx

How They Work Together:
Priority Fee = min(Price × Compute Units Used, Limit)
- If your limit is too low, even with a high price, your transaction may not get the boost you expect.
- If your price is too low, even with a high limit, your transaction may not be competitive enough.

Example Calculation
You set:
Price: 500 lamports per CU
Limit: 50,000,000 lamports (0.05 SOL)
Transaction actually uses: 150,000 compute units
Expected Priority Fee: 500 × 150,000 = 75,000,000 lamports (0.075 SOL)
Final Priority Fee Paid: Since 75,000,000 exceeds your limit of 50,000,000, you only pay 50,000,000.

There's no limit set atm but just added it because I saw a lot of other implementations have both implemented

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
chray-zhang: Flags will need to be camel case and we cannot pass in anything other than a string, so I'll need to parse a tuple

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
chray-zhang: Added with new revision:
Screenshot 2025-03-27 at 8 53 35 PM

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
chray-zhang: Added with new revision:
Screenshot 2025-03-27 at 8 23 41 PM
Screenshot 2025-03-27 at 8 25 37 PM
Screenshot 2025-03-27 at 8 24 15 PM

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -81,18 +108,27 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
 
-    if (!overrides.units && !!this.flags.computeUnits) overrides.units = this.flags.computeUnits
-    if (!overrides.price && !!this.flags.computePrice) overrides.price = this.flags.computePrice
-
+    // Default send with priority fees found using block estimator utils if not set
+    if (!overrides.price && validateHistoricalPriorityFeeInput(this.flags.priorityFeesHistorical)) {
+      const [percentile, nBlocks] = this.flags.priorityFeesHistorical.split(',')
+      overrides.price = await this.calculatePriceFromHistoricalBlocks(nBlocks, percentile)
+      logger.info(
+        `Found ${percentile} percentile priority fees in past ${nBlocks} as: ${overrides.price}. Setting as Priority Fee`,
```
augustbleeds, line 124: you'd want to multiply percentile by 100. So instead of logging 0.5 pecentile it's actually the 50th percentile

augustbleeds: you'd want to multiply percentile by 100. So instead of logging 0.5 pecentile it's actually the 50th percentile

augustbleeds: Also add "blocks" after ${nBlocks} to be explicit

augustbleeds:
```suggestion
        `Found ${parseFloat(percentile)*100}th percentile priority fees in past ${nBlocks} blocks as: ${overrides.price}. Setting as Priority Fee`,
```

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -81,18 +108,27 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+      )
+    } else if (!overrides.price && this.flags.priorityFeesConstant) {
+      overrides.price = this.flags.priorityFeesConstant
```
augustbleeds, line 130: Maybe you should also log here

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -81,18 +108,27 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+    if (this.flags.priorityFeesConstant && this.flags.priorityFeesHistorical) {
+      throw new Error('Cannot have both constant priority fees set and historical priority fees')
+    }
 
-    if (!overrides.units && !!this.flags.computeUnits) overrides.units = this.flags.computeUnits
-    if (!overrides.price && !!this.flags.computePrice) overrides.price = this.flags.computePrice
-
+    // Default send with priority fees found using block estimator utils if not set
+    if (!overrides.price && validateHistoricalPriorityFeeInput(this.flags.priorityFeesHistorical)) {
+      const [percentile, nBlocks] = this.flags.priorityFeesHistorical.split(',')
```
augustbleeds, line 122: the validateHistoricalPriorityFeeInput function should be moved inside of the if statement so that if it fails, you can error and tell the user that their input was invalid rather than just silently failing and trying the transaction with no priority fee.

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -107,7 +143,41 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
     logger.loading('Sending tx...')
-    return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+    try {
+      return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+    } catch (error) {
+      // Retry mechanism with greater priority fees
+      if (
+        error instanceof SendTransactionError &&
+        error.message.includes('congestion') &&
+        validateRetryPriorityInput(this.flags.priorityRetry)
```
augustbleeds, line 154: same thing here. move validation inside of the if statement so you can provide a helpful failure error response if the input is not valid

gauntlet/packages/gauntlet-solana/src/lib/utils.ts
```diff
@@ -1,14 +1,66 @@
...
+
+  // Convert strings to numbers
+  const bumpPercentageNum = parseFloat(bumpPercentage)
+  const numberOfRetriesNum = parseInt(numberOfRetries, 10)
+
+  console.log(bumpPercentage, numberOfRetries)
+  console.log(typeof bumpPercentageNum === 'number') // Check if it's a number
+  console.log(typeof numberOfRetriesNum === 'number') // Check if it's a number
+
+  return !isNaN(bumpPercentageNum) && !isNaN(numberOfRetriesNum) && bumpPercentageNum < 1 && numberOfRetriesNum > 0
```
augustbleeds: bumpPercentage could be more than 1 in theory. You could bump it by 200%, which would mean bumpPercentageNum would be 2.

gauntlet/packages/gauntlet-solana/src/lib/utils.ts
```diff
@@ -1,14 +1,66 @@
...
+export const validateRetryPriorityInput = (val: string): boolean => {
+  if (!val) return false
+
+  const [bumpPercentage, numberOfRetries] = val.split(',')
+
+  // Convert strings to numbers
+  const bumpPercentageNum = parseFloat(bumpPercentage)
+  const numberOfRetriesNum = parseInt(numberOfRetries, 10)
+
+  console.log(bumpPercentage, numberOfRetries)
```
augustbleeds: remove this line?

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -107,7 +143,41 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+      ) {
+        const [percentBump, numberOfRetries] = this.flags.priorityRetry.split(',')
```
augustbleeds, line 155: after validation succeeds, you can print out the new priority fee and how much it was bumped by

ex: "Retry with higher priority fee (x% bump): Y"

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -107,7 +143,41 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
     }
     const signedTx = await this.wallet.signTransaction(tx)
     logger.loading('Sending tx...')
-    return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+    try {
+      return await sendAndConfirmRawTransaction(this.provider.connection, signedTx.serialize())
+    } catch (error) {
+      // Retry mechanism with greater priority fees
+      if (
+        error instanceof SendTransactionError &&
```
augustbleeds: if the tx also remains unconfirmed for a while it is also a priority fee issue. So I think TransactionExpiredTimeoutError should also trigger a priority fee retry

augustbleeds (commented):
couple more comments. It'd also be nice to have a seperate gauntlet command that spits out the percentile priority fee over the last N blocks but that can be done in a follow up PR. we can just focus on getting this through.

Can you also add documentation just for these flags? Maybe in the gauntlet-solana README.md (even though it's empty right now)

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
chray-zhang: That's not the case. Solana web3 js library for some reason will timeout the `confirmTransaction` loop after 60s. I've had transactions pass after like 65s. Don't want to send another transaction if the other one will eventually pass

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
chray-zhang: It logs it by default on line 132

gauntlet/packages/gauntlet-solana/src/commands/internal/solana.ts
```diff
@@ -107,7 +146,39 @@ export default abstract class SolanaCommand extends WriteCommand<TransactionResp
...
+        const signature = error.signature
+        for (let i = 0; i < 3; i++) {
+          const status = await this.provider.connection.getSignatureStatus(signature)
+          if (status.value && status.value.confirmationStatus == 'confirmed') {
+            return signature
+          }
+          // exponential
+          this.sleep(3000 ** i)
+        }
+        throw error
```
augustbleeds: it's still possible for a transaction to expire (not be included) if the priority fee is too low. it's like on EVM if the transaction is stuck in the mempool, but you bump the fee to get it prioritized. in solana, the transaction would just expire though. 

to solve also your case, then we can postpone the priority retry after you've verified that the transaction won't be confirmed? so on line 177

chray-zhang: Removed and will iterate on another PR

augustbleeds (approved):

PR #1132, opened 2025-03-17
fix: return error from BindReaders instead of nil

poopoothegorilla:
This pull request includes changes to improve error handling and control flow in the `pkg/solana/chainreader` package. The most important changes include modifying the return value in the `BindReaders` method and changing the control flow in the `ContractReaderService`'s `Bind` method.

Error handling improvements:

* [`pkg/solana/chainreader/bindings.go`](diffhunk://#diff-c6c6e0830f66d2918730be74524b2200076c999410555b7fb7c720e9cf134701L339-R339): Changed the `BindReaders` method to return the accumulated error instead of `nil` to ensure proper error propagation.

Control flow improvements:

* [`pkg/solana/chainreader/chain_reader.go`](diffhunk://#diff-3d1b0738c31faff52e442c3cd928c34145fa409e2f4a1a291632555e594344c2L361-R361): Modified the `Bind` method in `ContractReaderService` to use `continue` instead of returning `nil` to ensure the loop continues execution after handling an error.


amit-momin (dismissed):

prashantkumar1982 (commented):
Could we have unit-tests for these?
Seem like basic errors that were missed because of a lack of unit-tests

reductionista (dismissed):

pkg/solana/chainreader/bindings_test.go
```diff
@@ -104,3 +107,121 @@ func (_m *mockBinding) QueryKey(
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := &namespaceBinding{
+				name:    tt.fields.name,
+				mu:      tt.fields.mu,
+				readers: tt.fields.readers,
+				bound:   tt.fields.bound,
+			}
+			err := b.BindReaders(tt.args.ctx, tt.args.address)
```
amit-momin: nit: Couldn't we just initialize a context and just pass it in directly here? Also I think we prefer using `tests.Context(t)` for test contexts. That way we can just replace the args struct altogether and pass the address directly for each test above.
```suggestion
			err := b.BindReaders(tests.Context(t), tt.args.address)
```

poopoothegorilla: good call cleaned up a bit

amit-momin (approved):

prashantkumar1982 (approved):

chris-de-leon-cll (approved):

PR #1133, opened 2025-03-17
Add validation of URL in Node config

prashantkumar1982:


amit-momin (dismissed):

pkg/solana/config/config.go
```diff
@@ -166,8 +166,10 @@ func (n *Node) ValidateConfig() (err error) {
 	}
 	if n.URL == nil {
 		err = errors.Join(err, config.ErrMissing{Name: "URL", Msg: "required for all nodes"})
+	} else if n.URL.String() == "" {
```
poopoothegorilla, line 169: nit: any reason for not doing `if n.URL == nil || n.URL.String() == "" {`

poopoothegorilla (dismissed):

pkg/solana/config/config.go
amit-momin: Oh ya good catch, think we want separate conditions because we should use `ErrEmpty` instead of `ErrMissing ` when `n.URL.String() == ""`

pkg/solana/config/config.go
```diff
@@ -166,8 +166,10 @@ func (n *Node) ValidateConfig() (err error) {
...
+		err = errors.Join(err, config.ErrMissing{Name: "URL", Msg: "required for all nodes"})
```
huangzhen1997: As Amit suggested, should we use `config.ErrEmpty` instead of `config.ErrMissing` ?

pkg/solana/config/config.go
prashantkumar1982: Thanks, good catch.
Fixed, and updated tests too

poopoothegorilla (approved):

huangzhen1997 (approved):

silaslenihan (approved):

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1138, opened 2025-03-19
Add IncludeReverted option to ChainReader config EventDef

reductionista:
### Description

  [NONEVM-1388](https://smartcontract-it.atlassian.net/browse/NONEVM-1388)

### Requires Dependencies

  - https://github.com/smartcontractkit/chainlink-solana/pull/1125

### Resolves Dependencies
  - https://github.com/smartcontractkit/chainlink/pull/16834

[NONEVM-1388]: https://smartcontract-it.atlassian.net/browse/NONEVM-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


prashantkumar1982 (dismissed):

ilija42 (commented):
there are log poller failures in CI test runs

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -684,26 +747,36 @@ type helper struct {
...
 }
 
 func (h *helper) Init(t *testing.T) {
 	t.Helper()
 
+	dbURL := sqltest.TestURL(t)
+	h.db = sqltest.NewDB(t, dbURL)
+
+	if dbURL == pg.DriverInMemoryPostgres {
+		h.inMemoryDB = true
```
ilija42, line 761: Why add in memory postgres now?

pkg/solana/config/chain_reader.go
```diff
@@ -47,6 +49,13 @@ func (f PollingFilter) GetStartingBlock() int64 {
 	return *f.StartingBlock
 }
 
+func (f PollingFilter) GetIncludeReverted() bool {
+	if f.IncludeReverted == nil {
+		return DefaultIncludeReverted
+	}
+	return *f.IncludeReverted
+}
+
```
ilija42, line 58: why not use false as default instead of nil?

reductionista: That's what I tried first, but it crashed the codec. Apparently, the codec requires all of these to be pointers in order for it to decode them without panicing.

integration-tests/relayinterface/chain_components_test.go
reductionista: I wouldn't characterize it as adding in-memory postgres. Any time there isn't a db available it always falls back to in-memory postgtres. That's just how `sqltest.TestURL` works.

The `dbURL == pg.DriverInMemoryPostgres` check and the `h.inMemoryDB` flag are added here so that we can selectively disable sub-tests which fail when run with in-memory postgres.  (Due to it lacking some postgres features we use.)

pkg/solana/config/chain_reader.go
ilija42: how is this flag connected to codec?

prashantkumar1982 (approved):

ilija42 (approved):
Approved, but  it would be nice to have the flag as bool

pkg/solana/config/chain_reader.go
reductionista: Sorry, I guess it's not the codec but this ChainReader function it crashes in:
https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/chainreader/chain_reader.go#L920-L949

There's probably a way of updating that code to get it to work. I tried a couple things and didn't succeed in getting it to work, so I decided it wasn't worth spending more time on it.

reductionista: The main thing I remember trying is, replacing this:
```
reflect.Indirect
```
with a conditional check that it's a pointer, only applying the Indirect if it's a pointer. But for some reason it still crashed.

jadepark-dev (approved):

PR #1141, opened 2025-03-20
Bump golang.org/x/net and golang.org/x/crypto to latest versions

prashantkumar1982:
Older versions of these deps are known to have vulnerabilities.


archseer:
Seems to have been resolved in develop?

PR #1150, opened 2025-03-24
Handle case when blocks can't be backfilled due to pruning

prashantkumar1982:
Return empty data when a block wasn't found due to pruning.
Also, updated mockery version.


pkg/solana/logpoller/job_get_block.go
```diff
@@ -68,6 +68,24 @@ func (j *getBlockJob) Run(ctx context.Context) error {
 		},
 	)
 	if err != nil {
+		oldestAvailableSlot, err2 := j.client.GetFirstAvailableSlot(ctx)
+		if err != nil {
```
reductionista:
```suggestion
		if err2 != nil {
```

reductionista:
To fix the CI error, you should be able to just run `make mockery && make generate`. If that doesn't work, try deleting whatever version of mockery is in your path and re-running both.

reductionista:
Also, it looks you may have forgotten to check in `pkg/solana/logpoller/mocks/rpc_client.go`

pkg/solana/logpoller/blocks_sorter.go
```diff
@@ -107,21 +107,18 @@ func (p *blocksSorter) writeOrderedBlocks(ctx context.Context) {
...
+	}
 
-		slotNumber := element.Value.(uint64)
-		block, ok := p.readyBlocks[slotNumber]
-		p.queue.Remove(element)
-		if !ok {
-			// block couldn't be fetched from the RPC. Skip it
-			continue
-		}
-		return &block
```
reductionista: Did these changes to the blocksSorter end up not being necessary? Maybe we should verify somehow that having the getBlock job send a nil block hash won't cause any other issues and results in the overall behavior we want.

prashantkumar1982: Yes, I realized I didn't need to filter out those blocks with a nil BlockHash.
The only place which consumes the fetched blocks is this code, which doesn't even read BlockHash:
https://github.com/smartcontractkit/chainlink-solana/blob/3fa4d72d5441c4ded99b0715ae2b5e559607edfd/pkg/solana/logpoller/log_poller.go#L398-L409

reductionista: Ah, okay... I guess that works!

btw, I liked an earlier idea of yours, that we should be calling that `GetFirstAvailableBlock` method during node startup, while it's figuring out where to pick up where it left off. If you want to add it, this is where it would go:
```
func (lp *Service) getLastProcessedSlot(ctx context.Context) (int64, error) {
	if lp.lastProcessedSlot != 0 {
		return lp.lastProcessedSlot, nil
	}

	latestDBBlock, err := lp.orm.GetLatestBlock(ctx)
	if err == nil {
		return latestDBBlock, nil // <---- HERE
	}
```
(in `log_poller.go`)

Instead of lastProcessedSlot returning latestDBBlock, it could return the max of that and GetFirstAvailableBlock. If you don't add it, we can add it later as a part of the lookback PR

prashantkumar1982: Yes, that's a separate change we can make later, not in this PR

huangzhen1997 (dismissed):

reductionista (approved):

huangzhen1997 (approved):

tt-cll (approved):

jadepark-dev (approved):

ilija42 (approved):

PR #1151, opened 2025-03-26
Return ChainID() that matches the genesis hash, but allow aliases in config validation

archseer:
core ref: 9f7e525871b41e74155f06c2a1f899b6877b77a4


prashantkumar1982 (dismissed):

huangzhen1997 (dismissed):

reductionista (dismissed):

chris-de-leon-cll (dismissed):

prashantkumar1982 (dismissed):

huangzhen1997 (dismissed):

chris-de-leon-cll (dismissed):

chris-de-leon-cll (dismissed):

pkg/solana/client/client.go
```diff
@@ -242,27 +242,22 @@ func (c *Client) ChainID(ctx context.Context) (mn.StringID, error) {
...
-		network = "testnet"
-	case MainnetGenesisHash:
-		network = "mainnet"
-	default:
-		c.log.Warnf("unknown genesis hash - assuming solana chain is 'localnet'")
-		network = "localnet"
+
+	hash := v.(solana.Hash).String()
+
+	// If the genesis hash does not match any known network, we return it as a 'localnet' instead of a genesis hash
```
prashantkumar1982: nit: Also comment that genesis hash of localnet changes on each run, thus it cannot be used as a validation, and instead needs a static name like "localnet".

pkg/solana/chain.go
```diff
@@ -154,13 +152,26 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
...
-	if !ignore && v.chainID != expectedChainID {
-		v.chainIDVerified = false
-		return v.chainIDVerified, fmt.Errorf("client returned mismatched chain id (expected: %s, got: %s): %s", expectedChainID, v.chainID, v.nodeURL)
+	ignore := v.expectedChainID == "localnet"
+
+	if !ignore {
+		var matches bool
+		switch v.chainID {
+		case "devnet":
+			matches = client.DevnetGenesisHash == v.expectedChainID
```
prashantkumar1982: My proposal was that if it is really a problem to simultaneously expect all NOPs and toml configs to be updated with genesis hashes, then we can do this here temporarily:

case "devnet":
			matches = (client.DevnetGenesisHash == v.expectedChainID) || (v.chainID == v.expectedChainID)

.
.
.
similarly other cases.

This way, we handle both cases gracefully and give everyone time to upgrade.

prashantkumar1982: But if we can get everyone to update their configs then this change is good.

huangzhen1997: Actually I think this is wrong, v.chainID is returned from client.ChainID() and it would be gensis hash or localnet.

We should do switch case for v.expectedChainID

huangzhen1997: So in this case, default is handling the gensishash matching, and all other case are for backwards compatibility

prashantkumar1982 (approved):

chris-de-leon-cll (approved):

PR #1158, opened 2025-03-28
CCIP msg.Receiver and user accounts fix

silaslenihan:
core ref: dfeaab49d014466594bb66d5e76a526c011fda2e

### Description
Updates the execute method transform method to handle the logic receiver separately from the `extraArgs.Accounts` list. Instead, the logic receiver is retrieved from `message.Receiver`.


pkg/solana/chainwriter/transform_registry.go
```diff
@@ -89,13 +89,36 @@ func CCIPExecuteArgsTransform(ctx context.Context, client client.MultiClient, ar
...
+		return nil, nil, nil, fmt.Errorf("encountered unexpected number of accounts, expected at least %d, got %d", MandatoryExecuteAccounts, len(accounts))
+	}
+
 	var tokenIndexes []uint8
 	// Accounts below are maintained to be in particular indexes in the Token Admin registry lookup table
 	poolProgram := poolLookupAccounts[2]
 	tokenProgram := poolLookupAccounts[6]
 	// Append token accounts to the account list and track at which index accounts for each token transfer starts
 	for _, message := range aggregatedMessages {
 		receiver := message.Receiver
```
amit-momin: Blocking until this [discussion](https://chainlink-core.slack.com/archives/C0779R3AH8R/p1743214129717669?thread_ts=1742616847.916209&cid=C0779R3AH8R) resolves. If this is changed to the logic receiver, we need to extract the token receiver from the message to properly calculate `userTokenAccount` below.

amit-momin (dismissed):

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -190,6 +168,128 @@ func fetchPoolLookupAccounts(ctx context.Context, client client.MultiClient, poo
...
+	}
+	registryTables, exists := tableMap["PoolLookupTable"]
+	if !exists {
+		return commonTokenTransferAccounts{}, fmt.Errorf("failed to find PoolLookupTable in table map, required for token transfer")
+	}
+	// Fetch all of the accounts in the pool lookup table with the proper IsWritable flag set
+	poolLookupAccounts, err := fetchPoolLookupAccounts(ctx, client, registryTables)
+	if err != nil {
+		return commonTokenTransferAccounts{}, fmt.Errorf("failed to fetch pool lookup accounts and set wrtiable flags, required for token transfer: %w", err)
+	}
```
silaslenihan:
```suggestion
		return commonTokenTransferAccounts{}, fmt.Errorf("failed to fetch pool lookup accounts and set writable flags, required for token transfer: %w", err)
	}
```

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -190,6 +168,128 @@ func fetchPoolLookupAccounts(ctx context.Context, client client.MultiClient, poo
...
+	return computeUnits, nil
+}
+
+func resolveCommonTokenTransferAccounts(ctx context.Context, tokenAccountsRequired bool, client client.MultiClient, toAddress string, args any, tableMap map[string]map[string][]*solana.AccountMeta) (commonTokenTransferAccounts, error) {
+	// Return empty struct if token accounts are not required
+	if !tokenAccountsRequired {
+		return commonTokenTransferAccounts{}, nil
+	}
+	registryTables, exists := tableMap["PoolLookupTable"]
```
silaslenihan, line 194: maybe we should also have a check for the size of the table here (or later when it's indexed)

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -190,6 +168,128 @@ func fetchPoolLookupAccounts(ctx context.Context, client client.MultiClient, poo
...
+	if !exists {
```
silaslenihan, line 195: might want to sanity check that this has is the right size as well. I don't see that done later

amit-momin: `fetchPoolLookupAccounts` that uses this table map is safe against a missing table and the length check later on `poolAccounts` sort of guards against unexpected data. But agree a more explicit check upfront is better

amit-momin (approved):

chris-de-leon-cll (approved):

PR #1159, opened 2025-03-28
Add support for lookback window

reductionista:
### Description

Uses the maximum of any non-permanent retention periods defined on any filters, or 30 days by default if all filters are configured for permanent retention.

This feature allows LogPoller to skip ahead to the beginning of the lookback window if the node has been down for a while, rather than processing every block.
  
  Example:

  [NONEVM-1496](https://smartcontract-it.atlassian.net/browse/NONEVM-1496)

[NONEVM-1496]: https://smartcontract-it.atlassian.net/browse/NONEVM-1496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/solana/logpoller/cat
Farber98, line 1: was this file pushed on purpose?

pkg/solana/logpoller/log_poller.go
```diff
@@ -113,7 +118,7 @@ func NewWithCustomProcessor(lggr logger.SugaredLogger, orm ORM, client RPCClient
 }
 
 func (lp *Service) start(_ context.Context) error {
-	lp.eng.GoTick(services.NewTicker(time.Second), func(ctx context.Context) {
+	lp.eng.GoTick(services.NewTicker(lp.blockTime), func(ctx context.Context) {
```
dhaidashenko: Do we need to tick every block? CCIP only needs finalized blocks, so it takes `12s + tick delay` to observe a new log. Reduction from 1s to 600 ms gives `1-12.6/13 = 3%` of improvement in time, but the number of requests increases from 1*x requests per second to `1/0.6 = 1.6666` requests per second i.e. by 66%.

pkg/solana/logpoller/log_poller.go
```diff
@@ -510,3 +513,32 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+	if lookback == 0 { // Use default lookback if all filters have permanent retention or less than
+		lookback = DefaultLookbackWindow
+	}
+
+	firstAvailableSlot, err := lp.client.GetFirstAvailableBlock(ctx) // Despite the name, this returns slot # not block number
+
+	// nolint:gosec
+	// G115: integer overflow conversion uint64 -&gt; int64
+	latestSlot, firstSlot := int64(latestFinalizedSlot), int64(firstAvailableSlot)
+	lookbackSlot := latestSlot - int64((lookback-1)/lp.blockTime) - 1 // = latestSlot - ceil(lookback/blockTime)
```
dhaidashenko, line 552: Arent division for integers already ceil? `int(10/3) = 3`

pkg/solana/logpoller/log_poller.go
```diff
@@ -510,3 +513,32 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+
+	// This is an optimization to avoid requesting pruned blocks. If there's an err we can just ignore
+	if err == nil && firstSlot > lookbackSlot {
```
dhaidashenko: nit (feel free to ignore): it would nice to move call get get GetFirstAvailableBlock closer to the actual error handling

pkg/solana/logpoller/log_poller.go
```diff
@@ -282,31 +287,29 @@ func (lp *Service) ReplayStatus() ReplayStatus {
...
+	if err != nil {
+		if !errors.Is(err, sql.ErrNoRows) {
+			return 0, fmt.Errorf("error getting latest block from db: %w", err)
+		}
+		lastProcessed = 0
 	}
 
-	latestFinalizedSlot, err := lp.client.SlotHeightWithCommitment(ctx, rpc.CommitmentFinalized)
+	var lookbackBlock int64
+	lookbackBlock, err = lp.computeLookbackWindow(ctx)
```
dhaidashenko: Do we need to calculate lookbackBlock on every iteration?

pkg/solana/logpoller/cat
reductionista: lol, nope!

pkg/solana/logpoller/log_poller.go
reductionista: They're floor. We want `ceil(10/3) = 4`

pkg/solana/logpoller/log_poller.go
reductionista: I don't think it matters all that much either way. If you'd prefer it be `2*lp.blockTime` or something else I'm happy to change it.  I just think it's a good to have it based on the block time, so later if we add more chains that have a different block rate we can make it configurable.

> 1*x requests per second to 1/0.6 = 1.6666

That's if you only count the GetSignaturesForAddress calls.  We still make a separate GetBlock call for every block regardless of how often we check, right?  And even for GetSignatureForAddress, it's the same bandwidth transferred either way... just broken up into two smaller transfers instead of one larger slower request. I don't feel strongly either way in this case, but I think in general it's usually better to spread out the requests evenly rather than save them up--so you don't have a burst of many rpc requests at once followed by a long period of idleness.

pkg/solana/logpoller/log_poller.go
reductionista: Ok, I'll add a cache for it

reductionista: Actually, I'm not sure there's much we could do. We have to find the latest finalized block and first available slot from the chain each time, because we have no idea how much either of those may have changed since last time. It's returning a number that just combines those results with MaxRetention... which is already cached.

Did you have in mind a way we could improve this?

reductionista: I guess we could estimate what the present block number is, based on the block time and elapsed time since the last call?

pkg/solana/logpoller/log_poller.go
reductionista: Moved!  (My original motivation was to group all of the integer conversion together so I didn't need two different lint exclusions, but I agree it ended up making the logic a lot more confusing to read.)

pkg/solana/logpoller/log_poller.go
dhaidashenko: > And even for GetSignatureForAddress, it's the same bandwidth transferred either way... just broken up into two smaller transfers instead of one larger slower request.

`GetSignatureForAddress` returns a 1k of the latest transactions. So the request size won't change.
Unless we expect every block to contain a transaction, the number of GetBlocks should not fluctuate too much from request to request.

`2*lp.blockTime` sounds good to me, but I also won't block te if you leave it as is `lp.blockTime`.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Can we calculate it once on the startup? Otherwise, for every iteration we are making 2 additional RPC requests. 
My understanding is that lookBack is used as an optimisation to jump forward and only fetch data for `[max(lookbackBlock, lastProcessed), latestFinalized]`. But `lookbackBlock > lastProcessed` only if Node was down for some time or we can't keep up with the block rate or if a filter was removed.
In the first case, it's sufficient to calculate lookbackBlock once on startup.
In the second we should optimise LogPoller, otherwise DON might miss msgs.
For the third, if we need to handle it, we can have a flag.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Oh, you are right they are floor. But `int64((lookback-1)/lp.blockTime) - 1`, where `lookback = 10` and `blockTime = 3` results in 2 .
You probably meant `int64((lookback-1)/lp.blockTime) + 1`

pkg/solana/logpoller/log_poller.go
```diff
@@ -525,3 +528,38 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+	if err != nil {
+		return 0, fmt.Errorf("error getting latest slot from RPC: %w", err)
+	}
+
+	if latestFinalizedSlot == 0 {
+		return 0, fmt.Errorf("latest finalized slot is 0 - waiting for next slot to start processing")
+	}
+
+	lookback := lp.filters.MaxRetention()
+	if lookback == 0 { // Use default lookback if all filters have permanent retention
```
dhaidashenko: Should not we use default lookback if at least one filter has perm retention?

pkg/solana/logpoller/log_poller.go
```diff
@@ -24,6 +24,8 @@ var (
 	ErrFilterNameConflict = errors.New("filter with such name already exists")
 )
 
+const DefaultLookbackWindow = 30 * 24 * time.Hour
```
prashantkumar1982: Why is the default 30 days?
We expect our RPCs to only have ~8 days of history.
I'm worried that if a NOP uses a3rd party RPC with more historical lookback, then the node will unnecessarily try to go back 30 days on initial start or long offline period.

pkg/solana/logpoller/log_poller.go
```diff
@@ -525,3 +528,38 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+	latestFinalizedSlot, err := lp.client.SlotHeightWithCommitment(ctx, rpc.CommitmentFinalized)
+	if err != nil {
+		return 0, fmt.Errorf("error getting latest slot from RPC: %w", err)
+	}
+
+	if latestFinalizedSlot == 0 {
+		return 0, fmt.Errorf("latest finalized slot is 0 - waiting for next slot to start processing")
+	}
+
+	lookback := lp.filters.MaxRetention()
```
prashantkumar1982: Why use MaxRetention for lookback window computation?
The way I understand, the MaxRetention is for how long to persist a log in the db right?

And Lookback is just how far in history should we look for logs.
I think this value should be set to default of ~8 days, as we aren't asking Solana RPCs to save more history than that.
Further, we should make this Lookback history a TOML setting. In case of entire DON wide downtimes etc, when we ask NOPs to manually connect to archive nodes, for the downtime duration, we would want to set a custom lookback window.

pkg/solana/logpoller/log_poller.go
prashantkumar1982: We should make this a TOML setting that can be overridden, and initialize the default to 8 days

pkg/solana/logpoller/log_poller.go
reductionista: Changing it to 2*lp.blockTime is trivial, so I don't mind making that change.

But this doesn't sound right to me:
> GetSignatureForAddress returns a 1k of the latest transactions. So the request size won't change.

1k is just a maximum limit on how much data can come back, it would only be relevant if there had been more than 1000 transactions since the previous timer tick. As I understand it, we always pass lastProcessed + 1 as the minimum slot # to start at... so the size of the request is determined by how frequently we send a request. Either way, same amount of data comes back over the long run... polling interval just determines how much of it is fetched each time.

pkg/solana/logpoller/log_poller.go
reductionista: > But int64((lookback-1)/lp.blockTime) - 1, where lookback = 10 and blockTime = 3 results in 2 .

That's different from what's in the code.  I think you're forgetting about the - sign in front of the int64. It's:
```
latestSlot -  int64(lookback-1)/lp.blockTime) - 1 
```
So for your example:  `latestSlot - 3 - 1 = latestSlot - 4`

reductionista: > You probably meant int64((lookback-1)/lp.blockTime) + 1

Yes, in that it's lookbackSlot minus this

pkg/solana/logpoller/log_poller.go
reductionista: Hmm, I suppose that might be good enough.

The only scenario I can think of where this might cause a problem is if you had ccip jobs running on multiple chains. For example, let's say you had connections to both testnet and mainnet rpc servers.  If someone takes down the testnet rpc servers for a month and then brings them back up, you might want the node to be able to catch up without having to disrupt operations on mainnet by restarting the whole thing.

That may not be a very likely scenario, but in the future if we support some Solana L2 chains it might become more of a realistic scenario (eg, wanting it to automatically catch up on an L2 without disrupting mainnet, and wanting to be able to transfer funds between the two chains).

Curious what @prashantkumar1982 thinks of this suggestion. Any other potential downsides of only allowing the node to skip ahead if it's being restarted?

pkg/solana/logpoller/log_poller.go
reductionista: You're right that they are logically two separate parameters. During ChainReader design it just seemed like there were so many different configurable time windows it didn't seem worth making these separately configurable.

Overall, there are at least 4 time windows relevant to operation:
- CCIP plugin's configurable lookback window (set to 8 hours, may need to be overridden to be longer for an emergency recovery).
- RPC Server's retention policy (configured based on disk space, but effectively sets the available history depth)
- LogPoller's per-event retention policy (defaults to 30 days)
- LogPoller's lookback window (configurable per event, determines how far back we want to look for a particular event type

The last two are close enough that combining them seems sensible, to me at least. The number of edge cases where someone might want to configure them independently seems small and the consequences of not being able to pretty minor. On the plus side, it simplifies both the UX and the design/development.

8 days would probably be fine for the lookback, but I think the argument in favor of having more like 30 days is pretty similar for both LogPoller retention and LogPoller lookback: it makes it easier to recover in some emergency situations.  If there's some sort of global network problem that lasts 10 days, then at least the nodes connected to an archive node might be able to recover without missing any logs.  If a majority of nodes miss a particular CCIP sequence number then it becomes impossible to recover without overriding the CCIP plugin's 8 hour window on all nodes.

Since that's admittedly an unlikely edge case to worry about, whether it's worth making them separately configurable probably boils down more to whether there would be any significant difference in startup delay between the two on an archive node.  I guess I would lean toward thinking if someone has gone to the trouble of connecting to an archive node then we might as well allow them to take advantage of the extra available history if they can, to have some extra resilliency at the cost of a slower startup. But if it's going to make a huge difference in startup time it may not be worth it. Curious what @dhaidashenko thinks about this question, since he's done actual performance testing and knows what these delays would be better than I.

reductionista: > Further, we should make this Lookback history a TOML setting. In case of entire DON wide downtimes etc, when we ask NOPs to manually connect to archive nodes, for the downtime duration, we would want to set a custom lookback window.

This is something I've been wondering about actually. When we designed ChainReader, the original intention/assumption was that it would be easier to update CR config than TOML config... since CR config doesn't require a node restart and can be done per job while changing toml config requires restarting the whole node.  But that was when we envisioned the config was a local json file that gets read from the filesystem and overrides what's in the job spec. I'm not too sure how or where the config is stored in the case of workflows (on chain?) Is it something that's difficult for a node op to update? If so we may need to rethink some other parameters as well (eg. StartingBlock) And default retention if that ends up being separate from default lookback.

pkg/solana/logpoller/log_poller.go
reductionista: Good point.  I wouldn't want it to use default lookback if there are some filters with a retention set for > default, eg 100 day retention. But if all of them were either permanent or less than the default yeah... I think that makes sense.

pkg/solana/logpoller/log_poller.go
dhaidashenko: Yes, we always pass lastProcessed + 1 as the minimum slot, but it's not a filter.
```
MinContextSlot: &f.to, // MinContextSlot is not filter. It defines min slot that RPC is expected to observe to handle the request
```
[Solana RPC docs](https://solana.com/docs/rpc/http/getsignaturesforaddress)
```
minContextSlot - The minimum slot that the request can be evaluated at
```

pkg/solana/logpoller/log_poller.go
dhaidashenko: Yes, you are right. I've missed the - sign in front of the int64

pkg/solana/logpoller/log_poller.go
dhaidashenko: > But if it's going to make a huge difference in startup time it may not be worth it.
It'll only make a difference on startup after a downtime and highly depends on number of events. Load tests showed that:
* If the events rate was 10 events/s, 8 hours of history, we'll process in 3 hours
* If events rate was 1.5 events/s, 8 hours - 13 minutes.
 So I agree with Prashant; it makes sense to make it configurable and reduce the default to 8h.

pkg/solana/logpoller/log_poller.go
prashantkumar1982: > If someone takes down the testnet rpc servers for a month and then brings them back up
I don't think that's a practical scenario we should think ok.

Imo, lets keep it simple. Lookback is only for the time when the Node is restarted, or brought up first time. So good to do it at startup once, and not repeatedly.

pkg/solana/logpoller/log_poller.go
prashantkumar1982: Yes, the reason I want them different is what @dhaidashenko pointed to. Catching up a new node to older logs is a slow process. In most practical situations, nodes are restarted individually, not the whole DON. So a single node shouldn't be forced to lookback to 30 days just because it was down for long, or is a new node starting from scratch. It will just make it tooo slow for it to come up to speed.
The default lookback window probably should be ~1 day or less.

In case there's an outage network-wide, and we want to reset all nodes, then we will ask NOPs to change this config to X days, whatever's ideal in that particular outage, and restart their nodes. In that situations, nodes taking hours to just sync till lookback is ok.
But it is not ok to force that for all new nodes or restarting nodes in regular situations.

The Logs Retention is a different thing, which says, for how long do we want these logs to be saved in local DB. In most cases, this duration could be very long, like 30 days. We don't want to tie this to lookback window.

So it is also very important to make the Lookback setting changeable via Configs. TOML seems easiest to me.
CR configs I don't think can be changed without reboots atleast in current implementation.

pkg/solana/logpoller/log_poller.go
reductionista: oh, I always assumed "minimum slot that the request can be evaluated at" meant that it won't return results for slots less than that.  I guess you've tested this and found that's not how it works?

That's very unfortunate... if there's no way to specify a minimum block, then I agree we should probably space these requests out more. What a strange api!

pkg/solana/logpoller/log_poller.go
reductionista: @dhaidashenko Wow, that's way slower than I was thinking! Sounds like we do need to implement those performance optimizations before the event rate reaches 10 events/s. Hopefully won't be anywhere near that for a while.

reductionista: > CR configs I don't think can be changed without reboots at least in current implementation.

That's really unfortunate... we'll have to fix that at some point, I think dynamic reconfigurability will be important to have in the long run. Not having that negates some of the main selling points of having a ChainReader in the first place.

reductionista: > So it is also very important to make the Lookback setting changeable via Configs. TOML seems easiest to me.

Unfortunately, I don't think there is any way to identify specific event types in TOML config... those are specific properties of specific contracts associated with a specific type of job, and the TOML config isn't aware of jobs or contracts or events.  It should probably be in both places, but for now it sounds like TOML is the more important place.  Hopefully just having one global parameter for the whole node will be good enough for initial launch.

pkg/solana/logpoller/log_poller.go
reductionista: > Lookback is only for the time when the Node is restarted, or brought up first time

That's not the same as the original feature, but we could decide to restrict functionality to this.

I was thinking the time it would take to catch up if it was only behind by a day or two would only be a minute or two, so I didn't bother counting this as a downside... but in light of the numbers @dhaidashenko gave for how slow it will be, I want to make sure everyone is aware that leaving out the ability to automatically catch up when it detects it's too far behind will mean one of two things:
1. It will take a lot longer to catch up after a network outage or an rpc server or db going down for maintenance or upgrade for a while.
or
2. Node op wil have to restart the node in these situations, to catch up more quickly.

As long as everyone is okay with that, we can just get by with a configurable `StartingLookback` setting in the toml config.

pkg/solana/logpoller/log_poller.go
prashantkumar1982: Yes. Just to make sure, this the place where I expect we will add a LookbackLimit config value:
https://github.com/smartcontractkit/chainlink-solana/blob/4234da19fdb3dbd55cba40bacccd2d1b5582e3d2/pkg/solana/config/toml.go#L113

pkg/solana/logpoller/log_poller.go
prashantkumar1982: Yes I think it should be fine.

pkg/solana/logpoller/log_poller.go
```diff
@@ -525,3 +531,35 @@ func (lp *Service) backgroundWorkerRun(ctx context.Context) {
...
+
+	// nolint:gosec
+	// G115: integer overflow conversion uint64 -&gt; int64
+	latestSlot := int64(latestFinalizedSlot)
+	lookbackSlot := latestSlot - int64((lookback-1)/lp.blockTime) - 1 // = latestSlot - ceil(lookback/blockTime)
+
+	firstAvailableSlot, err := lp.client.GetFirstAvailableBlock(ctx) // Despite the name, this returns slot # not block number
+	if err != nil {
+		// This is an optimization to avoid requesting pruned blocks. If there's an err we can just ignore
+		return lookbackSlot, nil
```
prashantkumar1982, line 558: In this function, we should log the final value we chose as the lookbackSlot, and why. This can be used for troubleshooting later, and to understand what/why did the node choose for LogPoller on a restart.

dhaidashenko (approved):

prashantkumar1982 (approved):

toblich (approved):

PR #1165, opened 2025-04-01
Verify ChainID matches a Chain Selector

DylanTinianov:
### Description
Verify ChainID matches a chain selector, and only compare against the client if the chainID is for mainnet, testnet, or devnet.

Ticket: https://smartcontract-it.atlassian.net/browse/NONEVM-1502

### DIFF
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -146,16 +146,9 @@ type verifiedCachedClient struct {
 	client.ReaderWriter
 }
 
-func IsKnownChainID(chainID string) bool {
-	switch chainID {
-	case
-		client.MainnetGenesisHash,
-		client.TestnetGenesisHash,
-		client.DevnetGenesisHash:
-		return true
-	default:
-		return false
-	}
+func isValid32ByteBase58String(chainID string) bool {
+	_, err := solanago.PublicKeyFromBase58(chainID)
+	return err == nil
 }
 
 func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error) {
@@ -178,13 +171,14 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error) 
 		return v.chainIDVerified, fmt.Errorf("failed to fetch ChainID in verifiedCachedClient: %w", err)
 	}
 
-	if IsKnownChainID(v.expectedChainID) {
+	// Check if configured chain ID could be a genesis hash
+	if isValid32ByteBase58String(v.expectedChainID) {
 		if v.chainID != v.expectedChainID {
 			v.chainIDVerified = false
 			return v.chainIDVerified, fmt.Errorf("client returned mismatched chain id (expected: %s, got: %s): %s", v.expectedChainID, v.chainID, v.nodeURL)
 		}
 	} else {
-		v.lggr.Warnf("Configured chainID %s does not match genesis hash for mainnet, testnet, or devnet. "+
+		v.lggr.Warnf("Configured chainID %s is not a valid genesis hash (must be a base58-encoded 32-byte string)."+
 			"Assuming localnet setup and skipping verification.", v.expectedChainID)
 	}
 
--- a/pkg/solana/chain.go
+++ b/pkg/solana/chain.go
@@ -178,7 +178,7 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error) 
 		return v.chainIDVerified, fmt.Errorf("failed to fetch ChainID in verifiedCachedClient: %w", err)
 	}
 
-	_, err = chainsel.GetChainDetailsByChainIDAndFamily(v.expectedChainID, "solana")
+	_, err = chainsel.GetChainDetailsByChainIDAndFamily(v.expectedChainID, chainsel.FamilySolana)
 	if err != nil {
 		v.chainIDVerified = false
 		return v.chainIDVerified, err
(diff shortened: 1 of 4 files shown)


huangzhen1997 (dismissed):

amit-momin (dismissed):

pkg/solana/chain.go
```diff
@@ -164,12 +171,15 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
 		return v.chainIDVerified, fmt.Errorf("failed to fetch ChainID in verifiedCachedClient: %w", err)
 	}
 
-	// if this is localnet, allow any chain ID as long as it's not spoofing an official network
-	if v.expectedChainID != "localnet" {
+	// Check if configured chain ID could be a genesis hash
```
Farber98: instead of checking if it `could be` a genesis hash, would be useful to check if it actually matches one of the `chainIDs` we have registered? Checking provided chainID against [this map](https://github.com/smartcontractkit/chain-selectors/blob/239b77a97d483b3fc4daf39beff47255b936e2be/solana.go#L70-L76) could work

lmk if I'm way off here, just saw the PR but don't have much context about it

DylanTinianov: Nice catch, we have a thread on this and decided to use `GetChainDetailsByChainIDAndFamily`

pkg/solana/chain.go
```diff
@@ -164,8 +178,13 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
 		return v.chainIDVerified, fmt.Errorf("failed to fetch ChainID in verifiedCachedClient: %w", err)
 	}
 
-	// if this is localnet, allow any chain ID as long as it's not spoofing an official network
-	if v.expectedChainID != "localnet" {
+	_, err = chainsel.GetChainDetailsByChainIDAndFamily(v.expectedChainID, "solana")
```
huangzhen1997, line 181: instead of "solana", you can use `chainsel.FamilySolana`

huangzhen1997, line 181: instead of "solana", you can use `chainsel.FamilySolana`

huangzhen1997: instead of "solana", you can use `chainsel.FamilySolana`

huangzhen1997 (commented):
Also could you add new tests that covers some of [these](https://github.com/smartcontractkit/chain-selectors/blob/main/test_selectors_solana.yml) ?

pkg/solana/chain.go
```diff
@@ -164,8 +178,13 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
...
+	if err != nil {
```
huangzhen1997, line 182: Should probably return and mark verified If no error

pkg/solana/chain.go
```diff
@@ -164,8 +178,13 @@ func (v *verifiedCachedClient) verifyChainID(ctx context.Context) (bool, error)
 		return v.chainIDVerified, fmt.Errorf("failed to fetch ChainID in verifiedCachedClient: %w", err)
 	}
 
-	// if this is localnet, allow any chain ID as long as it's not spoofing an official network
-	if v.expectedChainID != "localnet" {
+	_, err = chainsel.GetChainDetailsByChainIDAndFamily(v.expectedChainID, chainsel.FamilySolana)
```
prashantkumar1982: Can we instead add this basic check at chain startup itself?
If a NOP has bad config, we should just fail to start the chain, rather than fail later at this stage.

I mean here:
https://github.com/smartcontractkit/chainlink-solana/blob/e8b0555aed77a2c17f073062c42c9090a617a81d/pkg/solana/chain.go#L261-L263

DylanTinianov: Yeah that's even better!

prashantkumar1982: If this provided chainId isn't in chain-selectors, then return error directly from newChain()

huangzhen1997: This might not work for unit tests that uses localnet ?

pkg/solana/chain_test.go
```diff
@@ -201,13 +201,14 @@ func TestSolanaChain_VerifiedClients(t *testing.T) {
 			assert.Equal(t, uint64(1234), slot)
 
 			node.URL = config.MustParseURL(mockServer.URL + "/mismatch")
-			testChain.id = "incorrect"
+			invalidGenesisHash := "7eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d"
```
huangzhen1997, line 204: This should be reverted.

pkg/solana/chain_test.go
```diff
@@ -201,13 +201,14 @@ func TestSolanaChain_VerifiedClients(t *testing.T) {
...
+			testChain.id = invalidGenesisHash
 			c, err = testChain.verifiedClient(node)
 			assert.NoError(t, err)
 			_, err = c.ChainID(t.Context())
 			// expect error from id mismatch (even if using a cached client) when performing RPC calls
 			assert.Error(t, err)
-			assert.Equal(t, fmt.Sprintf("client returned mismatched chain id (expected: %s, got: %s): %s", "incorrect", tc.genesisHash, node.URL), err.Error())
+			assert.Equal(t, fmt.Sprintf("invalid chain id %s for solana", invalidGenesisHash), err.Error())
```
huangzhen1997, line 211: this should be reverted

PR #1174, opened 2025-04-04
BCFR-1212 fix bind race, bind noop, unbind fixed

Unheilbar:
core ref: d6dc628e37584d513c9c14846f1ad22907019a88

[JIRA ticket](https://smartcontract-it.atlassian.net/browse/BCFR-1212)
- Event Bind registers filters before unregisters
- Event Unbind used to set empty name before call unregister (so it didn't remove filters)
- Bind the same address twice is noop
- CR Close() doesn't remove filters

### DIFF
--- a/pkg/solana/chainreader/filter.go
+++ b/pkg/solana/chainreader/filter.go
@@ -85,7 +85,6 @@ func (r *syncedFilter) Unregister(ctx context.Context, registrar filterRegistrar
 
 func (r *syncedFilter) unregister(ctx context.Context, registrar filterRegistrar, name string) error {
 	if !registrar.HasFilter(ctx, name) {
-		fmt.Println("333")
 		return nil
 	}
 
--- a/pkg/solana/chainreader/filter.go
+++ b/pkg/solana/chainreader/filter.go
@@ -16,16 +16,43 @@ type syncedFilter struct {
 	mu         sync.RWMutex
 	addressSet bool
 	filter     logpoller.Filter
+
+	dirty bool
 }
 
 func newSyncedFilter() *syncedFilter {
 	return &syncedFilter{}
 }
 
-func (r *syncedFilter) Register(ctx context.Context, registrar filterRegistrar) error {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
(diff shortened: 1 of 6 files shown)


pkg/solana/chainreader/filter.go
```diff
@@ -43,11 +70,26 @@ func (r *syncedFilter) Unregister(ctx context.Context, registrar filterRegistrar
...
+	r.setName("")
+
+	r.dirty = false
+
+	return nil
+}
+
+func (r *syncedFilter) unregister(ctx context.Context, registrar filterRegistrar, name string) error {
+	if !registrar.HasFilter(ctx, name) {
+		fmt.Println("333")
```
ilija42: left a println here

ilija42 (commented):
There is a lot of "no-op" code that gets called after Bind returns in 
`(s *ContractReaderService) Bind()` after an already registered address is encountered. If we can avoid calling that code by just returning earlier, maintenance and complexity would be quite a bit better. I don't think I see any bugs right now, but this change could prevent them in the future

wdyt?

Unheilbar:
> There is a lot of "no-op" code that gets called after Bind returns in `(s *ContractReaderService) Bind()` after an already registered address is encountered. If we can avoid calling that code by just returning earlier, maintenance and complexity would be quite a bit better. I don't think I see any bugs right now, but this change could prevent them in the future
> 
> wdyt?

```go
func (s *ContractReaderService) Bind(ctx context.Context, bindings []types.BoundContract) error {
	for idx := range bindings {
		if err := s.bdRegistry.Bind(ctx, s.reader, &bindings[idx]); err != nil {
			return err
		}
		
		s.lookup.bindAddressForContract(bindings[idx].Name, bindings[idx].Address)
```

I thought maybe we can check if we already have this binding in our lookup? And if so call continue

Farber98 (dismissed):
looks good! https://github.com/smartcontractkit/chainlink-solana/pull/1174#pullrequestreview-2746343385 seems a good idea to add

Farber98 (approved):

reductionista (approved):

reductionista:
LGTM

ilija42 (approved):

smickovskid (approved):

PR #1179, opened 2025-04-10
Forwarder

augustbleeds:
### Description

 

Design of forwarder based off of https://github.com/smartcontractkit/chainlink-solana/pull/614

EDIT: please see the committed README.md docs for more info

instructions:
* initialize: creates a new forwarder state, and precomputes the bump for the forwarder authority pda which signs the CPI to the receiver
* transfer/accept_ownership: self-explanatory
* init/update/close_oracles_config: oracle configs (which are called [oracle sets](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/keystone/KeystoneForwarder.sol#L56) in EVM) are stored as separate data accounts since a new config version / config_id with the same DON would create a new oracle config so the size is unbounded (as opposed to finding a soft-limit on the new dons that are created)
* report: processes the report which is of format `data =  bump | len_signatures (N) | signatures (N*65) | raw_report (M) | report_context (96)`. A report will create a new execution_state pda data account if doesn't exist via create_account instruction. It executes a CPI to a receiver program where the forwarder authority pda is the signer.


contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+pub const FORWARDER_METADATA_LENGTH: usize = 45;
+pub const METADATA_LENGTH: usize = 109;
+
+#[program]
+pub mod keystone_forwarder {
+    use anchor_lang::solana_program::{instruction::Instruction, program::invoke_signed, system_instruction};
+
+    use super::*;
+
+    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
```
dhaidashenko, line 30: Curious about what steps we need to perform for deployment.  Is it 
1. Deploy the program. 
2. Initialize (can we do it as part of the deployment transaction?)
3. Init oracles config
4. Transfer ownership to MCMS
5. Accept ownership on MCMS

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    pub fn accept_ownership(ctx: Context<AcceptOwnership>) -> Result<()> {
+        let state = &mut ctx.accounts.state;
+        state.owner = state.proposed_owner;
+        state.proposed_owner = Pubkey::default();
+
+        Ok(())
+    }
+
+    pub fn init_oracles_config(ctx: Context<InitOraclesConfig>, don_id: u32, config_version: u32, f: u8, signer_addresses: Vec<EthAddress>) -> Result<()> {
```
dhaidashenko: What's the motivation behind having two operations that modify oracle's config? Can we merge it into a single function?

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    pub system_program: Program<'info, System>,
+
+    // remaining accounts passed to receiver
+}
+
+#[account]
+#[derive(Default)]
+#[derive(InitSpace)]
+pub struct ExecutionState {
```
dhaidashenko: I wonder how we can verify if the report was forwarded on the WriteTarget level

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+fn extract_raw_report(data: &[u8]) -> Vec<u8> {
+    let _execution_state_bump = data[0];
+    let len = data[1] as usize;
+    let data = &data[2..];
+    let _signatures = &data[..65*len];
+    let data = &data[65*len..];
+    let _report_context = &data[data.len()-96..];
+    let raw_report = &data[..data.len()-96];
+
+    return raw_report.to_vec()
```
archseer: This should just return the slice, no need to allocate a Vec

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+    pub proposed_owner: Signer<'info>,
+}
+
+// combination of solidity's OracleSet and the configId mapping
+#[account]
+#[derive(Default)]
+pub struct OraclesConfig {
+    config_id: u64,
+    f: u8,
+    signer_addresses: Vec<EthAddress>, // maybe think about doing ECDSA over ed25519?
```
archseer: Personally I'd use a constant size array plus a length int and make the type zero_copy

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+}
+
+// data =  bump | len_signatures (N) | signatures (N*65) | raw_report (M) | report_context (96)
+fn extract_raw_report(data: &[u8]) -> Vec<u8> {
+    let _execution_state_bump = data[0];
+    let len = data[1] as usize;
+    let data = &data[2..];
+    let _signatures = &data[..65*len];
+    let data = &data[65*len..];
+    let _report_context = &data[data.len()-96..];
```
archseer: I think it would make more sense to have the report context before the report

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+    #[account(mut, address = state.owner @ AuthError::Unauthorized)]
+    pub owner: Signer<'info>,  // must be the same owner as the one in the state account
+}
+
+// data =  bump | len_signatures (N) | signatures (N*65) | raw_report (M) | report_context (96)
+fn extract_raw_report(data: &[u8]) -> Vec<u8> {
+    let _execution_state_bump = data[0];
+    let len = data[1] as usize;
+    let data = &data[2..];
+    let _signatures = &data[..65*len];
```
archseer: use constants for 65/96 in the function

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+    let mut prev_signer: EthAddress = [0u8; 20];
+
+    for &curr_signer in signer_addresses.iter() {
+        // will also fail if there is a duplicate signer
+        require!(curr_signer > prev_signer, ForwarderError::SignersNotSortedInIncreasingOrder);
+
+        prev_signer = curr_signer;
+    }
+
+    oracles_config.config_id = ((don_id as u64) << 32) | (config_version as u64);
```
archseer: This should use get_config_id

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,504 @@
...
+impl OraclesConfig {
+    pub const INIT_SPACE: usize = 8 + 1 + 4;
+
+    pub fn space_with_signers(num_signers: usize) -> usize {
+        Self::INIT_SPACE + num_signers*20
+    }
+}
+
+fn get_config_id(don_id: u32, config_version: u32) -> u64 {
+    ((don_id as u64) << 32) | (config_version as u64)
```
archseer: Config id could technically just be a struct with two u32 fields, it would have the same underlying size

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: I kept it this way because the existing solana keyring assumes the report context comes after https://github.com/smartcontractkit/chainlink/blob/develop/core/services/keystore/keys/ocr2key/solana_keyring.go#L38 . But open to changing it as well. 

We should be using ocr2 signatures, not ocr3 right?

augustbleeds:
note: wonder if i should add anchor events and/or helpers to the node module to listen for events

contracts/Anchor.toml
```diff
@@ -5,30 +5,24 @@ anchor_version = "0.29.0"
...
+[programs.localnet]
+access_controller = "9xi644bRR8birboDGdTiwBq3C7VEeR7VuamRYYXCubUW"
+contract-reader-interface = "6AfuXF6HapDUhQfE4nQG9C1SGtA1YjP3icaJyRfU4RyE"
+contract-reader-interface-secondary = "9SFyk8NmGYh5D612mJwUYhguCRY9cFgaS2vksrigepjf"
+dummy_receiver = "5z38tFCAmcPJb1DXUHSoKQhR8qQ8o9aNZ8rZFWe6gH4L"
+keystone_forwarder = "whV7Q5pi17hPPyaPksToDw1nMx6Lh8qmNWKFaLRQ4wz"
+log-read-test = "J1zQwrBNBngz26jRPNWsUSZMHJwBwpkoDitXRV95LdK4"
+ocr_2 = "cjg3oHmg9uuPsP8D6g29NWvhySJkdYdAo9D25PRbKXJ"
+store = "HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny"
+
```
aalu1418: nit: was the auto-formatted? or can we keep this relatively in the same place in the file?
it would reduce the blob change :)

contracts/programs/dummy-receiver/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[features]
+no-entrypoint = []
+no-idl = []
+no-log-ix-name = []
+cpi = ["no-entrypoint"]
+default = []
+
+[dependencies]
+anchor-lang = "0.29.0"
```
aalu1418, line 19: have we considered a `0.31.1` upgrade for the repo?

contracts/programs/dummy-receiver/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[derive(Accounts)]
+pub struct OnReport<'info> {
+    #[account(owner = FORWARDER_ID)]
+    pub state: Account<'info, ForwarderState>,
+
+    /// CHECK: This is a PDA
+    /// Anchor is unable to compute PDA with other program id so you
+    /// must do inline check within on_report
+    /// #[account(seeds = [b"forwarder", state.key().as_ref()], bump = state.authority_nonce)]
```
aalu1418: does this constraint work? just a guess - I have not tried it

```
#[account(
    seeds = <seeds>,
    bump,
    seeds::program = <expr>
)]
```

scripts/anchor-go-gen.sh
```diff
@@ -4,6 +4,12 @@ set -e
 
 for idl_path_str in "contracts/target/idl"/*
 do
+  filename=$(basename "$idl_path_str")
+  if [[ "$filename" == "dummy_receiver.json" || "$filename" == "keystone_forwarder.json" ]]; then
+    echo "Skipping $filename"
+    continue
+  fi
```
aalu1418, line 11: why skip these? might be useful in tests/tooling in the future?

contracts/programs/keystone-forwarder/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+no-entrypoint = []
+no-idl = []
+no-log-ix-name = []
+cpi = ["no-entrypoint"]
+default = []
+
+[dependencies]
+anchor-lang = "0.29.0"
+# disables frozen-abi
+# solana-program = { version = "1.17.25", default-features = false }
```
aalu1418: unnecessary lines?

contracts/programs/keystone-forwarder/src/error.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// customizable error range
+// note: custom numeric error codes start from 6000 unless specified like #[error_code(offset = 1000)]
+// https://github.com/coral-xyz/anchor/blob/c25bd7b7ebbcaf12f6b8cbd3e6f34ae4e2833cb2/lang/syn/src/codegen/error.rs#L72
+// Anchor built-in errors: https://anchor.so/errors
+//
+// [0:100]   Global errors
+// [100:N]   Function errors
+
+// this "AuthError" is separated from the "McmError" for error type generation from "anchor-go" tool
```
aalu1418: `McmError`?

contracts/programs/keystone-forwarder/src/error.rs
```diff
@@ -0,0 +1,51 @@
...
+// Workaround: keep errors that not likely to change during development in the first error_code block(keeping hardcoded error types for this),
+// and other errors in the second block.
+#[error_code]
+pub enum AuthError {
+    #[msg("The signer is unauthorized")]
+    Unauthorized,
+}
+
+#[error_code]
+pub enum ForwarderError {
```
aalu1418, line 24: you could potentially just combine the two error enums? might clean things up a bit

contracts/programs/dummy-receiver/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+                && expected_bump == ctx.accounts.state.authority_nonce,
+            AuthError::Unauthorized
+        );
+
+        // in a production setting you'd also want to verify the metadata too...
+
+        let account_info = &ctx.remaining_accounts[0];
+        let mut latest_report: Account<'info, LatestReport> = Account::try_from(account_info)?;
+        latest_report.metadata = metadata;
+        latest_report.report = report;
```
aalu1418: i think you should be able to declare extra accounts in the call context

it would also simplify the writing to state

contracts/programs/dummy-receiver/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[msg("The signer is unauthorized")]
+    Unauthorized,
+}
+
+#[account]
+#[derive(Default)]
+pub struct LatestReport {
+    pub metadata: Vec<u8>,
+    pub report: Vec<u8>,
+}
```
aalu1418: would a receiver need to store the whole report + metadata? or just parse it and save specific parameters

it's just a little unusual to store `Vec<u8>` since dynamic array sizes are a bit more annoying to handle in accounts (needs resizing)

aalu1418: could be better to save reportID, and specific data rather than the whole report + metadata which includes signatures and extras

contracts/programs/dummy-receiver/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[derive(Accounts)]
+pub struct Initialize<'info> {
+    #[account(
+        init,
+        payer = signer,
+        space = 8 + 4 + 4 + 65 // 64 (metadata) + 1 (report)
```
aalu1418: with concrete types this would also allow you to use `::INIT_SPACE`

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+    };
+
+    use super::*;
+
+    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
+        // store forwarder PDA bump for later
+        let (_authority_pubkey, authority_nonce) = Pubkey::find_program_address(
+            &[b"forwarder", ctx.accounts.state.key().as_ref()],
+            &crate::ID,
+        );
```
aalu1418: nit: does this work? no need to name it

```suggestion
        let (_, authority_nonce) = Pubkey::find_program_address(
            &[b"forwarder", ctx.accounts.state.key().as_ref()],
            &crate::ID,
        );
```

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+pub struct ForwarderState {
+    pub version: u8,
+    pub authority_nonce: u8, // bump
+    pub owner: Pubkey,
+    pub proposed_owner: Pubkey,
+}
+
+impl ForwarderState {
+    // version + authority_nonce + owner + proposed_owner + vec prefix for config vec
+    pub const SPACE: usize = 1 + 1 + 32 + 32;
```
aalu1418: what's the `config vec` used for?

if it's not used anymore, you could just use the default `::INIT_SPACE` rather than hardcoding values

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    // data =  bump (1) | len_signatures (1) | signatures (N*65) | raw_report (M) | report_context (96)
+    pub fn report<'info>(
+        ctx: Context<'_, '_, '_, 'info, Report<'info>>,
+        data: Vec<u8>,
+    ) -> Result<()> {
+        let num_signatures = data[1] as usize;
```
aalu1418: nit: there's a chance to panic when using `1` index

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+            )?;
+        } else {
+            // revert if execution succeded already
+
+            let execution_state_info = execution_state.to_account_info(); // or just AccountInfo
+            let state =
+                ExecutionState::try_deserialize(&mut &execution_state_info.data.borrow()[..])?;
+
+            require!(!state.success, ForwarderError::ExecutionAlreadySucceded)
+        }
```
aalu1418: does `init_if_needed` reduce the complexity of this code?

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        })
+        .chain(std::iter::once(AccountMeta {
+            pubkey: *forwarder_authority_pda.key,
+            is_signer: true,
+            is_writable: false,
+        }))
+        .chain(ctx.remaining_accounts.iter().map(|acc| AccountMeta {
+            pubkey: *acc.key,
+            // assume that we (probably) won't support 3rd party accounts as signers
+            is_signer: false,
```
aalu1418, line 174: not sure it's possible to support 3rd part signers because it requires a signer during tx construction? or going through another program

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+            is_writable: acc.is_writable,
```
aalu1418, line 175: are we requiring all `remaining_accounts` to be passed as writable?

i don't think that's a good assumption - you should be able to reconstruct it from the account info that gets passed in

contracts/programs/dummy-receiver/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[account(mut)]
+    pub signer: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
+}
+
+#[derive(Accounts)]
+pub struct OnReport<'info> {
+    #[account(owner = FORWARDER_ID)]
+    pub state: Account<'info, ForwarderState>,
```
aalu1418, line 92: since you can initialize multiple forwarders with the same program, this receiver would allow inputs from any forwarder (an attacker could create a forwarder)

potential solution is to create an allow list of forwarder states

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+
+        let account_infos: Vec<AccountInfo> = std::iter::once(ctx.accounts.state.to_account_info())
+            .chain(std::iter::once(
+                ctx.accounts.forwarder_authority.to_account_info(),
+            ))
+            .chain(ctx.remaining_accounts.iter().cloned())
+            .collect();
+
+        // payload begins with the Anchor discriminator
+        let mut payload = hash::hash("global:on_report".as_bytes()).to_bytes()[..8].to_vec();
```
aalu1418, line 187: why always calculate this? why not set it as a constant to save compute units

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        // metadata is just workflow_cid, workflow_name, workflow_owner, and report_id (see format above)
+        let metadata = &raw_report[FORWARDER_METADATA_LENGTH..METADATA_LENGTH].to_vec();
+        let report = &raw_report[METADATA_LENGTH..].to_vec();
+        // Borsh serialize each part separately
+        payload.extend(&metadata.try_to_vec()?);
+        payload.extend(&report.try_to_vec()?);
+
+        let ix = Instruction::new_with_bytes(ctx.accounts.receiver_program.key(), &payload, metas);
+
+        // used to derive the forwarder authority PDA
```
aalu1418, line 198: nit: this isn't to derive the forwarder PDA, this is do derive the PDA signature for the CPI call

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        let forwarder_state = ctx.accounts.state.key();
+        let signers_seeds = &[
+            b"forwarder",
+            forwarder_state.as_ref(),
+            &[ctx.accounts.state.authority_nonce],
+        ];
+
+        let _ = invoke_signed(&ix, &account_infos, &[signers_seeds]);
```
aalu1418:
```suggestion
        invoke_signed(&ix, &account_infos, &[signers_seeds])?;
```

contracts/programs/keystone-forwarder/src/lib.rs
aalu1418: agreed

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        require!(
+            curr_signer > prev_signer,
+            ForwarderError::SignersNotSortedInIncreasingOrder
+        );
+
+        prev_signer = curr_signer;
+    }
+
+    oracles_config.config_id = ((don_id as u64) << 32) | (config_version as u64);
+    oracles_config.f = f;
```
aalu1418, line 292: missing validation for number of signers

```
if (signers.length <= 3 * f) revert InsufficientSigners(signers.length, 3 * f + 1);
```

https://github.com/smartcontractkit/chainlink/blob/74e4e0077825f20d09b32cd2f5b45353d2c80f5f/contracts/src/v0.8/keystone/KeystoneForwarder.sol#L227-L228

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+pub struct Initialize<'info> {
+    // the account is not a PDA but it is initialized by the program
+    #[account(
+        init,
+        payer = owner,
+        space = 8 + ForwarderState::SPACE
+    )]
+    pub state: Account<'info, ForwarderState>,
+    #[account(mut)]
+    pub owner: Signer<'info>,
```
aalu1418: right now, this allows anyone to deploy a forwarder and there can be many forwarders - is that the intention?

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// CHECK: existing account will be updated OR new account will be initialized
+    #[account(
+        mut,
+        seeds = [
+            b"execution_state", 
+            &extract_transmission_id(extract_raw_report(&data), receiver_program.key)
+        ],
+        bump = data[0]
+    )]
+    pub execution_state: UncheckedAccount<'info>,
```
aalu1418: can you use `init_if_needed` here to save some other code for initializing and deserializing the account?

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+
+declare_id!("whV7Q5pi17hPPyaPksToDw1nMx6Lh8qmNWKFaLRQ4wz");
+
+pub const STATE_VERSION: u8 = 1;
+
+pub const ANCHOR_DISCRIMINATOR: usize = 8;
+
+pub const REPORT_CONTEXT_LEN: usize = 96;
+
+pub const MAX_REPORT_SIGNERS: usize = 17;
```
aalu1418: nit: document inline where this comes from

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+
+pub const SIGNATURE_LEN: usize = 65;
+
+pub const FORWARDER_METADATA_LENGTH: usize = 45;
+pub const METADATA_LENGTH: usize = 109;
```
aalu1418: are these lengths or indexes?

might be better to have length + start index?

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+
+    // it is dependent on the state.key(), a predetermined bump, workflow execution id, config_id, report_id
+    /// CHECK: existing account will be updated OR new account will be initialized
+    #[account(
+        mut,
+        seeds = [
+            b"execution_state", 
+            &extract_transmission_id(extract_raw_report(&data), receiver_program.key)
+        ],
+        bump = data[0]
```
aalu1418: i think the `bump` in this case should be part of the PDA seeds and not replace the actual PDA bump

https://www.helius.dev/blog/a-hitchhikers-guide-to-solana-program-security#bump-seed-canonicalization

aalu1418 (commented):
overall looking good - just some things to consider

contracts/programs/dummy-receiver/src/lib.rs
augustbleeds: ~Can you elaborate on what you mean?~

~On the forwarder side, I don't think you can specify explicit accounts that are not defined in the account context. (the anchor tooling doesn't let me). I think the only way to specify extra accounts is through the remaining accounts.~

Are you talking about passing in extra accounts on the forwarder using ctx.remaining_accounts, and then making them explicit accounts in the on_report CPI to the receiver? That's a great point. Let me give it a go

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: that should be okay because if it'd be an invalid report then. is it best practice to use requires! in this situation instead of a panic?

contracts/programs/dummy-receiver/src/lib.rs
augustbleeds: Yeah, this is just for testing purposes where I wanted to assert on the contents of the receiver for tests. 

I saw your comment elsewhere talking about productionizing a receiver example. I think that we should do that as well and have a couple different examples. I'd probably house that in a separate anchor project though just to make a clear distinction for code meant for developers to use and play around with.

contracts/programs/dummy-receiver/src/lib.rs
augustbleeds: Although an attacker could potentially create a forwarder, it's up to the receiver program to verify that the instance is allowed. I guess I forgot to do that here so I'll need to add that. But it'd be a hardcoded thing exactly like you mention. The data feeds cache on EVM for example has a list of allowed callers. Good call out

contracts/programs/dummy-receiver/src/lib.rs
augustbleeds: sweet! it does work :D

scripts/anchor-go-gen.sh
augustbleeds: it was giving me some errors because the way the CI generated the bindings vs how I generated the bindings was different. I had to timebox the debugging.

If I get some time I'll revisit adding it back in

contracts/programs/dummy-receiver/Cargo.toml
augustbleeds: I haven't... i'll look into it though. If we did 0.31.1 repo wouldn't we need another audit for the old contracts? and then bump those up on mainnet?

contracts/programs/keystone-forwarder/Cargo.toml
augustbleeds: thanks!

contracts/programs/keystone-forwarder/src/error.rs
augustbleeds: I think the comments from the CCIP repo suggest that we should separate them? I won't be working on the go-bindings, but when I hand off it to Jim's team they may run into issues if the go bindings aren't generated correctly.

contracts/programs/keystone-forwarder/src/error.rs
augustbleeds: an artifact of copy/paste 🤣 from the CCIP repo

contracts/programs/dummy-receiver/Cargo.toml
augustbleeds: EDIT: oh I assumed that anchor versions were tied to solana versions, but it seems they are mix and match. however i think an anchor version change could still potentially change change deployed bytecode.

contracts/Anchor.toml
augustbleeds: yeah, auto format ftw! I can remove it to what it was like previously

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: can you elaborate on the scenario? we can add an event to the report function which you can listen to or you can query the execution status account associated with the transmission

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: it could be possible if this becomes a feature of the solana write capability in general availability. but we can update the program if that scenario comes up later on.

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: i think it's implemented that way already. we don't require the remaining accounts to be writable (it could be true or false).

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: good point!

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: oh wow, i wish i knew about that constraint earlier!

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: nice find

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: yeah, the ability to create multiple forwarders should exist just like how you may deploy multiple forwarders on EVM.

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: these are lengths https://github.com/smartcontractkit/chainlink/blob/74e4e0077825f20d09b32cd2f5b45353d2c80f5f/contracts/src/v0.8/keystone/KeystoneForwarder.sol#L97C29-L97C44 but also can be used to slice the array

contracts/programs/keystone-forwarder/src/lib.rs
aalu1418: might be nice to add a comment here that anyone can create a forwarder :)

contracts/programs/keystone-forwarder/src/lib.rs
aalu1418: would be nice to update the comment here :)

contracts/programs/dummy-receiver/src/lib.rs
aalu1418: note that a receiver might want to support multiple forwarders writing to it - so an array (with `realloc`) might be a good solution here (assuming it can be a bounded set of forwarders, if not then PDAs might be necessary - CCIP ran into something similar here on the token pools)

contracts/programs/keystone-forwarder/src/lib.rs
aalu1418: oops sorry misread this - been staring at code for too long 😆

contracts/programs/keystone-forwarder/src/lib.rs
aalu1418: i can't remember if there's a difference in cost between failing because of a panic or failing because of a revert - might be worth double checking.

it's more of a nit - if you fail with a revert, it's easier to find where the issue happened

contracts/programs/dummy-receiver/Cargo.toml
aalu1418: yeah it would potentially require a lot more work - i don't know if it's worth doing it now, but just pointing out that it might be good to have

especially since the team is responsible for solana DF contracts now

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: @chray-zhang

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+#[inline(never)]
+fn verify_signatures(
+    hashed_report: &[u8; 32],
+    signatures: &[u8],
+    oracles_config: &Account<OraclesConfig>,
+    num_signers: usize,
+) -> Result<()> {
+    // ensure MAX_SIGNERS fit in the bits of uniques
+    let mut uniques: u32 = 0;
+    assert!(uniques.count_ones() + uniques.count_zeros() >= MAX_ORACLES as u32);
```
silaslenihan, line 233: better to use require! and a custom error here?

contracts/programs/keystone-forwarder/src/utils.rs
```diff
@@ -0,0 +1,43 @@
...
+// don_id                 offset  37, size  4
+// don_config_version     offset  41, size  4
+// workflow_cid           offset  45, size 32
+// workflow_name          offset  77, size 10
+// workflow_owner         offset  87, size 20
+// report_id              offset 107, size  2
+
+pub fn extract_config_id(raw_report: &[u8]) -> [u8; 8] {
+    // don_id | don_config_version
+    raw_report[37..45].try_into().expect("Expected 8 bytes")
```
silaslenihan, line 34: Another case where this will panic, would it be better to guard this with require!

docs/forwarder/README.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+`state` is a keypair, program owned account which represents the forwarder instance we are working with. Think of this is a specific forwarder instance.
+
+`oracles_config` is a PDA which stores DON info
+
+`forwarder_authority` is the PDA which is responsible for signing the CPI to a receiver program via invoke_signed()
+
+`execution_state` is the PDA which stores a report's execution status. For now, it only stores successes.
```
silaslenihan, line 58: how will offchain components (writetarget) know when a transmission is failed / fatal?

augustbleeds: Good question. This is something that the platform expansion team will need to add in later iterations. I think the body of work is too large for our timeline at the moment. But I have some thoughts [here](https://chainlink-core.slack.com/archives/C024M6N5MS8/p1743181059358989). I think the first option is the most robust, but we have to have a new capabilities for that to happen.

As it is right now, the write targeter will just keep on trying until it reaches the max tries and then stop.

contracts/programs/keystone-forwarder/src/utils.rs
augustbleeds: good check, this something we have a ticket for @chray-zhang will pick up some of the small improvements

chray-zhang (dismissed):

contracts/tests/keystone_forwarder.spec.ts
```diff
@@ -285,6 +287,8 @@ describe("keystone_storage", function () {
       actualState4.proposedOwner.equals(PublicKey.default),
       "proposed owner is 0"
     );
+
+    await Promise.all([transferOwnershipEventPromise, acceptOwnershipEventPromise])
```
augustbleeds: The bug was caused due to the anchor listener waiting forever for the event to appear. This does not occur  if you await the promise after another `await`, potentially because the event list is hydrated.

amit-momin (approved):
Approving to help move along

Unheilbar (approved):

smickovskid (approved):

PR #1182, opened 2025-04-10
Added cache to BlockHistoryEstimator to support larger block ranges

amit-momin:
### Context
Solana RPCs do not support fetching blocks in batches. Previously, an RPC call had to be made for each block when fetching the latest `BlockHistorySize` blocks. This was not sustainable if a large `BlockHistorySize` needed to be configured.

### Solution
An in-memory cache was added to store the latest `BlockHistorySize` blocks to avoid refetching ones that are still relevant to estimates. This way only the new blocks that were generated between poll periods need to be fetched and stored in the cache. If the cache contains more blocks than `BlockHistorySize`, the oldest `len(cache) - BlockHistorySize` blocks are removed.

A new config `BlockHistoryBatchLoadSize` was introduced to avoid overloading the RPC on startup when the cache is empty. This config controls how many blocks are fetched each interval. This config is only used if `BlockHistorySize > 1` and if `BlockHistoryBatchLoadSize < BlockHistorySize`. Even if an RPC can handle loading all blocks upfront, it may take longer to process them than they stay relevant for. For example, if processing 150 blocks can take more than 1 min, 150 new blocks could have passed in that time invalidating that work. It's better to load blocks in smaller batches to enable the cache to save resources. Loading in smaller batches also means initial estimations would be on smaller block ranges than `BlockHistorySize` during startup until the cache is filled. If estimations are blocked until the cache is filled, the TXM could be blocked for too long of a period. There isn't a risk of bad estimations for broadcast since blocks are loaded from latest to oldest. The most recent market price would always be used.

### Depends on this PR
Core PR: https://github.com/smartcontractkit/chainlink/pull/17247

### DIFF
--- a/pkg/solana/config/config.go
+++ b/pkg/solana/config/config.go
@@ -79,30 +79,30 @@ type Config interface {
 }
 
 type Chain struct {
-	BlockTime                 *config.Duration
-	BalancePollPeriod           *config.Duration
-	ConfirmPollPeriod           *config.Duration
-	OCR2CachePollPeriod         *config.Duration
-	OCR2CacheTTL                *config.Duration
-	TxTimeout                   *config.Duration
-	TxRetryTimeout              *config.Duration
-	TxConfirmTimeout            *config.Duration
-	TxExpirationRebroadcast     *bool
-	TxRetentionTimeout          *config.Duration
-	SkipPreflight               *bool
-	Commitment                  *string
-	MaxRetries                  *int64
-	FeeEstimatorMode            *string
-	ComputeUnitPriceMax         *uint64
-	ComputeUnitPriceMin         *uint64
-	ComputeUnitPriceDefault     *uint64
-	FeeBumpPeriod               *config.Duration
-	BlockHistoryPollPeriod      *config.Duration
-	BlockHistorySize            *uint64
+	BlockTime                  *config.Duration
+	BalancePollPeriod          *config.Duration
+	ConfirmPollPeriod          *config.Duration
+	OCR2CachePollPeriod        *config.Duration
+	OCR2CacheTTL               *config.Duration
+	TxTimeout                  *config.Duration
+	TxRetryTimeout             *config.Duration
+	TxConfirmTimeout           *config.Duration
+	TxExpirationRebroadcast    *bool
+	TxRetentionTimeout         *config.Duration
+	SkipPreflight              *bool
+	Commitment                 *string
+	MaxRetries                 *int64
+	FeeEstimatorMode           *string
+	ComputeUnitPriceMax        *uint64
+	ComputeUnitPriceMin        *uint64
+	ComputeUnitPriceDefault    *uint64
+	FeeBumpPeriod              *config.Duration
+	BlockHistoryPollPeriod     *config.Duration
+	BlockHistorySize           *uint64
 	BlockHistoryCacheLoadBatch *uint64
-	ComputeUnitLimitDefault     *uint32
-	EstimateComputeUnitLimit    *bool
-	LogPollerStartingLookback *config.Duration
+	ComputeUnitLimitDefault    *uint32
+	EstimateComputeUnitLimit   *bool
+	LogPollerStartingLookback  *config.Duration
 }
 
 func (c *Chain) SetDefaults() {
--- a/pkg/solana/config/config.go
+++ b/pkg/solana/config/config.go
@@ -25,15 +25,16 @@ var defaultConfigSet = Chain{
 	MaxRetries:              ptr(int64(0)), // max number of retries (default = 0). when config.MaxRetries < 0), interpreted as MaxRetries = nil and rpc node will do a reasonable number of retries
 
 	// fee estimator
-	FeeEstimatorMode:         ptr("fixed"), // "fixed" or "blockhistory"
-	ComputeUnitPriceMax:      ptr(uint64(1_000)),
-	ComputeUnitPriceMin:      ptr(uint64(0)),
-	ComputeUnitPriceDefault:  ptr(uint64(0)),
-	FeeBumpPeriod:            config.MustNewDuration(3 * time.Second), // WARNING: If FeeBumpPeriod is shorter than blockhash expiration, multiple valid transactions can exist in parallel. This can result in higher costs and can cause unexpected behaviors if contracts do not de-dupe txs. Set to 0 to disable fee bumping.
-	BlockHistoryPollPeriod:   config.MustNewDuration(5 * time.Second),
-	BlockHistorySize:         ptr(uint64(1)),       // 1: uses latest block; >1: Uses multiple blocks, where n is number of blocks. DISCLAIMER: 1:1 ratio between n and RPC calls.
-	ComputeUnitLimitDefault:  ptr(uint32(200_000)), // set to 0 to disable adding compute unit limit
-	EstimateComputeUnitLimit: ptr(false),           // set to false to disable compute unit limit estimation
+	FeeEstimatorMode:           ptr("fixed"), // "fixed" or "blockhistory"
+	ComputeUnitPriceMax:        ptr(uint64(1_000)),
+	ComputeUnitPriceMin:        ptr(uint64(0)),
+	ComputeUnitPriceDefault:    ptr(uint64(0)),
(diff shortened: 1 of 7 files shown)


pkg/solana/config/config.go
```diff
@@ -29,15 +29,16 @@ var defaultConfigSet = Chain{
...
+	FeeEstimatorMode:          ptr("fixed"), // "fixed" or "blockhistory"
+	ComputeUnitPriceMax:       ptr(uint64(1_000)),
+	ComputeUnitPriceMin:       ptr(uint64(0)),
+	ComputeUnitPriceDefault:   ptr(uint64(0)),
+	FeeBumpPeriod:             config.MustNewDuration(3 * time.Second), // WARNING: If FeeBumpPeriod is shorter than blockhash expiration, multiple valid transactions can exist in parallel. This can result in higher costs and can cause unexpected behaviors if contracts do not de-dupe txs. Set to 0 to disable fee bumping.
+	BlockHistoryPollPeriod:    config.MustNewDuration(5 * time.Second),
+	BlockHistorySize:          ptr(uint64(1)),       // set to the number of blocks estimations should be made over. 1: uses latest block; >1: Uses multiple blocks, where n is number of blocks.
+	BlockHistoryBatchLoadSize: ptr(uint64(20)),      // set to the number of blocks that should be loaded into the cache every poll period if BlockHistorySize > 1. Ensure this value is greater than the number of blocks that would be produced between each BlockHistoryPollPeriod to avoid block gaps. BlockHistorySize is used instead if BlockHistorySize <  BlockHistoryBatchLoadSize.
```
Farber98, line 39: > ... the number of blocks that would be produced between each BlockHistoryPollPeriod ... 

This is `BlockHistorySize` but defined differently than the comment in the line above?

pkg/solana/config/toml.go
```diff
@@ -199,6 +199,15 @@ func setFromChain(c, f *Chain) {
...
 	}
+	if f.BlockHistoryBatchLoadSize != nil {
+		c.BlockHistoryBatchLoadSize = f.BlockHistoryBatchLoadSize
+	}
+	if f.ComputeUnitLimitDefault != nil {
+		c.ComputeUnitLimitDefault = f.ComputeUnitLimitDefault
+	}
+	if f.EstimateComputeUnitLimit != nil {
+		c.EstimateComputeUnitLimit = f.EstimateComputeUnitLimit
+	}
```
Farber98, line 210: These where defined in configs and missing here? 😟

pkg/solana/fees/block_history.go
```diff
@@ -44,6 +54,7 @@ func NewBlockHistoryEstimator(c func(context.Context) (client.ReaderWriter, erro
 		cfg:    cfg,
 		lgr:    lgr,
 		price:  cfg.ComputeUnitPriceDefault(), // use default value
+		cache:  blockMedianCache{medianMap: make(map[uint64]ComputeUnitPrice)},
 	}, nil
```
Farber98: nit: we already know the capacity of the cache. wdyt about preallocating the size?  
```suggestion
		cache:  blockMedianCache{medianMap: make(map[uint64]ComputeUnitPrice, cfg.BlockHistorySize)},
	}, nil
```

pkg/solana/fees/block_history.go
```diff
@@ -155,7 +169,40 @@ func (bhe *blockHistoryEstimator) calculatePriceFromLatestBlock(ctx context.Cont
...
+	bhe.lgr.Debugw("BlockHistoryEstimator: updated",
+		"computeUnitPriceAvg", avgOfMedians,
+		"numBlocks", len(blockMedians),
+	)
+
+	return nil
+}
+
+func (bhe *blockHistoryEstimator) populateCache(ctx context.Context, loadBatch, desiredBlockCount uint64) error {
+	batch := uint64(math.Min(float64(loadBatch), float64(desiredBlockCount)))
```
Farber98, line 206: This comment is a nit in the following lines. I can't point bcs it's not a git diff 😢 

```
	// Fetch the latest slot
	currentSlot, err := c.SlotHeight(ctx)
```

nit: worth adding a comment on the [underlying func](https://github.com/smartcontractkit/chainlink-solana/blob/c0b0b10a8440a354050b1fd47adc104607f4b87b/pkg/solana/client/client.go#L129-L131)? It's not clear without looking at the implementation that the latest slot refers to `Processed Commitment`

pkg/solana/fees/block_history.go
```diff
@@ -169,28 +216,34 @@ func (bhe *blockHistoryEstimator) calculatePriceFromMultipleBlocks(ctx context.C
...
 
 	// Determine the starting slot for fetching blocks
-	if currentSlot < desiredBlockCount {
+	if currentSlot < batch {
 		return fmt.Errorf("current slot is less than desired block count")
 	}
-	startSlot := currentSlot - desiredBlockCount + 1
+	startSlot := currentSlot - batch + 1
 
 	// Fetch the last confirmed block slots
```
Farber98: nit: `Fetch the last confirmed block slots` could be misleading. [GetBlocksWithLimit](https://github.com/smartcontractkit/chainlink-solana/blob/c0b0b10a8440a354050b1fd47adc104607f4b87b/pkg/solana/client/client.go#L402) uses the configured commitment under the hood. If it's not `Confirmed`, then this comment is wrong. 

Maybe we can just comment the underlying func

pkg/solana/fees/block_history.go
```diff
@@ -155,7 +169,40 @@ func (bhe *blockHistoryEstimator) calculatePriceFromLatestBlock(ctx context.Cont
...
+		blockMedians = append(blockMedians, median)
+	}
+
+	// Calculate avg from medians of the blocks.
+	avgOfMedians, err := mathutil.Avg(blockMedians...)
+	if err != nil {
+		return fmt.Errorf("failed to calculate price from avg of medians: %w", err)
+	}
+
+	// Update the current price to the calculated average (avg of medians of the last desiredBlockCount)
```
Farber98: nit: worth clarifying that `avg of medians of the last desiredBlockCount` is applicable once the cache is filled? bcs initial estimations would be on smaller block ranges until that

pkg/solana/fees/block_history_test.go
```diff
@@ -207,24 +207,63 @@ func TestBlockHistoryEstimator_MultipleBlocks(t *testing.T) {
...
+		waitForEstimation(t, estimator, pollPeriod)
+
 		// Calculated avg price should be equal to the one extracted manually from the blocks.
-		require.NoError(t, estimator.calculatePrice(ctx))
-		cfg.On("ComputeUnitPriceMax").Return(maxPrice)
+		require.NoError(t, estimator.calculatePriceFromMultipleBlocks(depth))
 		assert.Equal(t, uint64(multipleBlocksAvg), estimator.BaseComputeUnitPrice())
 	})
 
+	t.Run("Successful Estimation with partial cache fill", func(t *testing.T) {
```
Farber98, line 221: nit: should we test the eviction logic is working as expected when cache exceeds `BlockHistorySize`?

Farber98 (dismissed):
Good changes and very easy to follow! 🙌 left some comments but nothing major

pkg/solana/config/config.go
amit-momin: I think this is different than `BlockHistorySize` since that can span over a much larger range than just the blocks produced between poll periods. Like if our `BlockHistorySize` is 150 but we poll every 5 seconds, the network would have only produced 12-13 new blocks between polls. We just want `BlockHistoryBatchLoadSize` to be greater than 12-13. I can maybe update the comment to something like:
> ... the number of blocks that would be produced by the network between each BlockHistoryPollPeriod ...

pkg/solana/config/toml.go
amit-momin: Ya but we're using the per tx overrides to set these configs in our current use case so we didn't really need it here. But still sort of a miss.

pkg/solana/fees/block_history.go
amit-momin: The capacity probably exceeds the `BlockHistorySize` for a bit before we clean it up but still a good idea to allocate at least that much to avoid unnecessary reallocs. Good catch! I'll also initialize `storedBlockRange`

pkg/solana/config/config.go
Farber98: oh now I get it. It's okay to leave it as it is

Farber98 (approved):

PR #1191, opened 2025-04-16
Increase MaxRetries and make it configurable

reductionista:
### Description

Update: the "adds some important logging" part of this PR I've split off and merged separately, since the rest has become lower priority:

## Supports
https://github.com/smartcontractkit/chainlink/pull/17374

- Increases retry threshold from 6 to 18

[NONEVM-1626]: https://smartcontract-it.atlassian.net/browse/NONEVM-1626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/solana/config/config.go
+++ b/pkg/solana/config/config.go
@@ -43,7 +43,7 @@ var defaultConfigSet = Chain{
 	// log poller
 	LogPollerStartingLookback: config.MustNewDuration(24 * time.Hour),
 	LogPollerMaxRetries:       ptr(uint8(18)), // max number of retries for log poller
-	LogPollerLargeQueueNotify: ptr(2000),      // queue depth at which the worker group will log a warning
+	LogPollerLargeQueueNotify: ptr(200),       // queue depth at which the worker group will log a warning
 	LogPollerWorkerCount:      ptr(uint64(8)), // default number of workers for log poller
 }
 
--- a/pkg/solana/config/config.go
+++ b/pkg/solana/config/config.go
@@ -42,6 +42,9 @@ var defaultConfigSet = Chain{
 
 	// log poller
 	LogPollerStartingLookback: config.MustNewDuration(24 * time.Hour),
+	LogPollerMaxRetries:       ptr(uint8(18)), // max number of retries for log poller
+	LogPollerLargeQueueNotify: ptr(2000),      // queue depth at which the worker group will log a warning
+	LogPollerWorkerCount:      ptr(uint64(8)), // default number of workers for log poller
 }
 
 type Config interface {
@@ -76,6 +79,9 @@ type Config interface {
 
 	// log poller
 	LogPollerStartingLookback() time.Duration
+	LogPollerMaxRetries() uint8
+	LogPollerLargeQueueNotify() int
+	LogPollerWorkerCount() uint64
 }
 
 type Chain struct {
@@ -103,6 +109,9 @@ type Chain struct {
 	ComputeUnitLimitDefault   *uint32
 	EstimateComputeUnitLimit  *bool
 	LogPollerStartingLookback *config.Duration
+	LogPollerMaxRetries       *uint8
+	LogPollerLargeQueueNotify *int
+	LogPollerWorkerCount      *uint64
 }
 
 func (c *Chain) SetDefaults() {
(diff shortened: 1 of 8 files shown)


pkg/solana/logpoller/log_poller.go
```diff
@@ -229,6 +229,8 @@ func (lp *Service) Process(ctx context.Context, programEvent ProgramEvent) (err
 			log.ExpiresAt = &expiresAt
 		}
 
+		lp.lggr.Infof("found matching event %v", log)
```
archseer:
```suggestion
		lp.lggr.Infow("found matching event", "log", log)
```

archseer: We should always use `w` variants over `f` when using logger so it uses structured logging

pkg/solana/logpoller/worker/worker.go
```diff
@@ -21,11 +21,11 @@ var (
 
 const (
 	// DefaultMaxRetryCount is the number of times a job will be retried before being dropped.
-	DefaultMaxRetryCount = 6
+	DefaultMaxRetryCount = 18 // give up after ~ 14 hours
 	// DefaultNotifyRetryDepth is the retry queue depth at which the worker group will log a warning.
-	DefaultNotifyRetryDepth = 200
+	DefaultNotifyRetryDepth = 2000
```
reductionista, line 26: Increasing these to reduce noise--even for our manual e2e testing in staging it's normal for it to get up to at least 1600

pkg/solana/logpoller/worker/worker.go
```diff
@@ -21,11 +21,11 @@ var (
...
 	// DefaultNotifyQueueDepth is the queue depth at which the worker group will log a warning.
-	DefaultNotifyQueueDepth = 100
+	DefaultNotifyQueueDepth = 1000
```
reductionista: Increasing these to reduce noise--even for our manual e2e testing in staging it's normal for it to get up to at least 1600

reductionista: I'm not sure if this is what we want. Should we reduce it to just an hour or two? 14 hours is a long time to get stuck if it hits a problem with a particular request.  We detect at least one case where we shouldn't retry, and can move on, but different rpc servers might return slightly different errors for that... and there could easily be other cases where moving on after an hour would be better than freezing up for 14 hours.

prashantkumar1982: Can we create a TOML config for each of these?
Exactly for these situations where we don't know the right defaults. We don't want to find out in production the hard way, and rust to do a hotfix.

pkg/solana/client/test_helpers.go
```diff
@@ -28,17 +27,19 @@ func SetupLocalSolNodeWithFlags(t *testing.T, flags ...string) (string, string)
 	t.Helper()
 
 	port := utils.MustRandomPort(t)
-	portInt, _ := strconv.Atoi(port)
+	wsPort := utils.MustRandomPort(t)
 
 	faucetPort := utils.MustRandomPort(t)
 	url := "http://127.0.0.1:" + port
-	wsURL := "ws://127.0.0.1:" + strconv.Itoa(portInt+1)
+	wsURL := "ws://127.0.0.1:" + wsPort
```
dhaidashenko: we can't generate random WS port as it's not configurable on solana side.

pkg/solana/config/config.go
```diff
@@ -42,6 +42,9 @@ var defaultConfigSet = Chain{
 
 	// log poller
 	LogPollerStartingLookback: config.MustNewDuration(24 * time.Hour),
+	LogPollerMaxRetries:       ptr(uint8(18)), // max number of retries for log poller
```
dhaidashenko, line 45: MaxRetries only controls how many times the job should fail before we log critical error. I'm not sure that we should increase it. The job won't be removed from the queue until it succeeds.    

AFAIK backoff is caluclated calculated as follows: 
```
return time.Duration(2<<retries) * 100 * time.Millisecond
```
So the delay before 18th  retry  may take up to 873 minutes, if my math is correct.

pkg/solana/logpoller/log_poller.go
```diff
@@ -316,10 +316,10 @@ func (lp *Service) getLastProcessedSlot(ctx context.Context) (lastProcessed int6
 		return 0, err
 	}
 	if lookbackSlot > lastProcessed {
-		lp.lggr.Infow("last processed slot is older than lookback window, skipping ahead", "lastProcessed", lastProcessed, "lookbackSlot", lookbackSlot)
+		lp.lggr.Infow("last processed slot is older than lookback window, skipping ahead to lookback slot", "last_processed", lastProcessed, "lookback_slot", lookbackSlot)
 		return lookbackSlot, nil
 	}
-	lp.lggr.Infow("last processed slot is still within lookback window, resuming at last processed slot", "lastProcessed", lastProcessed, "lookbackSlot", lookbackSlot)
```
reductionista, line 322: Originally, this commit changed all the Infof's to Infow's, but looks like that just went into another commit I rebased on top of this morning so... not much is different.

pkg/solana/config/config.go
reductionista: > MaxRetries only controls how many times the job should fail before we log critical error. I'm not sure that we should increase it. The job won't be removed from the queue until it succeeds.

Oh, thanks I didn't realize this!  

I wonder if this is actually what we want. Agreed that we should start logging critical errors much earlier than this. But maybe it ought to give up at some point rather than allow itself to get permanently stuck. If it's already well past the 8 hour window that CCIP cares about, moving on seems like it might be better than requiring manual intervention. Maybe we need CCIP's feedback on this question:
@winder

PR #1195, opened 2025-04-17
Root Cause of Multiple Registered Filters

EasterTheBunny:
### Description
When a node restarts and `Bind` is called on a `ContractReader` instance, depending on the config a new log poller filter is registered. This registered filter may be a direct copy of an existing filter, but the name is different. There is no process that removes unused filters and over time/restarts filters accumulate in the database, often pulling the same data from blocks and adding load to log poller.

This PR adds a test to simulate the filter accumulation over multiple restarts and quantify the problem.

[NONEVM-1627](https://smartcontract-it.atlassian.net/browse/NONEVM-1627)
 

### Resolves Dependencies

[NONEVM-1627]: https://smartcontract-it.atlassian.net/browse/NONEVM-1627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/solana/chainreader/chain_reader.go
+++ b/pkg/solana/chainreader/chain_reader.go
@@ -635,7 +635,9 @@ func (s *ContractReaderService) addEventRead(
 	)
 
 	s.shouldStartLP = true
-	reader.SetFilter(toLPFilter(readDefinition.ChainSpecificName, pf, subkeys.subKeys[:], eventDef))
+	if err := reader.SetFilter(toLPFilter(readDefinition.ChainSpecificName, pf, subkeys.subKeys[:], eventDef)); err != nil {
+		return err
+	}
 
 	s.bdRegistry.AddReader(namespace, genericName, reader)
 	s.lookup.addReadNameForContract(namespace, genericName, []read{{readName: genericName, useParams: false}})
--- a/pkg/solana/chainreader/chain_reader_test.go
+++ b/pkg/solana/chainreader/chain_reader_test.go
@@ -592,6 +592,83 @@ func TestSolanaChainReaderService_GetLatestValue(t *testing.T) {
 	})
 }
 
+func TestSolanaChainReaderService_DatabaseBloat(t *testing.T) {
+	// For every service start, new filters are created from the provided config. This can result in database bloat
+	// over long periods of time and add unnecessary load to LogPoller. This test attempts to quantify and limit bloat
+	// caused by restarts.
+
+	ctx := t.Context()
+	lggr := logger.Nop()
+	mAccts := new(mockedMultipleAccountGetter)
+	mEvts := mocks.NewEventsReader(t)
+	pubKey1 := solana.NewWallet().PublicKey()
+	pubKey2 := solana.NewWallet().PublicKey()
+	bindCallCount := 5
+	restartCount := 10
+
+	boolType := codec.IdlType{}
+	require.NoError(t, boolType.UnmarshalJSON([]byte("\"bool\"")))
+
+	cfg := config.ContractReader{
+		Namespaces: map[string]config.ChainContractReader{
+			"myChainReader": {
+				IDL: codec.IDL{
+					Accounts: []codec.IdlTypeDef{{Name: "myAccount",
+						Type: codec.IdlTypeDefTy{
+							Kind:   codec.IdlTypeDefTyKindStruct,
+							Fields: &[]codec.IdlField{}}}},
+					Events: []codec.IdlEvent{
+						{Name: "myEvent", Fields: []codec.IdlEventField{{Name: "a", Type: boolType}}},
+					},
+				},
+				Reads: map[string]config.ReadDefinition{
+					"myRead": {
+						ChainSpecificName: "myEvent",
+						ReadType:          config.Event,
+						EventDefinitions: &config.EventDefinitions{
+							IndexedField0: &config.IndexedField{
+								OffChainPath: "A.B",
+								OnChainPath:  "A.B",
+							},
+							PollingFilter: &config.PollingFilter{},
+						}}}}},
+		AddressShareGroups: nil,
+	}
+
+	mEvts.EXPECT().Start(mock.Anything).Return(nil).Maybe()
+	mEvts.EXPECT().Ready().Return(nil).Maybe()
+	mEvts.EXPECT().HasFilter(mock.Anything, mock.Anything).Return(false).Maybe()
+
+	filters := []string{}
+
+	mEvts.EXPECT().RegisterFilter(mock.Anything, mock.Anything).Run(func(_ context.Context, filter logpoller.Filter) {
+		filters = append(filters, filter.Name)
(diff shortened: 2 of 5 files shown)


reductionista:
> When a node restarts and Bind is called on a ContractReader instance, depending on the config a new log poller filter is registered. This registered filter may be a direct copy of an existing filter, but the name is different.

I think there are two issues here. One is that filters should be registered when jobs are created and unregistered when they are deleted... and we're not doing the second one in all cases.

But the second issue is the random string added by ChainReader.

Prior to ChainReader, filters were always required to be deterministically generated from the parameters in the job spec. So I was expecting when we migrated to ChainReader we would continue to enforce this, by generating them deterministically from the parameters in the ChainReader config.  Instead, it's adding a random UUID on the end which violates the basic requirements of how filter names were supposed to be generated.

However, it does seem more challenging for ChainReader to guarantee uniqueness per job than for specific plugins. For specific plugins, at least each plugin understands the different roles it performs and knows what parameters can vary between different jobs of the same general type.  But ChainReader doesn't really know that.  I'm not sure how to restore determinism without messing up the uniqueness property, but that should be the goal.

reductionista:
For this PR, I wouldn't worry as much about this issue:
> I think there are two issues here. One is that filters should be registered when jobs are created and unregistered when they are deleted... and we're not doing the second one in all cases.

This is mostly an issue with individual plugins, who are responsible for calling Unbind at the right time. Or, if we determine that's not always possible, we could add some sort of filter-cleanup process to LogPoller.  But the randomness problem is a ChainReader bug... it would be nice if we could find a solution to that here.

EasterTheBunny:
> But the randomness problem is a ChainReader bug

This would have to be true for both the Solana and EVM implementations at this point. I'm not sure why a random UUID was chosen as a distinguishing characteristic. I do like the idea of a deterministic name. We should be able to do this fairly easily with hashing.

If we used deterministic names, would `LogPoller` correctly backfill after restarts?

reductionista:
> This would have to be true for both the Solana and EVM implementations at this point. I'm not sure why a random UUID  was chosen as a distinguishing characteristic. I do like the idea of a deterministic name. We should be able to do this fairly easily with hashing.

That's a good point, we probably have a lot of duplicates there too. Although it will be less of an issue on EVM because we don't store a separate copy of each log for every filter that's registered... everything  just gets merged into one big filter. This makes things a lot messier in terms of figuring out retention periods, and when we can delete logs, but duplicate filters don't increase storage space.

> If we used deterministic names, would LogPoller correctly backfill after restarts?

The name of the filter won't affect how it backfills. The tricky thing is guaranteeing uniqueness... I assume that's why the random string was added. If the same node has two jobs which both use the same ChainReader config, there needs to be a way for them to have two different filter names. Otherwise the second one will just get a name conflict error back when it tries to register its filter.

Since each product should have its own ChainReader config, I don't think we have to worry too much about conflicts between different product plugins--worst case, a new product can just prepend the product name in front if there's a conflict with an existing product.  But I'm not sure there is a general way that will always work for how to differentiate between two jobs of the same type.  I guess we could just make sure it works for CCIP, and assume we'll figure it out for future products as they come? CR can look at the entire config... so if there is a contract address or anything else that's expected to uniquely distinguish one job from another, it can use that.  (If not, we might have to have the plugin add a job id or something to CR config before it calls the constructor?)

EasterTheBunny:
> If the same node has two jobs which both use the same ChainReader config, there needs to be a way for them to have two different filter names. Otherwise the second one will just get a name conflict error back when it tries to register its filter.

I don't really understand this though. If two jobs have the exact same filter, why not have both of them querying the same logs? It seems like duplicate work on the part of `LogPoller` to support two jobs wanting the same data.

With the filter data hashing as the method for generating uniqueness, *only* exactly equal filters would overlap between jobs. This includes the retention configuration and others.

EasterTheBunny:
Of course unique filter names still doesn't solve the problem of stale filters after a config change. This would be more rare than node restarts at least, making the problem a little smaller.

reductionista:
Sorry for the very delayed response @EasterTheBunny, there were some last minute CCIP on Solana release blockers that came up.

> Of course unique filter names still doesn't solve the problem of stale filters after a config change. This would be more rare than node restarts at least, making the problem a little smaller.

Yeah, this is a separate issue of just making sure that UnBind is called whenever a job, workflow, etc is deleted or wanting to change the address bound to a contract. The general rule is that anything which calls Bind at some point should eventually call UnBind.  This is supposed to happen in CreateJob and DeleteJob, at least that's where we were originally calling RegisterFilter and UnRegisterFilter. Recently, we found some places where UnBind was being called during node shutdown (in `Close()`) instead, which is unacceptable... so we removed them. But that was considered a release blocker whereas re-adding the calls to UnBind in the right place is a lower priority.

The more tricky issue I think is how to make sure the filters are always unique without using any randomness.

reductionista:
> I don't really understand this though. If two jobs have the exact same filter, why not have both of them querying the same logs? It seems like duplicate work on the part of LogPoller to support two jobs wanting the same data.

There were two main reasons for changing the design from EVM to have each row in the `logs` table be associated with a specific `filter_id` on Solana. Details in this section of the design doc:
https://smartcontract-it.atlassian.net/wiki/spaces/BCF/pages/1041563791/Oct+2024+LogPoller+on+Solana+Design#Database-Schema

To summarize them: one has to do with different filters having different retention settings (corresponding in the `logs` table schema to `expires_at` field, and `sequence number` for MaxLogsKept feature). It's difficult to figure out whether a log can be deleted when you remove a filter unless each log is associated with a specific filter.  And the other has to do with the subkey paths. Each filter is allowed to have its own subkey paths.  Two plugins may want to track the same logs, but they need them indexed on different fields. You can't build a different db index for each plugin because the db schemas can only be changed at migration time (during node startup). So they all share the same index, but each store their own copy of logs.

However, even if we found a way to solve all of the problems associated with sharing a single log between filters, that wouldn't make the problem of registering many duplicate filters for the same filter go away. It would still be a bug we'd want to find some way of solving, the consequences would just be less bad.

> With the filter data hashing as the method for generating uniqueness, only exactly equal filters would overlap between jobs. This includes the retention configuration and others.

Including the retention settings and subkey paths in the hash used to generate the filter name is a good idea. It seems better than what we have, but it doesn't solve the main problem of filter name conflict which is that... if you have two jobs running, and one of them gets deleted, you need to know which filters were associated with that job and delete them (and their associated logs) without deleting any filters or logs associated with other running jobs.

Each client who calls RegisterFilter needs to use the same name for registering and unregistering its filters, without affecting any of the other clients using LogPoller. So it needs to have its own namespace.

I think we could probably do this by having whoever calls `NewContractReader` inserting a job id into the ContractReader Config just before it calls the constructor, right?  This should happen somewhere in `CreateJob`, and then the same job id needs to get passed from `DeleteJob` when it's calling `Unbind`.

PR #1222, opened 2025-04-28
pkg/solana/config: config enhancements

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-361

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/1103
- https://github.com/smartcontractkit/chainlink-common/pull/1189

Supports:
- https://github.com/smartcontractkit/chainlink/pull/17091

core ref: 901f1cbd90876144b8b37b0edeff08c31647e0bd


pkg/solana/config/config.go
```diff
@@ -9,41 +9,6 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/config"
 )
 
-// Global solana defaults.
-var defaultConfigSet = Chain{
```
jmank88, line 16: Initially I populated docs.toml from core, but then I noticed that these values were different, so I updated them. However, IIUC the core defaults were in effect, because core uses a cascading `SetFrom`, so I think they need to be switched back?
- [ ] ~switch docs.toml fields back to original core values~
Edit: These `defaultConfigSet` values are meant to take priority.

pkg/solana/config/example.toml
jmank88, line 1: - [ ] example

pkg/solana/config/toml.go
```diff
@@ -3,20 +3,34 @@ package config
...
+	"strings"
 	"time"
 
 	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/pelletier/go-toml/v2"
 	"golang.org/x/exp/slices"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/config"
-	commontypes "github.com/smartcontractkit/chainlink-common/pkg/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/config/configtest"
```
Unheilbar, line 17: The name says 'test', which is a bit confusing, since its used here outside of test-suites

.github/workflows/relay.yml
```diff
@@ -38,10 +36,8 @@ jobs:
           git diff --stat --exit-code
       - name: Install Solana CLI
         run: ./scripts/install-solana-ci.sh
-      - name: Compilation check
-        run: go test -run=xxx ./... # check compilation across tests + relayer / monitoring go code without running
```
Unheilbar, line 42: why do we drop test compilation check?

pkg/solana/config/toml.go
jmank88: Oh yeah that is historical baggage but there is a `package configdoc` now too that it be moved to.

.github/workflows/relay.yml
jmank88: What value does it add to do this before the actual run?

Unheilbar: This step will fail the job in case our integration-tests build is broken?

jmank88: Why is it better to fail on this step instead of the call where we actually invoke tests? With streaming logs, I don't think we save much time if any, and with this limited call you end up with less information that might lead you to iterate more.

Unheilbar (approved):

Unheilbar (approved):

pavel-raykov (approved):

archseer (approved):

PR #1232, opened 2025-05-12
Setup postgres db before running unit tests & interface tests

reductionista:
### Description

Set up postgres db, along with sql migrations from core repo before running relay unit tests & interface tests.
This will enable all tests which were previously being skipped due to `sqltest.SkipInMemory()`

Also:
 - refactored parsing of "core ref" from PR Body into an action that's used in 3 places.
 - fixed 2 tests which were failing: one due to a subtlety with timestamp conversion which causes it to fail when run in CI (was passing locally, I believe due to lack of a realtime clock in local env). The other was just broken, and nobody noticed since it was previously being skipped in CI.
 
JIRA:  [NONEVM-872](https://smartcontract-it.atlassian.net/browse/NONEVM-872)

[NONEVM-872]: https://smartcontract-it.atlassian.net/browse/NONEVM-872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/e2e_custom_cl.yml
+++ b/.github/workflows/e2e_custom_cl.yml
@@ -200,15 +200,11 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - name: Get core ref from PR body
-        if: github.event_name == 'pull_request'
-        run: |
-          comment="$(gh pr view https://github.com/${{ github.repository }}/pull/${{ github.event.pull_request.number }} --json body -q '.body')"
-          # shellcheck disable=SC2086
-          core_ref="$(echo $comment | grep -oP 'core ref: \K\S+' || true)"
-          if [ -n "$core_ref" ]; then
-            echo "CUSTOM_CORE_REF=${core_ref}" >> "${GITHUB_ENV}"
-          fi
+      - name: Checkout the repo
+        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+      - name: Get core ref
+        id: get-core-ref
+        uses: ./.github/actions/get-core-ref
       - name: Check if image exists
         id: check-image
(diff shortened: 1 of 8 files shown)


.github/workflows/e2e_custom_cl.yml
```diff
@@ -223,7 +219,7 @@ jobs:
         with:
           should_checkout: true
           cl_repo: smartcontractkit/chainlink
-          cl_ref: ${{ env.CUSTOM_CORE_REF || github.event.inputs.cl_branch_ref }}
+          cl_ref: ${{ steps.get-core-ref.outputs.core-ref || github.event.inputs.cl_branch_ref }}
           dep_solana_sha: ${{ github.event.pull_request.head.sha }}
           push_tag: ${{ env.CL_ECR }}:solana.${{ github.sha }}
           QA_AWS_REGION: ${{ secrets.QA_AWS_REGION }}
```
amit-momin, line 225: I can't comment further down but I think there are a couple other places that we use `Get core ref from PR body` below too

reductionista: Hmm... I see 2 more places in this file where it extracts "core ref" from the PR and writes it to `env.CUSTOM_CORE_REF`, but I'm not sure why those are there. It doesn't look like it's using it in any later step, since both of the jobs where that happens are testing already built chainlink images not checking out the source.

I also see it reading from `env.CUSTOM_CORE_REF` in `e2e_testnet_daily.yml`, but that doesn't make much sense either since it can't be triggered on a PR, only runs as a `workflow_dispatch` which already has an input `cl_branch_ref`.  I've removed that also.

amit-momin (dismissed):

prashantkumar1982 (dismissed):

chris-de-leon-cll (approved):

PR #1242, opened 2025-05-16
.github/workflows: simplify e2e timeouts

jmank88:
Specifying `timeout` and `test.timeout` is redundant, and in our case the longer `-timeout 24h` is trampling our parameterized `30m` timeouts. Here is a 6h run: https://github.com/smartcontractkit/chainlink-solana/actions/runs/15025545948/job/42225620676?pr=1222


augustbleeds (approved):

chainchad (approved):
Would it make sense to add timeouts for all of these jobs too?

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

PR #1251, opened 2025-05-21
Adding events to forwarder contract

chray-zhang:
### Description
- Adding events to forwarder contract

https://smartcontract-it.atlassian.net/browse/DF-21341
### Requires Dependencies
https://github.com/smartcontractkit/chainlink-solana/pull/1179

### DIFF
--- a/contracts/programs/keystone-forwarder/src/events.rs
+++ b/contracts/programs/keystone-forwarder/src/events.rs
@@ -23,12 +23,12 @@ pub struct InitializeEmit {
 
 #[event]
 pub struct OwnershipTransfer {
-    pub old_owner: Pubkey,
-    pub new_owner: Pubkey,
+    pub current_owner: Pubkey,
+    pub proposed_owner: Pubkey,
 }
 
 #[event]
 pub struct OwnershipAcceptance {
-    pub old_owner: Pubkey,
+    pub previous_owner: Pubkey,
     pub new_owner: Pubkey,
 }
--- a/contracts/programs/keystone-forwarder/src/events.rs
+++ b/contracts/programs/keystone-forwarder/src/events.rs
@@ -1,14 +1,14 @@
 use anchor_lang::prelude::*;
 
 #[event]
-pub struct ReportProcessedEvent {
+pub struct ReportProcessed {
     pub receiver: Pubkey,
     pub transmission_id: [u8; 32],
     pub result: bool,
 }
 
 #[event]
-pub struct ConfigSetEvent {
+pub struct ConfigSet {
     pub don_id: u32,
     pub config_version: u32,
     pub f: u8,
@@ -16,18 +16,19 @@ pub struct ConfigSetEvent {
 }
 
 #[event]
-pub struct InitializeEmitEvent {
-  pub owner: Pubkey,
-  pub authority_nonce: u8,
-  pub timestamp: i64,
+pub struct InitializeEmit {
+    pub owner: Pubkey,
+    pub authority_nonce: u8,
 }
 
 #[event]
-pub struct TransferOwnershipEvent {
-  pub new_owner: Pubkey
+pub struct OwnershipTransfer {
+    pub old_owner: Pubkey,
+    pub new_owner: Pubkey,
 }
 
 #[event]
-pub struct AcceptOwnershipEvent {
-  pub owner: Pubkey
+pub struct OwnershipAcceptance {
+    pub old_owner: Pubkey,
+    pub new_owner: Pubkey,
 }
--- a/contracts/programs/keystone-forwarder/src/events.rs
+++ b/contracts/programs/keystone-forwarder/src/events.rs
@@ -1,0 +1,33 @@
+use anchor_lang::prelude::*;
+
+#[event]
+pub struct ReportProcessedEvent {
+    pub receiver: Pubkey,
+    pub transmission_id: [u8; 32],
+    pub result: bool,
+}
+
+#[event]
+pub struct ConfigSetEvent {
+    pub don_id: u32,
+    pub config_version: u32,
+    pub f: u8,
+    pub signers: Vec<[u8; 20]>,
+}
+
+#[event]
+pub struct InitializeEmitEvent {
+  pub owner: Pubkey,
+  pub authority_nonce: u8,
+  pub timestamp: i64,
+}
+
+#[event]
+pub struct TransferOwnershipEvent {
+  pub new_owner: Pubkey
+}
+
+#[event]
+pub struct AcceptOwnershipEvent {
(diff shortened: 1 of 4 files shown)


contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -0,0 +1,33 @@
+use anchor_lang::prelude::*;
+
+#[event]
+pub struct ReportProcessedEvent {
```
augustbleeds: nit: remove the `Event` suffix because this isn't a common convention for Solana contracts

contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -0,0 +1,33 @@
...
+#[event]
+pub struct InitializeEmitEvent {
+  pub owner: Pubkey,
+  pub authority_nonce: u8,
+  pub timestamp: i64,
+}
+
+#[event]
+pub struct TransferOwnershipEvent {
+  pub new_owner: Pubkey
```
augustbleeds: typically transfer and accept ownership should also put the current owner / last owner in there

contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub config_version: u32,
+    pub f: u8,
+    pub signers: Vec<[u8; 20]>,
+}
+
+#[event]
+pub struct InitializeEmitEvent {
+  pub owner: Pubkey,
+  pub authority_nonce: u8,
+  pub timestamp: i64,
```
augustbleeds: what's the motivation behind including a timestamp? You'll be able to trace the associated time based off the block this event is included in

contracts/tests/keystone_forwarder.spec.ts
```diff
@@ -83,6 +83,13 @@ describe("keystone_storage", function () {
   });
 
   it("Is initialized!", async () => {
+    // Check event is emitted
+    const listener = program.addEventListener("InitializeEmitEvent", (event, slot) => {
+      assert.isTrue(event.owner.equals(provider.wallet.publicKey), "Owner set")
+      assert.isNotNull(event.authorityNonce)
+      assert.isNotNull(event.authorityNonce)
```
augustbleeds: event.authorityNonce check is repeated twice here

augustbleeds: you can actually test for the equality of the actual event.authorityNonce by precomputing the bump. It's a determinsitic value based on the seeds.

```
  const [forwarderAuthorityStorage, forwarderAuthorityBump] =
      PublicKey.findProgramAddressSync(
        [
          Buffer.from(anchor.utils.bytes.utf8.encode("forwarder")),
          forwarderState.publicKey.toBuffer(),
        ],
        program.programId
      );
   ```
   
   is done elsewhere in the testing file so you can do something similar here

File: contracts/tests/keystone_forwarder.spec.ts [hunk 5]
```diff
@@ -106,11 +113,29 @@ describe("keystone_storage", function () {
...
     assert.equal(actualState.version, 1, "version 1");
+    await program.removeEventListener(listener);
   });
 
   it("Transfer ownership and back", async () => {
     const proposedOwner = Keypair.generate();
 
+    // transfer ownership event emitted
+    const transferOwnershipEventPromise = new Promise((resolve, reject) => {
+      const listener = program.addEventListener("TransferOwnershipEvent", (event, slot) => {
```
augustbleeds: can you explain why this is in a Promise? Also, we never await the promise which means it's possible the promise never resolves?

augustbleeds: you should await the transferOwnershipEventPromise after the instruction is invoked.

contracts/tests/keystone_forwarder.spec.ts
augustbleeds: also, you should use the same promise pattern as below.

contracts/tests/keystone_forwarder.spec.ts
```diff
@@ -133,6 +158,19 @@ describe("keystone_storage", function () {
       "proposed owner set"
     );
 
+    const eventPromise = new Promise((resolve, reject) => {
```
augustbleeds: Because you're needing to do this pattern a couple times I'd recommend you make it in a utility function. That way we can use it in other tests as well. 

the function should accept 1. an event listener 2. callback function for encapsulating the assertion logic

contracts/tests/keystone_forwarder.spec.ts
chray-zhang: I found that if I don't wrap it in a promise then the tests won't fail on an assert failing

chray-zhang: Also because we're listening for evnts. I'm not sure if there's any other way to await the event in a listener

contracts/programs/keystone-forwarder/src/events.rs
chray-zhang: Added to both

contracts/tests/keystone_forwarder.spec.ts
chray-zhang: Kept it a if not null check because its in a weird state where the listener [initializes](https://github.com/smartcontractkit/chainlink-solana/pull/1251/files#diff-d6765c1dce1ae00e3caf0a1f9146de86def98b31b37625716051790478ea7e90R127) with the callback before the event is fire and the `actualState` is updated so its a nil pointer typescript error

contracts/tests/keystone_forwarder.spec.ts
chray-zhang: Created a new function

augustbleeds (approved):
looks good! some nits, but it's merge-ready

contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -0,0 +1,34 @@
...
+#[event]
+pub struct InitializeEmit {
+    pub owner: Pubkey,
+    pub authority_nonce: u8,
+}
+
+#[event]
+pub struct OwnershipTransfer {
+    pub old_owner: Pubkey,
+    pub new_owner: Pubkey,
```
augustbleeds: nit: current_owner and proposed_owner

contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+#[event]
+pub struct OwnershipAcceptance {
+    pub old_owner: Pubkey,
```
augustbleeds: nit: previous_owner

PR #1260, opened 2025-06-03
Zero oracle config and timeout on Forwarder Tests

chray-zhang:
Moved to : https://github.com/smartcontractkit/chainlink-solana/pull/1263


contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -290,7 +292,13 @@ fn set_oracles_config(
 
     oracles_config.config_id = get_config_id(don_id, config_version);
     oracles_config.f = f;
-    oracles_config.signer_addresses = signer_addresses.clone();
+    // Zero out the existing signer addresses array before writing new ones
+    oracles_config.signer_addresses = [[0u8; 20]; MAX_ACCTS];
```
augustbleeds, line 296: you can just use .clear() (see the arrayvec library) in order to avoid writing more than you need to

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -290,7 +292,13 @@ fn set_oracles_config(
...
+
+    // Add the new signer addresses
+    for (i, &addr) in signer_addresses.iter().enumerate() {
+        oracles_config.signer_addresses[i] = addr;
```
augustbleeds, line 300: you should be using arrayvec's .push or else the len property will not be updated (which is crucial for iteration)

contracts/programs/keystone-forwarder/src/state.rs
```diff
@@ -1,19 +1,20 @@
...
+#[account(zero_copy)]
+#[derive(bytemuck::Zeroable, bytemuck::Pod)]
+#[repr(C)]
 pub struct OraclesConfig {
-    pub config_id: u64,
+    pub config_id: u64,                              // 4 bytes
     pub f: u8,
-    pub signer_addresses: Vec<[u8; 20]>,
+    pub _padding: [u8; 7],                           // 7 bytes padding for alignment
+    pub signer_addresses: [[u8; 20]; MAX_ACCTS],    // Fixed size: 20 * MAX_ACCTS + 8 bytes
```
augustbleeds, line 12: you should be using the arrayvec library here (see https://github.com/smartcontractkit/chainlink-ccip/blob/main/chains/solana/contracts/programs/timelock/src/state/config.rs#L56 for examples on how it works)

contracts/programs/keystone-forwarder/src/state.rs
```diff
@@ -1,19 +1,20 @@
...
 }
 
 impl OraclesConfig {
-    pub const INIT_SPACE: usize = 8 + 1 + 4;
-
-    pub fn space_with_signers(num_signers: usize) -> usize {
-        Self::INIT_SPACE + (num_signers * 20)
-    }
+    // discriminator + config_id + f + padding + (address_size * max_addresses) + arrayvec_overhead
+    pub const INIT_SPACE: usize = 8 + 4 + 1 + 7 + (20 * 64) + 8;
```
augustbleeds, line 17: because the OraclesConfig is now a static size, you can use #[derive(InitSpace)] directly on it and access the INIT_SPACE field which autocomputes it

contracts/programs/keystone-forwarder/src/state.rs
chray-zhang: So apparently arrayvec does not work with anchor's zero copy account. It's because zero copy accounts must implement [Pod](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html). I think Arrayvec isn't pod because it doesn't implement Copy which Pod requires (https://github.com/bluss/arrayvec/issues/32).

augustbleeds: try removing bytemuck:Pod, bytemuck:Zeroable, and reprC. just follow the template (copy/paste) in the solana ccip repo and you should be good

PR #1261, opened 2025-06-04
Add support for buffering CCIP execute reports

amit-momin:
- Updated the TXM dependency transaction feature
  - Added support for multiple dependencies
  - Added support to track different statuses for each dependency
  - Added support to send transaction dependent on failure and silently drop on success
- Added a registry for unique buffering methods
  - Splits the payload into chunks
  - Creates all of the instructions required to write the payload into the buffer
  - Updates the account list with the new buffer PDA
  - Transforms the args to empty out the field that was written to the buffer
  - Creates a close buffer instruction used only if buffering or main transaction fail
- Added a method to broadcast buffer transactions
  - Enqueues all of the buffer transactions and tracks their individual unique transaction ID
  - Creates and queues the new main transaction with the new accounts list and transformed args
  - Marks the main transaction as dependent on all of the buffer transactions
  - Creates and queues the close buffer transaction
  - Marks the close buffer tx as dependent on the failure of the main transaction

Core PR: https://github.com/smartcontractkit/chainlink/pull/18111


pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
 		case <-time.After(backoff):
-			status, err := txm.GetTransactionStatus(waitCtx, transactionID)
-			if err != nil {
-				return status, fmt.Errorf("error fetching transaction status: %w", err)
+			txAwaitingDesiredStatus := make([]txmutils.DependencyTx, 0, len(remaining))
+			for _, meta := range remaining {
+				status, err := txm.GetTransactionStatus(waitCtx, meta.TxID)
+				// This would only happen if the transaction status has been cleared from storage
+				if err != nil || status == commontypes.Unknown {
+					// If failed to get status, considered the dependency tx as errored. Check if success was expected.
```
Farber98: What would happen if:
- desiredStatus is `finalized`
- The dependency tx got `finalized` successfully
- `retentionTimeout` was set to 0, so we don't have it in our storage

Would this be an issue? Should this be an invalid config?

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+					if !isErroredStatus(meta.DesiredStatus) {
+						txm.lggr.Debugw("dependency transaction required to be successful status but encountered errored status", "status", status, "desiredStatus", meta.DesiredStatus, "dependencyTxID", meta.TxID)
+						unexpectedStatuses++
+					}
+					// Allow status poll to continue before marking the dependent transaction as errored in case other transactions are dependent on it
+					// Returning the error early could cause unexpected behavior if the assumption is all preceding transactions are completed
```
Farber98: can you explain to me this unexpected behavior? 🙏

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+					continue
+				}
+				switch status {
+				case commontypes.Failed, commontypes.Fatal:
+					if !isErroredStatus(meta.DesiredStatus) {
```
Farber98, line 877: nit: would make sense to do something with this duplicate check? 

It's the same handling for:
- `err != nil || status == commontypes.Unknown`
- `case commontypes.Failed, commontypes.Fatal:`

```
if !isErroredStatus(meta.DesiredStatus) {
	txm.lggr.Debugw("dependency transaction required to be successful status but encountered errored status", "status", status, "desiredStatus", meta.DesiredStatus, "dependencyTxID", meta.TxID)
	unexpectedStatuses++
	// Allow status poll to continue before marking the dependent transaction as errored in case other transactions are dependent on it
	// Returning the error early could cause unexpected behavior if the assumption is all preceding transactions are completed
	continue
}
```

maybe it doesn't, but wanted to know your thoughts

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -495,6 +562,31 @@ func getLookupTableAddresses(ctx context.Context, client client.MultiClient, tab
 	return alt.Addresses, nil
 }
 
+func CalculateTxSize(tx *solana.Transaction) (int, error) {
+	if tx == nil {
+		return 0, errors.New("tx is nulll")
+	}
+	copyTx := utils.DeepCopyTx(*tx)
+
+	// Set instructions and fields that are added further downstream with arbitrary values to get an accurate tx size
```
silaslenihan, line 573: What happens if you don't set these? It's unclear to me what ComputeUnit price and limit have to do with tx size

pkg/solana/chainwriter/chain_writer.go
```diff
@@ -360,12 +362,28 @@ func (s *SolanaChainWriterService) SubmitTransaction(ctx context.Context, contra
...
+	txSize, err := CalculateTxSize(tx)
+	if err != nil {
+		return errorWithDebugID(fmt.Errorf("failed to calculate tx size: %w", err), debugID)
+	}
+
+	if txSize > MaxSolanaTxSize {
+		s.lggr.Debugw("Transaction size exceeds the Solana max", "size", txSize, "max", MaxSolanaTxSize, "tx", transactionID, "debugID", debugID)
+		// Return error if transaction too large and method to write to buffer is not provided
+		if methodConfig.BufferPayloadMethod == "" {
+			return errorWithDebugID(fmt.Errorf("transaction size %d exceeds limit %d", txSize, MaxSolanaTxSize), debugID)
```
silaslenihan:
```suggestion
			return errorWithDebugID(fmt.Errorf("transaction size %d exceeds limit %d with no buffer payload method set", txSize, MaxSolanaTxSize), debugID)
```

silaslenihan (commented):
Out of curiosity, have we figured out how billing is handled for these larger transactions that need buffers? I know we had previously decided to add a flat fee to account for ATAs but this fee could be much larger

pkg/solana/chainwriter/buffer_payload.go
```diff
@@ -0,0 +1,277 @@
...
+		IsSigner:   false,
+	})
+
+	closeBufferIx, err := ccip_offramp.NewCloseExecutionReportBufferInstruction(bufferID, bufferPDA, offrampConfigPDA, feePayer, solana.SystemProgramID).ValidateAndBuild()
+	if err != nil {
+		return nil, nil, nil, nil, fmt.Errorf("failed to build close execution report buffer instruction: %w", err)
+	}
+
+	// Transform args to clear out the report since the buffer will be used instead
+	execCallArgs.Report = []byte{}
```
silaslenihan, line 103: so do we not have to worry about setting TokenIndexes etc. when using a buffered execution?

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -778,6 +806,75 @@ func TestChainWriter_SubmitTransaction(t *testing.T) {
...
+			rw.On("LatestBlockhash", mock.Anything).Return(&rpc.GetLatestBlockhashResult{Value: &rpc.LatestBlockhashResult{Blockhash: recentBlockHash, LastValidBlockHeight: uint64(100)}}, nil).Once()
+			txID := uuid.NewString()
+
+			bufferTXM.On("Enqueue", mock.Anything, admin.String(), mock.Anything, &txID, mock.Anything).Return(nil).Once()
+
+			submitErr := bufferCW.SubmitTransaction(ctx, "buffer_payload", "execute", args, txID, bufferProgramID.String(), nil, nil)
+			require.NoError(t, submitErr)
+		})
+
+		t.Run("submits buffer transactions, main transaction, and conditional close buffer transaction if tx too large", func(t *testing.T) {
```
silaslenihan, line 902: can we add some error cases here too?

pkg/solana/chainwriter/chain_writer.go
amit-momin: Both of these get set as compute budget instructions in the tx which take up space. So when they're eventually added in the TXM, the tx goes over the limit.

pkg/solana/chainwriter/buffer_payload.go
amit-momin: We do but the `execCallArgs` here are the transformed ones from the args transform further up the process. So the token indexes should already be set. We just have to empty out the report here since we're using buffering instead

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+
+	select {
+	case txm.chSend <- msg:
+		txm.lggr.Debugw("enqueued tx after dependencies reached desired status", "id", msg.id, "dependencyTxCount", len(depMeta.DependencyTxs))
+	default:
+		txm.lggr.Errorw("failed to enqueue tx after dependencies", "queueFull", len(txm.chSend) == MaxQueueLen, "tx", msg)
+	}
+}
+
+func (txm *Txm) waitForTxStatus(ctx context.Context, depMeta txmutils.DependencyTxMeta) error {
```
silaslenihan: probably a name change makes sense
```suggestion
func (txm *Txm) waitForDependencyTxs(ctx context.Context, depMeta txmutils.DependencyTxMeta) error {
```

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+				switch status {
+				case commontypes.Failed, commontypes.Fatal:
+					if !isErroredStatus(meta.DesiredStatus) {
+						txm.lggr.Debugw("dependency transaction required to be successful status but encountered errored status", "status", status, "desiredStatus", meta.DesiredStatus, "dependencyTxID", meta.TxID)
+						unexpectedStatuses++
+						// Allow status poll to continue before marking the dependent transaction as errored in case other transactions are dependent on it
+						// Returning the error early could cause unexpected behavior if the assumption is all preceding transactions are completed
+						continue
+					}
+					txm.lggr.Debugw("dependency transaction reached desired status", "status", status, "desiredStatus", meta.DesiredStatus, "id", meta.TxID)
```
silaslenihan: Is there a case where we'd want to distinguish between failed and fatal statuses for dependency txs?

amit-momin:
> Out of curiosity, have we figured out how billing is handled for these larger transactions that need buffers? I know we had previously decided to add a flat fee to account for ATAs but this fee could be much larger

I think billing is remaining the same for now IIRC. Buffering isn't that expensive actually since we're able to close it after and recover the cost. We only end up paying for a few extra transaction's fees which is relatively cheap

pkg/solana/chainwriter/buffer_payload.go
```diff
@@ -0,0 +1,277 @@
...
+	closeBufferUUID := fmt.Sprintf("CloseBuffer-%s", uuid.NewString())
+	closeOpts := []txmutils.SetTxConfig{
+		txmutils.SetEstimateComputeUnitLimit(true),
+		// Mark close buffer transaction as dependent on the main transaction. Only send the close buffer transaction if main transaction marked as failed
+		// Main transaction would be marked as failed if any of the buffer transactions failed or if itself failed
+		txmutils.AppendDependencyTxs([]txmutils.DependencyTx{{TxID: txID, DesiredStatus: types.Failed}}),
+		// Ignore dependency errors because this transaction is expected to be dropped in the happy path
+		txmutils.SetDependencyTxMetaIgnoreError(true),
+	}
+	s.lggr.Debugw("Queuing close buffer transaction, only sends if buffer or main transcation fails", "contract", contractName, "method", method, "closeBufferTxID", closeBufferUUID, "mainTxID", txID)
```
silaslenihan, line 235: Wait why? I thought the close buffer is required to get the fee returned?

amit-momin: Execute actually automatically closes the buffer at the end so we don't have to worry about it if the main tx passes

amit-momin: I should add a comment to explain that here though

pkg/solana/txm/txm.go
amit-momin: Since the distinction between the 2 are dependent on the RPC error type, I don't think we do for this feature. We're only anticipating to track this status for clean up txs which wouldn't care one way or the other.

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+				case commontypes.Finalized, commontypes.Unconfirmed:
+					if isErroredStatus(meta.DesiredStatus) {
+						txm.lggr.Debugw("dependency transaction required to be errored status but encountered successful status", "status", status, "desiredStatus", meta.DesiredStatus, "dependencyTxID", meta.TxID)
+						unexpectedStatuses++
+						// Allow status poll to continue before marking the dependent transaction as errored in case other transactions are dependent on it
+						// Returning the error early could cause unexpected behavior if the assumption is all preceding transactions are completed
+						continue
+					}
+					if status < meta.DesiredStatus {
```
Farber98, line 895: only for `commontypes.Unconfirmed` right? would make sense to be more explicit?

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+			if len(remaining) == 0 {
+				// Return error if any of the dependency transactions were in unexpected statuses to avoid queueing the dependent transaction
+				if unexpectedStatuses > 0 {
+					return fmt.Errorf("%d transactions have unexpected statuses", unexpectedStatuses)
 				}
-				// otherwise keep polling
+				return nil
 			}
 		}
 		backoff *= 2
```
Farber98: Can't make suggestion bcs below lines are out of diff 😢 

nit: we can use go built-in min [method](https://pkg.go.dev/builtin#min) here
`backoff = min(backoff*2, maxBackoff)`

same idea applies [here](https://github.com/smartcontractkit/chainlink-solana/blob/e4951a3e4158545b07af3349ea3d864b617051b3/pkg/solana/txm/txm.go#L324-L327) as well:
`deltaT = min(deltaT*2, MaxRetryTimeMs)`

pkg/solana/txm/txm.go
```diff
@@ -820,32 +793,124 @@ func (txm *Txm) Enqueue(ctx context.Context, accountID string, tx *solanaGo.Tran
...
+	select {
+	case txm.chSend <- msg:
+		txm.lggr.Debugw("enqueued tx after dependencies reached desired status", "id", msg.id, "dependencyTxCount", len(depMeta.DependencyTxs))
+	default:
+		txm.lggr.Errorw("failed to enqueue tx after dependencies", "queueFull", len(txm.chSend) == MaxQueueLen, "tx", msg)
+	}
+}
+
+func (txm *Txm) waitForTxStatus(ctx context.Context, depMeta txmutils.DependencyTxMeta) error {
 	waitCtx, cancel := context.WithTimeout(ctx, txm.cfg.TxConfirmTimeout())
```
Farber98: should we add some guardrails here and ensure `TxConfirmTimeout` is set?

pkg/solana/txm/txm.go
amit-momin: Good question, I think we would consider that as an invalid config. The 0 `retentionTimeout` is just the default to support the existing Data Feeds behavior but it's required for CCIP. But you bring up a good point. This is something I can easily check in `waitForDependencyTxs` so I can add an error log if we notice this to surface the misconfig more easily

pkg/solana/txm/txm.go
amit-momin: Some context on the CCIP execute report buffer, it doesn't have to be explicitly allocated before writing. It's initialized automatically by the first write transaction. It's also possible to reinitialize the same buffer after it's been closed. So if we exit early here as soon as we get an error for one of the buffer tx, we'd immediately mark the main tx as errored. That would cause the close buffer tx to immediately trigger. In that scenario, there's a chance we call close buffer while other write buffer txs are in-flight. That could mean we potentially close the buffer but write buffer txs get confirmed after re-opening it.

pkg/solana/txm/txm.go
amit-momin: That's true now that Pending is its own case. It just seemed easier to keep the condition like this instead of something like `if meta.DesiredStatus == commontypes.Finalized && status == commontypes.Unconfirmed`. But I've been staring at this code for a while so if it's easier to read explicitly like above as a new reader I don't mind updating

pkg/solana/txm/txm.go
amit-momin: Good catch! 0 is a valid config for `TxConfirmTimeout` even though we don't use it in practice so definitely should have a sanity check here

pkg/solana/txm/txm.go
amit-momin: Actually I should check this even earlier in `Enqueue`. No need to even go into the dependency tx processing if retention is set to 0.

pkg/solana/txm/txm.go
Farber98: good point. Let's keep it like this then, it's easy to spot

silaslenihan (dismissed):

silaslenihan (approved):

PR #1262, opened 2025-06-05
DF-21320: Cache Program

augustbleeds:
### Description

Mainly contains new additions for the cache program, but also updates the forwarder as well.

See the [Cache README.md](https://github.com/smartcontractkit/chainlink-solana/blob/09f5d8ea6a7c990776dcb9a0969edfdfe69b9ce3/docs/data-feeds-cache/README.md)


contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,951 @@
...
+        state.proposed_owner = proposed_owner;
+
+        Ok(())
+    }
+
+    pub fn accept_ownership(ctx: Context<AcceptOwnership>) -> Result<()> {
+        let state = &mut ctx.accounts.state.load_mut()?;
+
+        state.owner = state.proposed_owner;
+        state.proposed_owner = Pubkey::default();
```
jadepark-dev: nit: can be simplified with (take() resets it to default)
```suggestion
		state.owner = std::mem::take(&mut state.proposed_owner);
```

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,951 @@
...
+    use crate::{
+        common::ANCHOR_DISCRIMINATOR,
+        event::LegacyFeedsReported,
+        state::{CacheTransmission, DecimalReport, LegacyFeedEntry, WritePermissionFlag},
+    };
+
+    use super::*;
+
+    // todo, add feed admins? who will also be a zero_copy array
+    pub fn initialize(ctx: Context<Initialize>, feed_admins: Vec<Pubkey>) -> Result<()> {
```
jadepark-dev, line 44: how many admins are we talking about here? (asking mainly for tx size limit concern)

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,951 @@
...
+        verify_feed_admin(&ctx.accounts.feed_admin, &state.feed_admins)?;
+
+        let state_key = ctx.accounts.state.key();
+
+        let data_ids_account_infos = ctx.remaining_accounts;
+
+        require!(
+            data_ids.len() == data_ids_account_infos.len(),
+            DataCacheError::ArrayLengthMismatch
+        );
```
jadepark-dev: nit:(not a big deal, but auditors tend to nitpick this)
```suggestion
        require_eq!(
            data_ids.len(), data_ids_account_infos.len(),
            DataCacheError::ArrayLengthMismatch
        );
```

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,926 @@
...
+            require!(
+                &prev_admin <= admin,
+                DataCacheError::AddressesMustStrictlyIncrease
+            );
+            state.feed_admins.push(*admin);
+            prev_admin = *admin;
+        }
+
+        // todo: can probably remove
+        state.proposed_owner = Pubkey::default();
```
jadepark-dev: Yes, this is already zeroed out :)

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,926 @@
...
+                    ],
+                    &[seeds],
+                )?;
+
+                let mut dst = curr_report_account_info.try_borrow_mut_data()?;
+                dst[..ANCHOR_DISCRIMINATOR].copy_from_slice(&DecimalReport::discriminator());
+            }
+
+            // todo: probably optional... just makes sure the account is the right type
+            DecimalReport::try_deserialize(&mut &curr_report_account_info.data.borrow()[..])?;
```
jadepark-dev, line 231: I think this is still needed if you want to verify existing account data. Probably in `else` of `if curr_report_account_info.data_is_empty()` ?

jadepark-dev (commented):
Leaving nits from first day, will continue looking into it

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+
+    // todo, add feed admins? who will also be a zero_copy array
+    pub fn initialize(ctx: Context<Initialize>, feed_admins: Vec<Pubkey>) -> Result<()> {
+        let state = &mut ctx.accounts.state.load_init()?;
+        state.owner = ctx.accounts.owner.key();
+
+        let mut prev_admin = Pubkey::default();
+        for admin in feed_admins.iter() {
+            require!(
+                &prev_admin <= admin,
```
jadepark-dev: This part allows duplicate admins. Is it possible to have duplicate in the list?

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+        Ok(())
+    }
+
+    // previews the permission accounts to be closed when
+    // calling set_decimal_feed_configs
+    // used for tx simulation only
+    // no state changes
+    pub fn preview_decimal_feed_configs<'info>(
+        ctx: Context<'_, '_, 'info, 'info, PreviewDecimalFeedConfigs<'info>>,
+        data_ids: Vec<[u8; 16]>,
```
jadepark-dev, line 290: I believe the estimated max number of data IDs was shared in the walkthrough. If the maximum value can be calculated, it would be helpful to have it documented. (It would be nice if a custom error could be returned, but since it would fail at the RPC level…)

contracts/crates/chainlink-solana-data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+use anchor_lang::prelude::*;
```
jadepark-dev, line 1: Why do we want to keep this query instructions in crates?

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+                        .chain(
+                            ordered_legacy_feed_account_infos
+                                .iter()
+                                .map(|val| val.to_account_info()),
+                        )
+                        .collect();
+
+                // payload begins with the Anchor discriminator
+                let mut payload =
+                    hash::hash("global:cache_submit".as_bytes()).to_bytes()[..8].to_vec();
```
jadepark-dev: So this is expected impl in the receiver program, I think you can make this constant to avoid runtime hashing

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+                let cache_state_key = ctx.accounts.cache_state.key();
+
+                let ix = Instruction::new_with_bytes(legacy_store.key(), &payload, metas);
+                let signer_seeds = &[
+                    b"legacy_writer",
+                    cache_state_key.as_ref(),
+                    &[ctx.accounts.cache_state.load()?.legacy_writer_nonce],
+                ];
+
+                invoke_signed(&ix, &account_infos, &[signer_seeds])?;
```
jadepark-dev: If the update fails somehow, you might want to return a specific mapped error

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+        state.proposed_owner = Pubkey::default();
+
+        Ok(())
+    }
+
+    pub fn close_decimal_reports<'info>(
+        ctx: Context<CloseDecimalReports<'info>>,
+        data_ids: Vec<[u8; 16]>,
+    ) -> Result<()> {
+        let state = &mut ctx.accounts.state.load()?;
```
jadepark-dev: It seems this doesn't have to be `mut` here? 
```suggestion
        let state = &ctx.accounts.state.load()?;
```

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+
+        Ok(())
+    }
+
+    pub fn init_decimal_reports<'info>(
+        ctx: Context<'_, '_, 'info, 'info, InitDecimalReports<'info>>,
+        data_ids: Vec<[u8; 16]>,
+    ) -> Result<()> {
+        // check feed admin here
+        let state = &mut ctx.accounts.state.load()?;
```
jadepark-dev: same here

```suggestion
        let state = &ctx.accounts.state.load()?;
```

contracts/programs/data-feeds-cache/src/lib.rs
```diff
@@ -0,0 +1,1094 @@
...
+                &[b"decimal_report", state_key.as_ref(), data_id],
+                &crate::ID,
+            );
+
+            require!(
+                &decimal_report == curr_report_account_info.key,
+                DataCacheError::AccountMismatch
+            );
+
+            // if already initialized, skip it
```
jadepark-dev: nit: this comment a bit misleading

contracts/programs/data-feeds-cache/src/lib.rs
augustbleeds: In practice, I don't think feed admins will be more than 10. I set max size to 64 though to be safe. If you bump into txs limit uou can always add more and remove them in the set_feed_admin

contracts/programs/data-feeds-cache/src/lib.rs
augustbleeds: good catch, it should not be permissible to allow duplicates. So I'll change it to a "<"

contracts/programs/data-feeds-cache/src/lib.rs
augustbleeds: I wondered if it's optional because the only way that decimal report PDA can be created is via this instruction. It's not as if the user can pass in a random account (because we verify the address generation).

contracts/programs/data-feeds-cache/src/lib.rs
augustbleeds: Sounds good, I will add that

contracts/programs/data-feeds-cache/src/lib.rs
augustbleeds: thanks, that's a great point. i just ran into a painful debugging session cause i didn't know where the error was returned from

contracts/crates/chainlink-solana-data-feeds-cache/src/lib.rs
augustbleeds: just for easy on-chain usage by developers. they can also directly import the data-feeds-cache if they want to, but I think this would be easier for them

contracts/programs/keystone-forwarder/src/common.rs
```diff
@@ -5,9 +5,15 @@ pub const ANCHOR_DISCRIMINATOR: usize = 8;
 pub const REPORT_CONTEXT_LEN: usize = 96;
 
 // our don size is directly limited by the transaction size during on_report's signature verification
-pub const MAX_ORACLES: usize = 17;
```
alecgard, line 8: Why is this changing? Does this mean that if we have a DON with 17 nodes it won't be able to transmit to via Forwarder?

contracts/programs/data-feeds-cache/src/event.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[event]
+pub struct StaleDecimalReport {
+    pub data_id: [u8; 16],
+    pub received_timestamp: u32,
+    pub latest_timestamp: u32,
+}
+
+#[event]
+pub struct DecimalReportUpdate {
```
alecgard, line 36: Minor, can we name this `DecimalReportUpdated` to match EVM? Also matches the other events we have here (past tense).

contracts/programs/keystone-forwarder/src/common.rs
augustbleeds: It changed because for f = 5, the minimum amount of oracles you need is 16 not 17. So that was a typo on my part. You can always create a DON with more than this number, but if f = 5 then we typically deploy the least amount of nodes that gets us to f. If you add more nodes offchain, you won't be able to add them on-chain so the verification logic will be incomplete. 

We could set a higher max oracles size to 31 but at that point the application payload would be variable based on the f - number of the DON. At 31 nodes, f is 10 and I think you'd have less than a byte for the application payload. I think it's better if we have DONs that are strictly a lower size for Solana where the user knows exactly the lower bound of the space they have to work with. I think the sweet spot is 16 for now. 16 oracles (f = 5) means the user has around 330 bytes for theirselves. And it's a robust enough DON as well at f = 5.

smickovskid (approved):

PR #1268, opened 2025-06-13
Update CCIP execute args transform to use on-chain account derivation

amit-momin:
Updated the CCIP execute args transform method to use on-chain account derivation to build out the accounts and lookup tables lists.

Core PR: https://github.com/smartcontractkit/chainlink/pull/18171
Chainlink CCIP PR: https://github.com/smartcontractkit/chainlink-ccip/pull/1018


Farber98 (dismissed):

pkg/solana/chainwriter/transform_registry_test.go
```diff
@@ -523,39 +557,128 @@ func verifyTxOpts(t *testing.T, options []txmutils.SetTxConfig, exec bool, overh
...
-		LookupTable:          lookupTablePubkey,
-		// set all accounts as writable
-		WritableIndexes: [2]ag_binary.Uint128{{Endianness: ag_binary.LE, Lo: math.MaxUint64, Hi: math.MaxUint64}},
+func mockExecuteAccountDerivation(t *testing.T, rw *clientmocks.ReaderWriter, offrampStr string, sourceChainSel uint64, userMessagingAccounts []solana.PublicKey, ttAccounts []tokenTransferAccounts, logicReceiver solana.PublicKey, lookupTables []solana.PublicKey) {
+	recentBlockHash := solana.Hash{}
+	rw.On("LatestBlockhash", mock.Anything).Return(&rpc.GetLatestBlockhashResult{Value: &rpc.LatestBlockhashResult{Blockhash: recentBlockHash, LastValidBlockHeight: uint64(100)}}, nil).Once()
+	mockGatherBasicInfoStage(t, rw, offrampStr, sourceChainSel)
+	mockMainAccountListStage(t, rw, offrampStr, userMessagingAccounts, logicReceiver, ttAccounts)
+	mockRetrieveLUTStage(t, rw, offrampStr, ttAccounts)
+	mockTokenTransferStages(t, rw, offrampStr, ttAccounts, lookupTables)
```
PabloMansanet: Other than the Static token transfer stage, which is important for token indices, the rest of the stages here are a bit of an implementation detail onchain, and it would be good not to couple them to this repo, so they don't have to be maintained if the stages change onchain.

It's OK if these are part of the mock mainly for organization purposes, but maybe we can leave a comment saying that it's fine if these drift from their onchain counterparts.

amit-momin: Good point, I added the note

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -5,28 +5,35 @@ import (
...
 	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
 	txmutils "github.com/smartcontractkit/chainlink-solana/pkg/solana/txm/utils"
 )
 
-const MandatoryExecuteAccounts = 12
-
-func FindTransform(id string) (func(context.Context, client.MultiClient, any, solana.AccountMetaSlice, map[string]map[string][]*solana.AccountMeta, string, uint32, []txmutils.SetTxConfig) (any, solana.AccountMetaSlice, []txmutils.SetTxConfig, error), error) {
+func FindTransform(id string) (func(context.Context, client.MultiClient, logger.Logger, any, solana.AccountMetaSlice, map[solana.PublicKey]solana.PublicKeySlice, map[string]map[string][]*solana.AccountMeta, solana.PublicKey, string, uint32, []txmutils.SetTxConfig, string) (any, solana.AccountMetaSlice, map[solana.PublicKey]solana.PublicKeySlice, []txmutils.SetTxConfig, error), error) {
```
winder: Might be worth making some sort of `TransformHandler` type for this beast.

pkg/solana/chainwriter/buffer_payload.go
```diff
@@ -119,7 +122,12 @@ func buildBufferExecutionReportIx(bufferID []byte, reportLen uint32, chunkPayloa
 	if ixErr != nil {
 		return nil, fmt.Errorf("failed to build buffer instruction: %w", ixErr)
 	}
-	return ix, nil
+	data, dataErr := ix.Data()
+	if dataErr != nil {
+		return nil, fmt.Errorf("failed to encode instruction data: %w", ixErr)
+	}
+	solanaIx := solana.NewInstruction(programID, ix.Accounts(), data)
```
toblich, line 129: Qq, given that the buffer ix is always for the offramp program, is there a reason to allow overriding the invoked program? Or is it maybe just a convenience for some tests? Anyway, there's no problem with doing it, I was just curious about the rationale behind

pkg/solana/chainwriter/transform_registry.go
amit-momin: Good call! Added in the latest commit

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -80,35 +90,164 @@ func CCIPExecuteArgsTransform(ctx context.Context, client client.MultiClient, ar
...
+	var messageTokenData [][]byte
+	if len(message.TokenAmounts) > 0 {
+		tokenTransfers = make([]ccip_offramp.TokenTransferAndOffchainData, 0, len(message.TokenAmounts))
+		if len(argsTransformed.ExtraData.DestExecDataDecoded) != len(message.TokenAmounts) {
+			return nil, nil, nil, nil, fmt.Errorf("unexpected number of DestExecData encountered. expect the same number as token transfers %d, got %d", len(message.TokenAmounts), len(argsTransformed.ExtraData.DestExecDataDecoded))
+		}
+		// If message contains token transfers, extract offchain token data for the message if any exists
+		if len(report.OffchainTokenData) > 0 {
+			messageTokenData = report.OffchainTokenData[0]
+		}
```
toblich: I know we expect the OffchainTokenData to have a single entry, as there is up to 1 token transfer per message today (and 1 message per report), but it feels like something that should be validated here and throw an unexpected number of... error if the assumption does not hold

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -316,7 +415,154 @@ func appendTokenTransferAccounts(tokenAccountsRequired bool, accounts solana.Acc
...
+		deriveAccountsIxRaw.AccountMetaSlice = append(deriveAccountsIxRaw.AccountMetaSlice, accountsToAskWith...)
+		deriveAccountsIx, err := deriveAccountsIxRaw.ValidateAndBuild()
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to build derive execute accounts instruction: %w", err)
+		}
+		deriveAccountsIxData, err := deriveAccountsIx.Data()
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to encode account derivation instruction data: %w", err)
+		}
+		deriveAccountsSolIx := solana.NewInstruction(offramp, deriveAccountsIx.Accounts(), deriveAccountsIxData)
```
toblich, line 457: Similar to a previous question, is there a reason to override the program id of an instruction meant for the offramp with the offramp ID again? There's no negative impact from this anyway other than extra lines of code, I'm just curious 👀

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -356,3 +602,48 @@ func getPoolChainConfig(sourceChainSelector, destTokenAddress []byte, poolProgra
...
+	// Use FillBytes to fill the array with big-endian data, zero-padded
+	bi.FillBytes(paddedBytes)
+
+	// Reverse the array for little-endian encoding
+	for i, j := 0, len(paddedBytes)-1; i < j; i, j = i+1, j-1 {
+		paddedBytes[i], paddedBytes[j] = paddedBytes[j], paddedBytes[i]
+	}
+
+	return paddedBytes
+}
```
toblich, line 655: This seems ok to me, but out of curiosity: didn't we have other places where we're doing the same thing, so we could just invoke an existing function instead of creating a new one?

toblich (dismissed):
LGTM! I just left a couple of small nitpicky comments, but feel free to dismiss any of them if you think they don't apply

pkg/solana/chainwriter/buffer_payload.go
amit-momin: Ya this was because of tests. Doing `ccip_offramp.SetProgramID(offrampPubKey)` caused errors in tests when it was called multiple times. Then doing something like `ccip_offramp.ProgramID = offrampPubKey` caused some race issues in tests so I just went with setting it manually like this

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -316,7 +415,154 @@ func appendTokenTransferAccounts(tokenAccountsRequired bool, accounts solana.Acc
...
+		return nil, nil, nil, fmt.Errorf("failed to parse offramp address: %w", err)
+	}
+	config, _, err := state.FindOfframpConfigPDA(offramp)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to calculate offramp config address: %w", err)
+	}
+	var derivedAccounts, accountsToAskWith solana.AccountMetaSlice
+	lookupTableMap := make(map[solana.PublicKey]solana.PublicKeySlice)
+	tokenIndexes := []uint8{}
+	mandatoryAccountsLen := cap(ccip_offramp.NewExecuteInstructionBuilder().AccountMetaSlice)
```
winder, line 440: The v1 transformer contained this check:
```
	if len(accounts) < mandatoryAccountsLen {
		return nil, nil, nil, nil, fmt.Errorf("encountered unexpected number of accounts, expected at least %d, got %d", mandatoryAccountsLen, len(accounts))
```

Is it still needed?

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya it was the same issue as above

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya good eye. We use the same logic in the plugin code in the core repo but pulling this from there introduces a circular dependency. I was thinking about moving this into some utility in chainlink-common but forgot to clean it up. Think it might be better to tackle this as a follow up clean up (if we ever do) at this point in the review cycle.

pkg/solana/chainwriter/transform_registry.go
amit-momin: This check was needed in v1 since we relied on the CW to calculate the mandatory accounts before args transform. Since we use account derivation in v2, we actually don't have any accounts going into the args transform. But for sanity, I'll add the check back to make sure accounts is empty

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya good point! If I'm making this assumption, I should make some noise if it doesn't hold.

pkg/solana/chainwriter/transform_registry.go
```diff
@@ -80,35 +90,169 @@ func CCIPExecuteArgsTransform(ctx context.Context, client client.MultiClient, ar
...
+	var tokenTransfers []ccip_offramp.TokenTransferAndOffchainData
+	var tokenReceiver solana.PublicKey
+	var messageTokenData [][]byte
+	if len(message.TokenAmounts) > 0 {
+		tokenTransfers = make([]ccip_offramp.TokenTransferAndOffchainData, 0, len(message.TokenAmounts))
+		if len(argsTransformed.ExtraData.DestExecDataDecoded) != len(message.TokenAmounts) {
+			return nil, nil, nil, nil, fmt.Errorf("unexpected number of DestExecData encountered. expect the same number as token transfers %d, got %d", len(message.TokenAmounts), len(argsTransformed.ExtraData.DestExecDataDecoded))
+		}
+		// If message contains token transfers, extract offchain token data for the message if any exists
+		if len(report.OffchainTokenData) != len(report.Messages) {
```
winder: nit: this works because you previously confirmed `len(report.Messages) ==1`, but it looks weird here since the check looks like it's to ensure `report.OffchainTokenDat[0]` is valid.
```suggestion
		if len(report.OffchainTokenData) != 1 {
```

winder (dismissed):

pkg/solana/chainwriter/transform_registry.go
amit-momin: Ya I was trying to reduce where we'd need to make changes if we did support multiple messages. It might not be obvious that `OffchainTokenData` length was tied to `Messages` length otherwise. But it does look out of place here. I'll make this change and just leave a comment instead

toblich (dismissed):

winder (dismissed):

winder (approved):

toblich (approved):

PR #1282, opened 2025-06-30
[PLEX - 1529] - Implement GetChainInfo

ilija42:
core ref: 73888c355ce58735cb1a5136880b89a956cd772b

### Resolves Dependencies

https://github.com/smartcontractkit/chainlink/pull/18397


Unheilbar (dismissed):

jmank88 (dismissed):

pkg/solana/chain.go
```diff
@@ -413,6 +415,26 @@ func (c *chain) GetChainStatus(ctx context.Context) (types.ChainStatus, error) {
 	}, nil
 }
 
+func (c *chain) GetChainInfo(_ context.Context) (types.ChainInfo, error) {
+	chainIDHash := c.id
+	chainEnvName := c.cfg.ChainID
```
prashantkumar1982, line 420: There's no guarantee this will be chain env name.
It may as well be the chain id hash.
It depends on what was set in the toml config.
Previously it was the env var, but we at ccip launch time, we were changing it to the hash.

Might be more reliable to fetch env name from the network name.

ilija42: c.ID is either hash or whatever c.cfg is, c.cfg can be hash or chainEnvName or something invalid, if cfg is not invalid this code will work as intended. Where can I get network name from?

prashantkumar1982: See here:
https://github.com/smartcontractkit/chain-selectors/blob/611fd5777fe3789610be5daf3a79aed982c3fd24/selectors_solana.yml#L1-L11

The name field has the env name too.

I thought this is why you created the selectors.ExractEnvName function for?

ilija42: Yes, but to get this name field I already have to have the chainIDHash, in which case c.ID will have the chainEnvName value. Look at the NewChain constructor

smickovskid (approved):

PR #1284, opened 2025-07-01
add deterministic filter name

Unheilbar:
[JIRA](https://smartcontract-it.atlassian.net/browse/NONEVM-2064)

Context: Because a unique filter name is generated on every bind call, each node restart results in duplicate filter creation. Since Solana LP saves log per every filter it satisfies, this causes unnecessary duplication and bloats the database over time.

Solution: filter name will be derived based on filter params such as:
1. eventSig
2. chain specific name
3. address
4. event name
5. subkeypaths
6. subkeys
7. namespace+generic name (no changes here)

If we call bind with  existing filter params then starting_block, retention, max_block_kept will be updated.


pkg/solana/chainreader/event_read_binding.go
```diff
@@ -138,11 +139,39 @@ func (b *eventReadBinding) update(ctx context.Context) error {
...
+		}
+	}
+	for _, sub := range b.indexedSubKeys.subKeys {
+		for _, key := range sub {
+			data = append(data, key...)
+		}
+	}
+
+	// include retention, starting block, max logs kept
+	tail := fmt.Sprint(b.filter.filter.Retention, b.filter.filter.StartingBlock, b.filter.filter.MaxLogsKept)
```
dhaidashenko: Update of retention, StartingBlock, and MaxLogsKept should not cause the creation of a new filter.

dhaidashenko (changes requested):

pkg/solana/chainreader/event_read_binding.go
dhaidashenko: Also we need a PR with migration to cleanup the db

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -138,11 +139,35 @@ func (b *eventReadBinding) update(ctx context.Context) error {
...
+			data = append(data, []byte(key)...)
+		}
+	}
+	for _, sub := range b.indexedSubKeys.subKeys {
+		for _, key := range sub {
+			data = append(data, key...)
+		}
+	}
+
+	hash := sha3.New256().Sum(data)
```
ilija42: this isn't the correct usage Sum(data) doesn’t feed data into the hash, it appends the empty input digest to data you will always end up with `data + 167 255 198 248 191 30 215 102 81 193 71 86 160 97 214 98 245 128 255 77 228 59 73 250 130 216 ...`
You probably want something like this
```
h := sha3.New256()
h.Write(data)
digest := h.Sum(nil)
```

ilija42 (changes requested):

ilija42 (dismissed):

pkg/solana/chainreader/event_read_binding.go
```diff
@@ -138,11 +139,40 @@ func (b *eventReadBinding) update(ctx context.Context) error {
 		return nil
 	}
 
-	newName := fmt.Sprintf("%s.%s.%s", b.namespace, b.genericName, uuid.NewString())
+	newName := b.deriveName()
 
 	return b.filter.Update(ctx, b.reader, newName)
 }
 
+func (b *eventReadBinding) deriveName() string {
```
ilija42, line 147: does it make sense to add a short unit test for this?

Unheilbar: added

Unheilbar: added

pkg/solana/chainreader/event_read_binding_test.go
```diff
@@ -52,6 +57,43 @@ func TestBind(t *testing.T) {
...
+		// name shoudln't have changed (address the same)
+		name2 := reader.deriveName()
+		require.Equal(t, name1, name2)
+
+		require.NoError(t, reader.Bind(ctx, address2))
+
+		// name should have changed
+		name3 := reader.deriveName()
+		require.NotEqual(t, name2, name3)
+		lpSource.AssertExpectations(t)
```
mateusz-sekara, line 98: Can we have some more real examples in the test? I would love to understand from tests what would be the actual final filter name - for instance hardcoded in the test code

```go
reader.Bind(sth)
requre.Equal("Commit.0xasdf.0xasdf", reader.deriveName())
```

Unheilbar: added

mateusz-sekara (approved):

ilija42 (approved):

dhaidashenko (approved):

PR #1286, opened 2025-07-02
PLEX-583: add cache config to write targeter

augustbleeds:

### DIFF
--- a/pkg/solana/write_target/target_strategy.go
+++ b/pkg/solana/write_target/target_strategy.go
@@ -213,7 +213,6 @@ func (t *targetRequest) GetRemainingAccounts(ctx context.Context, client client.
 	if t.Config.CacheDetails != nil {
 		// assume that the receiver is the cache program
 		cacheProgram := t.Receiver
-		remainingAccounts = make([]solana.AccountMeta, 2*len(t.Config.CacheDetails.FeedIds))
 
 		cacheStateKey, err := solana.PublicKeyFromBase58(t.Config.CacheDetails.State)
 		if err != nil {
@@ -228,6 +227,30 @@ func (t *targetRequest) GetRemainingAccounts(ctx context.Context, client client.
 		if cacheStateAccount.Value == nil {
 			return nil, fmt.Errorf("cache state account does not exist %v", cacheStateKey)
 		}
+
+		remainingAccounts = []solana.AccountMeta{
+			solana.AccountMeta{
+				PublicKey:  cacheStateKey,
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  cacheProgram, // legacy store omitted
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  cacheProgram, // legacy feed config omitted
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  cacheProgram, // legacy writer omitted
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  solana.SystemProgramID,
+				IsWritable: false,
+			},
+		}
+		derivedAccounts := make([]solana.AccountMeta, 2*len(t.Config.CacheDetails.FeedIds))
 
 		// derive pdas and check existence on-chain
 		for i, feedId := range t.Config.CacheDetails.FeedIds {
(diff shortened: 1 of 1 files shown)


pkg/solana/write_target/target_strategy.go
```diff
@@ -185,6 +192,111 @@ type targetRequest struct {
 	Inputs   Inputs
 }
 
+func (t *targetRequest) GetRemainingAccounts(ctx context.Context, client client.Reader, forwarderAuthority solana.PublicKey) ([]solana.AccountMeta, error) {
+	if len(t.Config.RemainingAccounts) > 0 && t.Config.CacheDetails != nil {
+		return nil, fmt.Errorf("only one of 'remaining_accounts' or 'cache_details' should be specified")
+	}
+
+	var remainingAccounts []solana.AccountMeta
+	for _, acc := range t.Config.RemainingAccounts {
```
Unheilbar, line 201: if len(t.Config.RemainingAccount) > 0 {
...
return
}

pkg/solana/write_target/target_strategy.go
```diff
@@ -185,6 +192,111 @@ type targetRequest struct {
...
+		}
+
+		remainingAccounts = append(remainingAccounts, solana.AccountMeta{
+			PublicKey:  key,
+			IsWritable: acc.IsWritable,
+		})
+	}
+
+	if t.Config.CacheDetails != nil {
+		// assume that the receiver is the cache program
```
Unheilbar, line 214: probably put  the whole if block under another func and return it

pkg/solana/write_target/target_strategy.go
```diff
@@ -185,6 +192,111 @@ type targetRequest struct {
...
+			if writeFlagAccount.Value == nil {
+				return nil, fmt.Errorf("write flag account %v does not exist for data id %v", writeFlagKey, feedId)
+			}
+
+			// write flag accounts go after all the decimal report accounts
+			remainingAccounts[len(t.Config.CacheDetails.FeedIds)+i] = solana.AccountMeta{PublicKey: writeFlagKey, IsWritable: false}
+		}
+
+	}
+
```
Unheilbar, line 320: if we reached the end return err "either 'remaining_accounts' or 'cache_details' should be provided"

pkg/solana/write_target/target_strategy.go
```diff
@@ -185,6 +192,136 @@ type targetRequest struct {
...
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  cacheProgram, // legacy feed config omitted
+				IsWritable: false,
+			},
+			solana.AccountMeta{
+				PublicKey:  cacheProgram, // legacy writer omitted
+				IsWritable: false,
+			},
```
Unheilbar, line 247: it's not clear to me, why do we add three exactly the same accounts to the remaining?

pkg/solana/write_target/target_strategy.go
```diff
@@ -185,6 +192,136 @@ type targetRequest struct {
...
+			solana.AccountMeta{
+				PublicKey:  solana.SystemProgramID,
+				IsWritable: false,
+			},
```
Unheilbar, line 251: SystemProgramID is already included in NewReportInstruction constructor

augustbleeds: https://github.com/smartcontractkit/chainlink-solana/pull/1262/files#diff-5135d90ef0327bf1db52f2cd77107c50d4c2cebe15a020a62256b9aa8ba21497R1618

The system program was included in the account context originally. I think this is no longer needed so we'll need to update the contract later

augustbleeds: But generally, as is it right now we also have to pass it into the remaining accounts because it's used by the receiver program. You could technically pass the system program from the forwarder report to a receiver on report, but you'd have to know the index ahead of time and whatnot.

pkg/solana/write_target/target_strategy.go
augustbleeds: https://github.com/solana-foundation/anchor/pull/2101 , the Optioanl account will be None if the program id is passed in

pkg/solana/write_target/target_strategy.go
augustbleeds: you actually don't need the conditional since you already have the check at the top. if only cache details is provided, this loop is never used. if only the remaining account is provided, the cache details conditional is never used.

PR #1300, opened 2025-07-17
IsLoadBalancedRPC flag

yashnevatia:
core ref: rpc-proxy-flag


pkg/solana/logpoller/worker/worker.go
```diff
@@ -40,6 +40,11 @@ type worker struct {
 func (w *worker) Do(ctx context.Context, job Job) {
 	defer func() {
 		// put itself back on the queue when done
+		// Recover from panics and treat them as job failures
+		if rec := recover(); rec != nil {
+			panicErr := fmt.Errorf("job panicked: %v", rec)
+			w.Lggr.Errorw("job panicked during execution", "job", job.String(), "panic", rec, "error", panicErr)
```
amit-momin: Following the pattern below I think we should also pass this job over the failed job chan and return early here
```
w.FailedJobs <- failedJob{Job: job, Err: panicErr}
return
```

DylanTinianov (dismissed):

Unheilbar (dismissed):

PR #1317, opened 2025-08-08
Enable CR Tests

yashnevatia:
- The testing interface in chainlink-common works for evm as you can do a fresh deployment for every test and as a result get a fresh `cr.binding`. This isolates every test and the test assertions assume that isolation.
- We cannot replicate that for solana due to how the contracts in solana are compiled. We can deploy only one contract. And that means we use the same cr.binding across every test, which means the assumption of isolation will not hold.
- So I have lumped event based tests together and adapted them to fit the solana world while not compromising on testing scope.
- For tests related to `TestStruct` **reading** and **writing**, I have kept them separate, same as chainlink-common, but adapted them to accommodate for sending in test_idx and list_idx as a parameter to differentiate PDAs of different tests running in parallel.


integration-tests/relayinterface/chain_components_test.go
```diff
@@ -61,6 +64,13 @@ const (
 
 var trueVal = true
 
+// StoreTestStruct is used to send data on chain
+// TestIdx is hardcoded per test in chainReaderConfig
+type StoreTestStruct struct {
+	Data    TestStruct // test data
+	ListIdx uint8      // identifier to differentiate between different test structs in the same test
```
silaslenihan, line 71: Question: Why can't we just use the existing testIdx identifier here?

yashnevatia: https://github.com/smartcontractkit/chainlink-framework/blob/357f3d8240f8199f88b429aefc52066141e37896/tools/evm-chain-bindings/examples/basic/contracts/ChainReaderTester.sol#L79

Each test, can send in multiple test structs, and then it expects to read them all.
The testIdx differentiates between tests but not structs within the same test.

integration-tests/relayinterface/chain_components_test.go
```diff
@@ -226,7 +241,346 @@ func RunChainComponentsSolanaTests[T WrappedTestingT[T]](t T, it *SolanaChainCom
...
+					resultAnyContract, testDataAnyContract := result[bound], batchCallEntry[bound]
+					returnValue, err := resultAnyContract[i].GetResult()
+					assert.NoError(t, err)
+					assert.Contains(t, resultAnyContract[i].ReadName, MethodTakingLatestParamsReturningTestStruct)
+					assert.Equal(t, testDataAnyContract[i].ReturnValue, returnValue)
+				}
+			},
+		},
+		{
+			Name: "Solana" + ContractReaderBatchGetLatestValueDifferentParamsResultsRetainOrder,
```
amit-momin: These tests look different from each other but the name of this one and the previous one are the same

amit-momin: Just noticed this uses the secondary contract so think we just need to update the name to be `ContractReaderBatchGetLatestValueDifferentParamsResultsRetainOrderMultipleContracts` here

yashnevatia: [done](https://github.com/smartcontractkit/chainlink-solana/pull/1317/commits/3476e418b3db21d68c1e746c71c42c88a2198c32)

amit-momin (approved):

DylanTinianov (approved):

silaslenihan (approved):

PR #1329, opened 2025-08-20
Add CCIP provider to Solana Relayer with Default Accessor

amit-momin:


amit-momin:
Opting for the changes in this PR instead: https://github.com/smartcontractkit/chainlink-solana/pull/1333

PR #1333, opened 2025-08-26
Add CCIP provider to Solana Relayer with Solana Accessor

amit-momin:


pkg/solana/ccip/chainaccessor/config.go
```diff
@@ -0,0 +1,156 @@
...
+
+	var cfg router.Config
+	err = a.client.GetAccountDataBorshInto(ctx, configPDA, &cfg)
+	if err != nil {
+		return ccipocr3.OnRampDynamicConfig{}, fmt.Errorf("failed to get fee quoter config account: %w", err)
+	}
+	return ccipocr3.OnRampDynamicConfig{
+		FeeQuoter:              cfg.FeeQuoter.Bytes(),
+		ReentrancyGuardEntered: false,
+		MessageInterceptor:     []byte{}, // TODO: unimplemented on Solana?
```
archseer: yep, this value can be blank

pkg/solana/ccip/chainaccessor/config.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return ccipocr3.OnRampDestChainConfig{
+		SequenceNumber:   destChain.State.SequenceNumber,
+		AllowListEnabled: destChain.Config.AllowListEnabled,
+		Router:           routerAddr.Bytes(),
+	}, nil
+}
+
+func (a *SolanaAccessor) getCurseInfo(_ context.Context) (ccipocr3.CurseInfo, error) {
```
archseer: I think you do need to implement this on Solana, it should fetch data from RMNRemote

pkg/solana/ccip/chainaccessor/solana_accessor.go
```diff
@@ -0,0 +1,312 @@
...
+		return fmt.Errorf("address is unexpected length to be solana public key %d, expect %d", len(contractAddress), solana.PublicKeyLength)
+	}
+	addr := solana.PublicKeyFromBytes(contractAddress)
+	a.bindingsMu.Lock()
+	defer a.bindingsMu.Unlock()
+	a.bindings[contractName] = addr
+	return nil
+}
+
+// Solana as source chain methods
```
archseer, line 214: We also need to implement all of these, TON is still unimplemented

pkg/solana/ccip/chainaccessor/config.go
```diff
@@ -0,0 +1,305 @@
...
+
+	offrampRefAddress, err := a.getOfframpReferenceAddresses(ctx, offrampAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	a.lggr.Debugw("getOffRampSourceChainConfigs", "sourceChainSelectors", sourceChainSelectors, "offrampAddr", offrampAddr.String())
+
+	var sourceChainConfigs = make(map[ccipocr3.ChainSelector]ccipocr3.SourceChainConfig, len(sourceChainSelectors))
+	for _, selector := range sourceChainSelectors {
```
archseer, line 123: You can use an `errgroup`  to fetch these in parallel, we just need to `SetLimit()` to limit concurrency

pkg/solana/ccip/chainaccessor/config.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// getOnRampDynamicConfig retrieves dynamic configuration from the on-ramp contract
+func (a *SolanaAccessor) getOnRampDynamicConfig(ctx context.Context) (ccipocr3.OnRampDynamicConfig, error) {
+	routerAddr, err := a.getBinding(consts.ContractNameRouter)
+	if err != nil {
+		return ccipocr3.OnRampDynamicConfig{}, fmt.Errorf("failed to get binding for router: %w", err)
+	}
+
+	configPDA, _, err := state.FindConfigPDA(routerAddr)
```
archseer: I think it would be nice to have a separate map on the accessor for caching PDA addresses, we only need to recompute them any time `Sync` is called

pkg/solana/ccip/chainaccessor/config.go
amit-momin: Is it ok to keep these as separate RPC calls? I was thinking about eventually using the multiple account read call to get all of these in one go. I'd just have to configure the solana codec to decode the account data bytes which might be a little more complicated

pkg/solana/ccip/chainaccessor/config.go
amit-momin: Is calculating PDAs that expensive of an operation? Shouldn't be hard to add though. I'll keep a note to add it once I have things running

pkg/solana/ccip/chainaccessor/config.go
```diff
@@ -0,0 +1,305 @@
...
+	return newSigners
+}
+
+func convertTransmittersType(transmitters [16][32]uint8) [][]byte {
+	newTransmitters := make([][]byte, 0, len(transmitters))
+	for _, transmitter := range transmitters {
+		newTransmitters = append(newTransmitters, transmitter[:])
+	}
+	return newTransmitters
+}
```
archseer: You probably need to truncate these to `ConfigInfo.N`? Though `n` technically determines only signer len, but they should match. It's up to 16 values, but some of them might be zeroed/empty

pkg/solana/ccip/chainaccessor/config.go
archseer: It's not super expensive (like on chain where we do need to cache it) but the `FindPDA` variants do hash the inputs in a loop with a incrementing bump until they find a result that's not on the ed25519 curve

pkg/solana/ccip/chainaccessor/config.go
archseer: Ah right, I didn't realize `getMultipleAccounts` returns the body as well 👍🏻 Does that call fail if one of the accounts doesn't exist though?

You should be able to skip the codec package altogether:

```go
for _, account := range result.Value {
  var sourceChain offramp.SourceChain
  bin.NewBorshDecoder(value.Data.GetBinary()).Decode(sourceChain)
}
```

pkg/solana/ccip/chainaccessor/solana_accessor.go
```diff
@@ -0,0 +1,627 @@
...
+	offrampAddr, err := a.getBinding(consts.ContractNameOffRamp)
+	if err != nil {
+		return nil, fmt.Errorf("offramp not bound: %w", err)
+	}
+
+	expressions := []query.Expression{
+		logpoller.NewAddressFilter(offrampAddr),
+		logpoller.NewEventSigFilter(logpollertypes.NewEventSignatureFromName(consts.EventNameCommitReportAccepted)),
+		query.Timestamp(uint64(ts.Unix()), primitives.Gte), // nolint:gosec // G115: timestamp is always positive
+		query.Confidence(primitives.Finalized),             // solana log poller only operates with finalized confidence
```
archseer, line 454: Can we do the same optimization as on TON? Where we query only commit reports with a merkle root: https://github.com/smartcontractkit/chainlink-ton/blob/7e96e1d25490b3a2d1656c2a07291d0d2f0dbdc8/pkg/ccip/chainaccessor/ton_accessor.go#L421

On TON this requires extra code since the prefix is 1 bit, but on Solana the byte right before the root will be `0`/`1` to encode so it should be easier to query (https://borsh.io/#pills-specification)

amit-momin: Oh didn't realize we wanted to skip price only commits here. Should be possible with some changes to the log poller but in the meantime should I be filtering out reports without merkle roots in this method?

ogtownsend: This is called by the exec plugin and also the commit report cache (which filters out reports right after calling this function). The side effect of filtering in the query is that we won't update the `latestFinalizedReportTimestamp` as often which could lead to refetching the same time/block ranges: https://github.com/smartcontractkit/chainlink-ccip/blob/main/execute/internal/cache/commit_report_cache.go#L154

An improvement in the future would be to update `CommitReportsGTETimestamp()` to return a block number or timestamp on any successful fetch to update the cache (if it's kept in future versions), and continue filtering in the query

pkg/solana/ccip/chainaccessor/cache.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type rmnRemotePDACache struct {
+	curse solana.PublicKey
+}
+
+func newPDACache() pdaCache {
+	return pdaCache{
+		offrampCache:     newOfframpPDACache(),
+		offrampCacheMu:   sync.RWMutex{},
```
archseer: Should be enough to just group all of these under the same mutex, since updates will be rare once the initial sync concludes

pkg/solana/ccip/chainaccessor/config.go
amit-momin: I think it returns null for the particular account so we don't have to worry about the whole call failing. I may just need to make sure we batch this in groups of 100 though according to the docs. https://solana.com/docs/rpc/http/getmultipleaccounts

pkg/solana/ccip/chainaccessor/cache.go
```diff
@@ -0,0 +1,337 @@
...
+
+type pdaCache struct {
+	// Note: we might need to update this in the future to map[string][]address.Address
+	// to support multi-bind addresses for the price aggregator contract: smartcontractkit/chainlink-ccip@main/pkg/contractreader/extended.go#L77-L79
+	bindings       map[string]solana.PublicKey
+	offrampCache   offrampPDACache
+	feeQuoterCache feeQuoterPDACache
+	routerCache    routerPDACache
+	rmnRemoteCache rmnRemotePDACache
+	cacheMu        sync.RWMutex
```
archseer, line 21: Nit: another good candidate for removing the mutex and using `sync.Map` for the bindings:

> The sync.Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow

We can investigate this change later though

archseer (approved):

PR #1352, opened 2025-09-29
V2 DataFeeds1 SDK

augustbleeds:
### Description
* Updating SDK to version 2, where direct reads is the only Chainlink approved pattern (product-decision)
* new function `read_feed_v2`
* Designed to make data types opaque for backwards compatibility with future account types. Not only by what is accessible to end user using the crate, but read_feed_v2 takes in an ~AccountInfo~ data slice
* note: ~like Pyth, the SDK is tied to anchor version. So if we have users on different anchor versions we just create a 2.x.x version where x.x is the minor and patch versions of that anchor version. (ex: 2.31.1 for anchor 0.31.1 and 2.29.x for anchor 0.29.x). This one is 2.31.1 because that is the latest version of anchor and the one that the customer will use.~
* note: relies on solana-program version:  we really only have the solana-program dependency in order to use Pubkey field. Otherwise we wouldn't need it. ~Kept solana-program version at 1.x.x  in order to avoid bumping up tooling dependencies.~ However this SDK can be used with any solana-program versions because the function only accepts a RefCell
* note: we can assume no users are using CPI anymore due to deprecation (see https://docs.chain.link/data-feeds/solana)

~TODO: update ci to build anchor 0.31.1~


contracts/crates/chainlink-solana/Cargo.toml
```diff
@@ -15,6 +15,8 @@ name = "chainlink_solana"
 default = []
 
 [dependencies]
-solana-program = "1.17.25"
-borsh = "0.9.1"
-borsh-derive = "0.9.1"
+anchor-lang  = {version = "0.31.1" }
```
archseer: Is depending on Anchor really necessary? It forces all downstream consumers to depend on Anchor. We deliberately avoided pulling it in when we released the crate

archseer (changes requested):

contracts/crates/chainlink-solana/src/lib.rs
```diff
@@ -3,27 +3,61 @@
...
-    program::invoke,
-    program_error::ProgramError,
-    pubkey::Pubkey,
-};
+use crate::data_feeds_v1_store::{Transmission, Transmissions, HEADER_SIZE, TRANSMISSIONS_DISCRIMINATOR};
 
-// The library uses this to verify the keys
-solana_program::declare_id!("HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny");
+pub(crate) mod data_feeds_v1_store {
+    use anchor_lang::prelude::{borsh::{BorshDeserialize, BorshSerialize}, *};
```
archseer: You can remove the anchor_lang dep by just directly importing `borsh`, anchor just re-exports these: https://docs.rs/borsh/1.5.1/borsh/index.html

contracts/crates/chainlink-solana/Cargo.toml
archseer: Pull in the changes from here: https://github.com/smartcontractkit/chainlink-solana/pull/1271/files

contracts/crates/chainlink-solana/src/lib.rs
```diff
@@ -39,73 +73,192 @@ pub struct Round {
...
+        let account = mock_account_info(
+            &key,
+            true,
+            true,
+            &mut lamports,
+            &mut buffer[..],
+            &owner,
+        );
+
+        let feed = read_feed_v2(account.try_borrow_data().unwrap()).unwrap();
```
archseer: I'd create separate namespaces: data feeds::v2::read() vs v1. That way the v1 module can be marked as deprecated

contracts/crates/chainlink-solana/Cargo.toml
```diff
@@ -15,6 +15,9 @@ name = "chainlink_solana"
 default = []
 
 [dependencies]
-solana-program = "1.17.25"
-borsh = "0.9.1"
-borsh-derive = "0.9.1"
+solana-program = { version = "1.17.25" }
```
archseer: Why the downgrade? We wanted to release a new library version with a newer solana program version (1.17 is EOL). Is it a CI issue?

archseer (commented):
Looks good to me, we just need to sort out the solana program version

archseer (dismissed):

sdk/chainlink-solana/Cargo.toml
```diff
@@ -15,6 +15,9 @@ name = "chainlink_solana"
 default = []
 
 [dependencies]
-solana-program = "1.17.25"
-borsh = "0.9.1"
-borsh-derive = "0.9.1"
+solana-program = { version = "2.1.10" }
```
archseer, line 18: One more tweak you might want to make, specifying `2.1` will make this crate work on any `2.x` Solana version from 2.1 onwards. That would make it work with 2.1 and 2.3

archseer: We might have to do something for 3.x too: https://docs.rs/solana-program/3.0.0/solana_program/index.html

chray-zhang (approved):

PR #1358, opened 2025-10-07
Track Tx Sigs in TXM

yashnevatia:

### DIFF
--- a/pkg/solana/txm/pendingtx.go
+++ b/pkg/solana/txm/pendingtx.go
@@ -59,6 +59,8 @@ type PendingTxContext interface {
 	IsTxReorged(sig solana.Signature, currentState utils.TxState) (string, bool)
 	// GetPendingTx returns the pendingTx for the given ID if it exists
 	GetPendingTx(id string) (pendingTx, error)
+	// GetTransactionSig returns the signature of the transaction for the given ID if it exists
+	GetTransactionSig(id string) (solana.Signature, error)
 }
 
 // finishedTx is used to store info required to track transactions to finality or error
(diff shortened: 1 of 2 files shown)


Unheilbar (approved):

dhaidashenko (approved):

pkg/solana/txm/pendingtx.go
```diff
@@ -680,6 +686,15 @@ func (c *pendingTxContext) GetPendingTx(id string) (pendingTx, error) {
 	return tx, nil
 }
 
+func (c *pendingTxContext) GetTransactionSig(id string) (solana.Signature, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	if tx, exists := c.finalizedErroredTxs[id]; exists {
+		return tx.signatures[0], nil
```
amit-momin, line 693: If we just want to return a single signature then I don't think we have to store all of the relevant signatures for a tx in the `finishedTx` struct. I assume we only want the relevant signature that corresponds to the tx that landed on chain.

pkg/solana/txm/pendingtx.go
```diff
@@ -470,6 +473,7 @@ func (c *pendingTxContext) OnFinalized(sig solana.Signature, retentionTimeout ti
 		finalizedTx := finishedTx{
 			state:       utils.Finalized,
 			retentionTs: time.Now().Add(retentionTimeout),
+			signatures:  tx.signatures,
```
amit-momin, line 476: Don't think we have to store all of the tx signatures here. Probably just `sig` we pass when calling `OnFinalized`. That should be the relevant signature that was finalized on chain

PR #1360, opened 2025-10-07
Track only final sig

yashnevatia:

### DIFF
--- a/pkg/solana/txm/pendingtx.go
+++ b/pkg/solana/txm/pendingtx.go
@@ -78,7 +78,7 @@ type pendingTx struct {
 type finishedTx struct {
 	retentionTs time.Time
 	state       utils.TxState
-	signatures  []solana.Signature
+	sig         solana.Signature
 }
 
 type txInfo struct {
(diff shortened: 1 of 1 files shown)


pkg/solana/txm/pendingtx.go
```diff
@@ -690,7 +688,11 @@ func (c *pendingTxContext) GetTransactionSig(id string) (solana.Signature, error
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 	if tx, exists := c.finalizedErroredTxs[id]; exists {
-		return tx.signatures[0], nil
+		if !tx.sig.IsZero() {
+			return tx.sig, nil
+		} else {
+			return solana.Signature{}, ErrTransactionNotFound
```
amit-momin: Think it'd help to explicitly call out here that the sig was zero and maybe include the tx state so we know if the caller is trying to call this for an errored tx

pkg/solana/txm/pendingtx.go
```diff
@@ -690,7 +688,11 @@ func (c *pendingTxContext) GetTransactionSig(id string) (solana.Signature, error
...
+		}
```
amit-momin: nit: just go convention stuff but think it'd be better to write this like
```
tx, exists := c.finalizedErroredTxs[id]
if !exists {
  return solana.Signature{}, ErrTransactionNotFound
}
if tx.sig.IsZero() {
  return solana.Signature{}, fmt.Errorf(...)
}
return tx.sig, nil
```

amit-momin (approved):

PR #1364, opened 2025-10-15
Add new TXM and Block History Estimator metrics

amit-momin:
- Added a compute unit price metric to the Block History Estimator
- Added a fee bump metric to the TXM
- Cleaned up contexts around the codebase


yashnevatia (dismissed):

pkg/solana/fees/block_history.go
```diff
@@ -310,3 +339,8 @@ func (bhe *blockHistoryEstimator) populateCache(ctx context.Context, loadBatch,
 	bhe.cache.storedBlockRange = bhe.cache.storedBlockRange[excessBlocks:]
 	return nil
 }
+
+func (bhe *blockHistoryEstimator) recordComputeUnitPrice(ctx context.Context, avgOfMedians ComputeUnitPrice) {
+	promBHEComputeUnitPrice.WithLabelValues(bhe.chainID).Set(float64(avgOfMedians))
+	bhe.computeUnitPrice.Record(ctx, int64(avgOfMedians), metric.WithAttributes(attribute.String("chainID", bhe.chainID))) //nolint:gosec // compute unit prices cannot exceed int64 max
```
toblich: Nitpick: the CU price is actually a `u64` (https://github.com/anza-xyz/agave/blob/67412607f511ded3770031280b6aaf10607713fc/sdk/src/compute_budget.rs#L48-L50), so it is theoretically possible for it to exceed the `int64` range. Is there any chance you can change the gauge type to reflect this? 

I know that that exceeding the `int64` max would be a very high price, so it's very unlikely to ever happen, but I'd still advocate for either having the metric reflect the actual type/range of the value, or at least have the conversion code be fail-proof (even if you just want to floor any higher price back to `int64::max`, that's fine by me)

toblich (commented):
LGTM in general, with just one nitpick thing about the price metric type

pkg/solana/fees/block_history.go
amit-momin: Ya I was thinking it was unrealistic so I just casted but good point. We only have int64 and float64 gauges so I'll just cap the price to int64 max but I'll add a warn log in there to call some attention to it

amit-momin: Actually I just converted it to a float64 gauge. Realized the prom metric was using float also anyways

toblich (dismissed):

toblich (approved):

PR #1376, opened 2025-10-31
Update LogPoller retry logic to drop jobs after max retry threshold

amit-momin:
### Context
Solana LogPoller had the concept of a max retry count (set to 6) but it would only determine when to log a Critical message instead of an Error. It would allow jobs to retry indefinitely. If we encounter an RPC issue for a sustained period, the retry job count can balloon and eventually overwhelm the RPC once it is functioning as expected again. 

### Solution
Updated the retry logic to drop jobs once they reach the max retry count threshold. Reaching that point for any job likely required manual intervention. Dropping the retry jobs allow us to fix the issue in a more controlled environment without potentially spiking traffic on our RPCs and taking them down.


pkg/solana/logpoller/worker/worker.go
```diff
@@ -222,22 +222,25 @@ func (g *Group) runRetryQueue(ctx context.Context) {
 
 				wait := calculateExponentialBackoff(min(retry.count, g.maxRetryCount))
 				if retry.count > g.maxRetryCount {
-					g.lggr.Criticalf("job %s failed %d times in a row, next retry in %s. Resolution most likely requires manual intervention. Errors: %s", failedAttempt.Job, retry.count, wait, errors.Join(retry.errs...))
-				} else {
-					g.lggr.Errorf("retrying job %s in %s", failedAttempt.Job, wait)
+					g.lggr.Criticalf("job %s exceeded max retry count %d. Resolution most likely requires manual intervention. Errors: %s", failedAttempt.Job, g.maxRetryCount, errors.Join(retry.errs...))
```
toblich, line 233: Agreed here :+1: An improvement on this would probably be to just implement a circuit breaker on the RPC, but I get that it may not be trivial to add (or it should be a responsibility of the RPC proxy).

So, my opinion is:
- We should try to auto-recover whenever possible, which is what we were doing prior to this change, but it was too aggressive.
- To make this less aggressive but still easy to implement, we should set a higher ceiling on the exponential backoff. 
  - I'm thinking something in the ~5 mins range (I think it is 12s today IIRC), what do you think? If needed, we can also tweak after how many retries it gets dropped, we can do some math depending on how much traffic we expect RPCs to handle safely.
- With this new change, if things get bad enough, jobs just get dropped (to not saturate RPCs further), and we're pretty sure to need manual intervention. We should be alerted on that, so we can be proactive.

toblich (commented):
The main change LGTM, but I suggest also tweaking the exponential backoff (and potentially the max retry count) together with this. The circuit breaker might be ideal but may not be trivial to implement

pkg/solana/logpoller/worker/worker.go
amit-momin: Ya good point. Some of the comments in this part of the code made it seem like dropping after max retries was the expected behavior so I assumed there was maybe some math done to land on 6. But think even in past incidents we relied on retrying more than that so that value is maybe invalidated now. Also I agree on increasing the exponential backoff cap.
For the alerting, I already created a [ticket](https://smartcontract-it.atlassian.net/browse/NONEVM-2948) under the monitoring epic to add something in here so we know when jobs are getting dropped. Think it might make sense to do that separately since it might require some reworking to setup metrics outside of the ORM in the LogPoller.

toblich (approved):

PR #1382, opened 2025-11-12
PLEX-1731 implementation SolanaService

Unheilbar:
Supports [chainlink-common](https://github.com/smartcontractkit/chainlink-common/pull/1643)


pkg/solana/solana_service.go
```diff
@@ -0,0 +1,536 @@
...
+func (ss *solanaService) QueryTrackedLogs(ctx context.Context, filterQuery []query.Expression,
+	limitAndSort query.LimitAndSort) ([]*commonsol.Log, error) {
+	lp := ss.chain.LogPoller()
+	// TODO derive query name from filterQuery
+	logs, err := lp.FilteredLogs(ctx, filterQuery, limitAndSort, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to filter logs: %w", err)
+	}
+
+	res := make([]*commonsol.Log, len(logs))
```
silaslenihan: Found a bug with this while working on my integration tests: 
```suggestion
	res := make([]*commonsol.Log, 0, len(logs))
```

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+		}
+	}
+
+	var txs []commonsol.TransactionWithMeta
+	if n := len(block.Transactions); n > 0 {
+		txs = make([]commonsol.TransactionWithMeta, n)
+		for i, tx := range block.Transactions {
+			var perTxBT *commonsol.UnixTimeSeconds
+			if tx.BlockTime != nil {
+				t := commonsol.UnixTimeSeconds(int64(*block.BlockTime))
```
silaslenihan: should this be `tx.BlockTime`?

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+	}
+
+	f, err := convertFilter(req)
+	if err != nil {
+		return err
+	}
+
+	err = lp.RegisterFilter(ctx, f)
+	if err != nil {
+		return fmt.Errorf("failed to register fitler: %w", err)
```
silaslenihan:
```suggestion
		return fmt.Errorf("failed to register filter: %w", err)
```

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+}
+
+func convertAccountResult(acc *rpc.GetAccountInfoResult) *commonsol.GetAccountInfoReply {
+	if acc == nil {
+		return nil
+	}
+
+	var a *commonsol.Account
+	if acc.Value != nil {
+		acc.Value.Data.GetBinary()
```
silaslenihan: is this doing anything?

pkg/solana/solana_service_test.go
```diff
@@ -0,0 +1,291 @@
...
+	for j := range s {
+		s[j] = i
+	}
+	return s
+}
+
+func csig(i byte) commonsol.Signature { return commonsol.Signature(sig(i)) }
+
+func uint64Ptr(v uint64) *uint64    { return &v }
+func uint64PtrVal(v uint64) *uint64 { return &v }
```
silaslenihan: can get rid of one of these

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+	queryName := deriveNameFromFilterQuery(filterQuery)
+	logs, err := lp.FilteredLogs(ctx, filterQuery, limitAndSort, queryName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to filter logs: %w", err)
+	}
+
+	res := make([]*commonsol.Log, 0, len(logs))
+	for _, l := range logs {
+		res = append(res, &commonsol.Log{
+
```
silaslenihan:
```suggestion
```

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+	return res, nil
+}
+
+func deriveNameFromFilterQuery(filter []query.Expression) string {
+	var address string
+	var eventSig string
+
+	for _, expr := range filter {
+		if expr.IsPrimitive() {
```
silaslenihan, line 197: So if `IsPrimitive()` is false, we're just going to return "-" here? Is that the expected outcome or should we error if name and eventSig are nil

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,741 @@
...
+		return nil
+	}
+	out := &commonsol.TransactionResultEnvelope{}
+	data := tx.Transaction.GetData()
+	out.AsDecodedBinary = commonsol.Data{
+		Content:  data.Content,
+		Encoding: commonsol.EncodingType(data.Encoding),
+	}
+	if pr, err := tx.Transaction.GetTransaction(); err != nil {
+		out.AsParsedTransaction = convertTransaction(pr)
```
silaslenihan: little confused why we call `convertTransaction()` if there is an error. Is this the intended outcome or should we return nil / an error here?

pkg/solana/solana_service.go
Unheilbar: good point, I actually picked up the approach from evmService impl. Do we assume that LogTrigger filters always have either Address or EventSig or Both?

silaslenihan: It will always have both coming from LogTrigger

Unheilbar: Cool, it shouldn't be used outside of LogTrigger, so I will add a validation

yashnevatia (dismissed):

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,748 @@
...
+
+func convertDataBytesOrJSON(obj *rpc.DataBytesOrJSON, enc commonsol.EncodingType) *commonsol.DataBytesOrJSON {
+	if obj == nil {
+		return nil
+	}
+	if enc == "" {
+		enc = commonsol.EncodingJSON // default
+	}
+	var txJSON []byte
+	if b, err := json.Marshal(obj); err == nil {
```
yashnevatia: err not being handled here ?

Unheilbar: yeah you are right, if the caller called for json we should either return an error or fallback to binary. Pushing fixes now

pkg/solana/solana_service.go
```diff
@@ -0,0 +1,806 @@
...
+	commonsol "github.com/smartcontractkit/chainlink-common/pkg/types/chains/solana"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
+	solprimitives "github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives/solana"
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/retry"
+	logpollertypes "github.com/smartcontractkit/chainlink-solana/pkg/solana/logpoller/types"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/txm/utils"
+)
+
+type solanaService struct {
+	commontypes.UnimplementedSolanaService
```
silaslenihan, line 24: this should be removed now right?

Unheilbar: nope, its done on purpose so updating solanaService in common in the future won't block other people on bumping common in cl-solana.

yashnevatia (approved):

PR #1383, opened 2025-11-12
adding reportFail on forwarder with go tests

yashnevatia:
[Blocked on figuring out write target flow]
### Description

 

- Added a new function called report_failure
- It takes exactly the same context as report
- report_failure
  - verifies signature
  - verifies account hash
  - proceeds to mark execution as failed

https://smartcontract-it.atlassian.net/browse/PLEX-1927


contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
 
         Ok(())
     }
+
+    pub fn report_failure<'info>(
+        ctx: Context<'_, '_, '_, 'info, Report<'info>>,
+        data: Vec<u8>,
```
PabloMansanet, line 314: Nit: pass as &[u8] unless you have a good reason to take ownership. As far as I can see all uses of this are through a slice, so no need to commit to a vec here.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
...
+    ) -> Result<()> {
+        require!(report_size_ok(&data), ForwarderError::InvalidReport);
+
+        // get config
+        let oracles_config = ctx.accounts.oracles_config.load()?;
+
+        verify_report_signatures(&data, &oracles_config)?;
+
+        let raw_report_len = data.len() - REPORT_CONTEXT_LEN;
```
PabloMansanet: Whenever possible always use safe subtraction methods; `checked_sub` or `saturating_sub`.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
...
+/// Verifies that the report has valid signatures from the configured oracles
+#[inline(never)]
+pub fn verify_report_signatures(data: &[u8], oracles_config: &OraclesConfig) -> Result<()> {
+    let num_signatures = data[0] as usize;
+
+    require_gte!(
+        num_signatures,
+        (oracles_config.f + 1) as usize,
+        ForwarderError::InvalidSignatureCount
+    );
```
PabloMansanet, line 392: Nit: maybe just `gt` instead of `gte` and removing the +1 is clearer.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
...
+
+    // extract signatures
+    let data = &data[1..];
+    let total_signature_len: usize = SIGNATURE_LEN * num_signatures;
+
+    let signatures: &[u8] = &data[..total_signature_len];
+    // raw_report | report context
+    let data = &data[total_signature_len..];
```
PabloMansanet: It's a bit more idiomatic to use [split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at), that way you can't get an index mismatch.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
...
+
+    // Build the preimage the same way the OCR keyring does:
+    // SHA256( [u8(len(raw_report))] || raw_report || ctx)
+    let mut preimage = vec![0u8; 1 + data.len()];
+
+    let raw_report_len = data.len() - REPORT_CONTEXT_LEN;
```
PabloMansanet: Ditto on subtractions.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +294,103 @@ pub mod keystone_forwarder {
...
+    // OCR keyring also does not error on overflow
+    let raw_report_len_u8: u8 = raw_report_len as u8;
```
PabloMansanet, line 408: micro-nit, but you don't need the type hint here if you have the cast.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +320,90 @@ pub mod keystone_forwarder {
 
         Ok(())
     }
+
+    pub fn report_failure<'info>(
+        ctx: Context<'_, '_, '_, 'info, Report<'info>>,
+        data: Vec<u8>,
+    ) -> Result<()> {
+        require!(report_size_ok(&data), ForwarderError::InvalidReport);
+
```
augustbleeds, line 317: so the intent is that anyone can send this report failure as long as the report is valid?

contracts/programs/keystone-forwarder/src/lib.rs
augustbleeds: failure may have been set previously by a node so we should set it back to false

contracts/programs/keystone-forwarder/src/state.rs
```diff
@@ -46,4 +46,7 @@ pub struct ExecutionState {
     pub transmission_id: [u8; 32],
     // until failure states are reported by the write target, success will always be true
     pub success: bool,
+    // having an explicit failure state allows us to differentiate between a failed transmission and one
+    // not yet initiated
+    pub failure: bool,
```
augustbleeds: Two options here:

1. 
You can tell if the transmission is being initialized for the first time because the transmission_id and transmitter will all be default 0 values. 

so for example, this will let you differentiate between what a "false" success means and one which was just initialized (so no report beforehand)

2. 
but to make it cleaner we can use enums so that 

enum Status {
    NotReported, // 0
    Success,     // 1
    Failure,     // 2
}

when you initialize the execution status for the first time it'll show NotReported.

contracts/programs/keystone-forwarder/src/lib.rs
```diff
@@ -334,6 +320,90 @@ pub mod keystone_forwarder {
...
+        let oracles_config = ctx.accounts.oracles_config.load()?;
+        let num_signatures = data[0] as usize;
```
augustbleeds: when it comes to extracting the info from the byte array, there's redundancy between report_failure and report. there's lots of indexing back and forth so it's not straightforward.

would be nice to make this cleaner. 

perhaps creating a struct where you can initialize it with the byte array and call methods on it to get the info you're interested in

let r = ReportArray::init(data)
r.get_num_signatures();
...
r.get_all() // or just return all info you want in one method

contracts/programs/keystone-forwarder/src/lib.rs
yashnevatia: added code to check report failure can only be called by configured transmitter

contracts/programs/keystone-forwarder/src/lib.rs
yashnevatia: changed this to enum like you suggested below.

contracts/programs/keystone-forwarder/src/state.rs
yashnevatia: excellent suggestion, using enum

contracts/programs/keystone-forwarder/src/lib.rs
yashnevatia: so i was trying to simply pull out the replicated code into a function.
but its somehow blowing up the stack even though i pass data by reference.
i fear it will be worse if we init a new ReportArray.
let me try and pass it by reference and see why is the stack blowing up by printing the stack size after each step.

yashnevatia: extracted code to common function

PR #1386, opened 2025-11-18
Remove TxHash from RampMessageHeader in SolanaAccessor

amit-momin:


pkg/solana/ccip/chainaccessor/event.go
```diff
@@ -252,9 +252,8 @@ func (a *SolanaAccessor) convertCCIPMessageSent(logs []logpollertypes.Log, onram
 				DestChainSelector:   ccipocr3.ChainSelector(event.Message.Header.DestChainSelector),
 				SequenceNumber:      ccipocr3.SeqNum(event.Message.Header.SequenceNumber),
 				Nonce:               event.Message.Header.Nonce,
-				TxHash:              logs[i].TxHash.ToSolana().String(),
 				OnRamp:              ccipocr3.UnknownAddress(onrampAddr.Bytes()),
-				// MsgHash: , TODO: need to set this field also?
+				// TxHash:              logs[i].TxHash.ToSolana().String(), // Populating TxHash causes inconsistent state with non-LOOPP
```
archseer: Is the preference to remove this field? We could also just switch nodes to LOOP and keep the extra metadata

amit-momin: I think we're still confirming if we want to remove this or not but I just opened this preemptively in case we do. But I think the issue arises when we're enabling in mainnet and we have a mixed state of NOPs. Although it sounds like they're adding this field for EVM too for CCTPv2 so not sure how they'll manage this issue then

Unheilbar (approved):

PR #1388, opened 2025-11-19
use chain_writer_utils

yashnevatia:
core ref: pkg-chain-submitTx

I basically moved everything out of chain_writer that submitTx uses and put it into chain_writer_utils (we can rename this).
And then i just called the implementation of the SubmitTx function from both chain.go and chain_writer.go


pkg/solana/chain.go
```diff
@@ -728,3 +730,33 @@ func solanaValidateBalance(ctx context.Context, reader client.Reader, from solan
 	}
 	return nil
 }
+
+type ChainProgramConfig struct {
+	Programs map[string]chainwriterutils.ProgramConfig `json:"programs"`
+}
+
+func (c *chain) SubmitTransaction(ctx context.Context, programConfig ChainProgramConfig, contractName, method string, args any, transactionID string, toAddress string) error {
+	contractConfig, exists := programConfig.Programs[contractName]
```
Unheilbar, line 739: hmm, what's the point of accepting a map with program configs and a contract name, just to resolve program config? Why not accept config as an argument

pkg/solana/chain.go
```diff
@@ -728,3 +730,33 @@ func solanaValidateBalance(ctx context.Context, reader client.Reader, from solan
...
+	if !exists {
+		return fmt.Errorf("failed to find program config for contract name: %s", contractName)
+	}
+
+	// Parse programs and create codec with proper encoder definitions
+	_, encoder, err := chainwriterutils.ParseProgramsToCodec(programConfig.Programs)
```
Unheilbar, line 745: Same thing here, we call one endpoint here, so we need just one encoder?

PR #1396, opened 2025-11-26
add ATA creation utils

Unheilbar:

### DIFF
--- a/pkg/solana/chainwriter/helpers.go
+++ b/pkg/solana/chainwriter/helpers.go
@@ -1,6 +1,7 @@
 package chainwriter
 
 import (
+	"context"
 	"crypto/sha256"
 	_ "embed"
 	"encoding/binary"
@@ -11,8 +12,11 @@ import (
 	"strings"
 
 	"github.com/gagliardetto/solana-go"
+	associatedtokenaccount "github.com/gagliardetto/solana-go/programs/associated-token-account"
 
+	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/client"
 )
 
 type TestArgs struct {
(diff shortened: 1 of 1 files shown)


pkg/solana/chainwriter/helpers.go
```diff
@@ -192,6 +196,68 @@ func traversePath(data any, path []string) ([]any, error) {
 		return nil, errors.New("unexpected type encountered at path: " + path[0])
 	}
 }
+func GetOrCreateATAIx(
+	ctx context.Context,
+	client client.MultiClient,
+	mint, owner, payer solana.PublicKey,
```
amit-momin, line 202: I think we'll also want a `tokenProgram` `solana.PublicKey` in the signature too

pkg/solana/chainwriter/helpers.go
```diff
@@ -192,6 +196,68 @@ func traversePath(data any, path []string) ([]any, error) {
...
+) (ata solana.PublicKey, ix solana.Instruction, err error) {
+	// Derive ATA
+	ata, _, err = solana.FindAssociatedTokenAddress(owner, mint)
```
amit-momin, line 205: The naming is kind of confusing but I think we'll want the `tokens.FindAssociatedTokenAddress` here instead. The import being `"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/tokens"`

pkg/solana/chainwriter/helpers.go
```diff
@@ -192,6 +196,68 @@ func traversePath(data any, path []string) ([]any, error) {
 		return nil, errors.New("unexpected type encountered at path: " + path[0])
 	}
 }
+func GetOrCreateATAIx(
+	ctx context.Context,
+	client client.MultiClient,
+	mint, owner, payer solana.PublicKey,
+) (ata solana.PublicKey, ix solana.Instruction, err error) {
+	// Derive ATA
```
amit-momin, line 204: Maybe it's overkill but in our existing ChainWriter code we also validate that the mint is a valid one before continuing. Maybe we can add that here too for safety. https://github.com/smartcontractkit/chainlink-solana/blob/acc1c8e69992ed2c4887e93e22a29a63ee5b4a45/pkg/solana/chainwriter/ata_creation.go#L61-L69

pkg/solana/chainwriter/helpers.go
```diff
@@ -192,6 +196,68 @@ func traversePath(data any, path []string) ([]any, error) {
...
+	if err == nil {
+		return ata, nil, nil
+	}
+
+	if err != nil && !errors.Is(err, rpc.ErrNotFound) {
+		return solana.PublicKey{}, nil, err
+	}
+
+	// Build ATA creation since Probe doesn't exists
+	ix = associatedtokenaccount.NewCreateInstruction(
```
amit-momin, line 226: I think we want to stick to the tokens package and use `tokens.CreateAssociatedTokenAccount` here instead also

PR #1418, opened 2026-01-14
[ARCH-332] KMS/TXM support

connorwstein:
Analogue to https://github.com/smartcontractkit/chainlink-evm/pull/328


pkg/solana/keys/tx.go
```diff
@@ -95,23 +97,75 @@ func CreateTxKey(ks keystore.Keystore, name string) (*TxKey, error) {
...
 }
 
+// GetTxKeysOption configures GetTxKeys behavior.
+type GetTxKeysOption func(*getTxKeysOptions)
+
+type getTxKeysOptions struct {
+	noPrefix bool
+}
+
+// WithNoPrefix disables adding the solana/tx prefix to key names.
```
connorwstein, line 107: noting this approach is cleaner/simpler than LoadTxKey in chainlink-evm. I plan to go back and update evm.

pkg/solana/keys/tx.go
```diff
@@ -95,23 +97,75 @@ func CreateTxKey(ks keystore.Keystore, name string) (*TxKey, error) {
...
+
+		// If no prefix, sanity check key type (for KMS-backed keystores)
+		if options.noPrefix {
+			if key.KeyInfo.KeyType != keystore.Ed25519 {
+				return nil, errors.New("tried to load a non-Ed25519 key: " + key.KeyInfo.Name)
+			}
+		}
+
+		// If no names are provided and we're using prefix, filter only solana keys
+		if !options.noPrefix && len(names) == 0 && !path.HasPrefix(prefixPath) {
```
connorwstein, line 162: critical filtering needed for multi-family keystores. Need to port same fix to EVM

pkg/solana/keys/core_keystore.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+var _ core.Keystore = &TxKeyCoreKeystore{}
+
+type TxKeyCoreKeystore struct {
+	ks interface {
+		keystore.Reader
+		keystore.Signer
+	}
+	cache           map[string]string
```
huangzhen1997, line 20: Don't have the full context here for the key store, both would write the same value to the cache, but any potential race condition concurrent write from txm?

pkg/solana/keys/core_keystore.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(keys) == 0 {
+		return nil, errors.New("no keys found")
+	}
+	for _, key := range keys {
+		if key.Address().String() == account {
+			s.cache[account] = key.KeyPath().String()
+			signReq := keystore.SignRequest{
```
huangzhen1997: nit: could be the same as line 71, without creating this var. Could also refactor for duplicate code where cache hit from line 71 to 77

huangzhen1997 (dismissed):
lgtm, just a question and a nit comment

pkg/solana/keys/core_keystore.go
connorwstein: I think its probably possible we have concurrent Sign calls, couldn't hurt to add a lock for that, will do

huangzhen1997 (approved):
lgtm!

PR #1419, opened 2026-01-15
Splitting out into commoncodec

yashnevatia:
- This is the first PR in a series of PRs.
- The aim of the series is to add support for multiple anchor versions to our codec lib
- This is what the final state will look like https://github.com/smartcontractkit/chainlink-solana/pull/1408
- This PR simply splits out the anchor version agnostic components out of the codec package and puts them in the commoncodec package.
- Discussion [Doc](https://docs.google.com/document/d/1Ek6VDZEp_8lwGW9fxV3P6KGvTPCXD9HUligfGy768qI/edit?tab=t.0#heading=h.ww5zhpr4mlw6)

PR #1420, opened 2026-01-15
Anchor version (anchor > 0.3x) support using codecv2 

yashnevatia:
- This is the second PR in a series of PRs (first PR https://github.com/smartcontractkit/chainlink-solana/pull/1419)
- The aim of the series is to add support for multiple anchor versions to our codec lib
- This is what the final state will look like https://github.com/smartcontractkit/chainlink-solana/pull/1408
- This PR adds codecv2 which uses the anchor-go repo for anchor > 0.3x idl parsing and definitions
- Discussion [Doc](https://docs.google.com/document/d/1Ek6VDZEp_8lwGW9fxV3P6KGvTPCXD9HUligfGy768qI/edit?tab=t.0#heading=h.ww5zhpr4mlw6)


pkg/solana/codecv2/solana.go
```diff
@@ -0,0 +1,375 @@
...
+	switch vv := def.Ty.(type) {
+	case *anchoridl.IdlTypeDefTyStruct:
+		return asStruct(refs, name)
+	case *anchoridl.IdlTypeDefTyEnum:
+		variants := vv.Variants
+		if !variants.IsAllSimple() {
+			return name, nil, fmt.Errorf("%w: variants are not supported", commontypes.ErrInvalidConfig)
+		}
+		return name, refs.builder.Uint8(), nil
+	// TODO: I only see these being used with chain_reader, skipping
```
amit-momin: We've technically officially migrated to the solana accessor as of 2.31.2-ccip but the feature flag is still there to switch between it and ChainReader. Not sure when that'll be removed but if a NOP accidentally disables LOOPP this could break the CR functionality for them right? Could we still support this just in case?

pkg/solana/codecv2/solana_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"testing"
+
+	anchoridl "github.com/gagliardetto/anchor-go/idl"
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codecv2"
+	solcommoncodec "github.com/smartcontractkit/chainlink-solana/pkg/solana/commoncodec"
+)
+
+// TODO: do this for idls in testutils folder
```
amit-momin: Is this left for the next phase? Just making sure this didn't get left behind accidentally

yashnevatia: Had added those tests.
Did not remove the comment.
TY.

pkg/solana/codecv2/solana.go
yashnevatia: I think we this code path will only be used by CCIP when do an anchor upgrade.
When we do an anchor upgrade, the ChainReader feature flag will have to removed because ChainReader will completely break after the upgrade.
So even if a NOP disables LOOPP mode, this code path should not be hit.
Hope that makes sense.
Thank you for the review @amit-momin

amit-momin: Gotcha, I think in that case we're ok. We may just want to call it out to them in case they were planning to leave the flag in for an extra release or something just in case.

amit-momin (dismissed):

pkg/solana/codecv2/solana.go
```diff
@@ -0,0 +1,375 @@
...
+		return name, nil, fmt.Errorf("named type %q not found", name)
+	}
+	// as opposed to codecv1, namedType.Ty is an interface instead of a concrete type
+	// hence we cannot access def.Ty.Kind or def.Ty.Fields directly
+	var structType anchoridl.IdlTypeDefTyStruct
+	switch vv := namedType.Ty.(type) {
+	case *anchoridl.IdlTypeDefTyStruct:
+		structType = *vv
+	default:
+		panic(fmt.Errorf("unhandled type: %T", vv))
```
silaslenihan: why panic not return err?

pkg/solana/codecv2/solana.go
```diff
@@ -0,0 +1,375 @@
...
+			named[idx] = commonencodings.NamedTypeCodec{Name: cases.Title(language.English, cases.NoLower).String(fieldName), Codec: typedCodec}
+		}
+		structCodec, err := commonencodings.NewStructCodec(named)
+		if err != nil {
+			return name, nil, err
+		}
+
+		return name, structCodec, nil
+	case anchoridl.IdlDefinedFieldsTuple:
+		panic(fmt.Errorf("unhandled type: %T", fields))
```
silaslenihan: again, why panic

pkg/solana/codecv2/solana.go
```diff
@@ -0,0 +1,375 @@
...
+		fieldName := field.Name
+		typedCodec, err := processFieldType(ixName, field.Ty, refs)
+		if err != nil {
+			return nil, err
+		}
+		named[idx] = commonencodings.NamedTypeCodec{Name: cases.Title(language.English, cases.NoLower).String(fieldName), Codec: typedCodec}
+	}
+
+	var isVecOrArray bool
+	switch fields[0].Ty.(type) {
```
silaslenihan: could this be empty and cause panic?

pkg/solana/codecv2/solana.go
```diff
@@ -0,0 +1,375 @@
...
+		structCodec, err := commonencodings.NewStructCodec(named)
+		if err != nil {
+			return name, nil, err
+		}
+
+		return name, structCodec, nil
+	case anchoridl.IdlDefinedFieldsTuple:
+		panic(fmt.Errorf("unhandled type: %T", fields))
+	default:
+		return name, nil, nil
```
silaslenihan: should return err here probably

pkg/solana/codecv2/solana_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"testing"
+
+	anchoridl "github.com/gagliardetto/anchor-go/idl"
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-solana/pkg/solana/codecv2"
+	solcommoncodec "github.com/smartcontractkit/chainlink-solana/pkg/solana/commoncodec"
+)
+
+func TestCreateCodecEntry(t *testing.T) {
```
silaslenihan, line 16: maybe comes in later PRs but can we expand these tests to Encode/Decode?

yashnevatia: yup comes later.

silaslenihan (approved):

amit-momin (approved):

PR #1422, opened 2026-01-15
Enable codecv2 (anchor > 0.3) on LogPoller / ChainAccessor

yashnevatia:
- This is the fourth PR in a series of PRs
- First PR (split out commoncodec from codecv1) https://github.com/smartcontractkit/chainlink-solana/pull/1419)
- Second PR (add codec v2)  #1420
- Third PR (add codecv2 to chainwriter)
- This PR enables LogPoller and ChainAccessor to now work with codecv2 which uses anchor > 0.3x idl parsing and definitions
- This is what the final state will look like https://github.com/smartcontractkit/chainlink-solana/pull/1408
- Discussion [Doc](https://docs.google.com/document/d/1Ek6VDZEp_8lwGW9fxV3P6KGvTPCXD9HUligfGy768qI/edit?tab=t.0#heading=h.ww5zhpr4mlw6)


pkg/solana/ccip/chainaccessor/event.go
```diff
@@ -226,6 +210,43 @@ func (a *SolanaAccessor) registerFilterIfNotExists(
 	return nil
 }
 
+// CreateEventIdlWrapper creates an EventIdlWrapper from an IDL string and event name.
+// It tries to unmarshal as codecv2 (anchor-go) IDL first, then falls back to codec IDL.
+func CreateEventIdlWrapper(eventName, idlString string) (logpollertypes.EventIdlWrapper, error) {
```
dhaidashenko: We should try to hide from users the fact that we support two versions of IDL. All the clients (chain writer, logpoller) do not care what's inside of the IDL definition they only need it to create encoder/decoders.

pkg/solana/ccip/chainaccessor/event.go
```diff
@@ -226,6 +210,43 @@ func (a *SolanaAccessor) registerFilterIfNotExists(
...
+	var innerEventIDL logpollertypes.EventIdl
+	var wrapper logpollertypes.EventIdlWrapper
+
+	// Try to unmarshal into codecv2 (anchor-go) IDL first
+	// The full IDL structures are different enough that JSON unmarshal actually fails for the wrong type
+	var codecv2IDL anchoridl.Idl
+	if err := json.Unmarshal([]byte(idlString), &codecv2IDL); err == nil {
```
dhaidashenko, line 222: It's possible to unmarshal v1 into v2 struct and vice versa without any errors, but we'll lose data. We should make a decision on which version to use based on the provided version in the IDL

pkg/solana/logpoller/types/types.go
```diff
@@ -136,20 +137,116 @@ type Decoder interface {
...
+type EventIdl interface {
+	Equal(other EventIdl) bool
+	_isEventIdl() // private method - seals the interface to this package only
+}
 
-func (e *EventIdl) Scan(src interface{}) error {
-	return scanJSON("EventIdl", e, src)
+// EventIdlWrapper ican hold either CodecEventIdl or Codecv2EventIdl
+// It tries both types and uses whichever one works
+type EventIdlWrapper struct {
```
dhaidashenko, line 149: Does LogPoller need to know if the filter was defined using v1 or v2 IDL? Can we resolve it during the creation of the decoder? Also decoder constructor should be inside the codec pkg

pkg/solana/ccip/chainaccessor/event.go
yashnevatia: The version in the IDL does not reflect the anchor version.
I understand what you mean about json unmarshall being a bit lenient.
If you look at the tests, they would not pass if we could unmarshal one into the other.
Specifically for the event type, the idl is very different between v1 and v2, so if you prefer I can use that difference.

pkg/solana/ccip/chainaccessor/event.go
yashnevatia: Thats what we are doing in terms of functionality.
You can pass the component any string, and it will figure out what version of codec to use.
Are you talking about making the method itself internal to the accessor ?

yashnevatia: Made it a local func.
[Commit](https://github.com/smartcontractkit/chainlink-solana/pull/1422/commits/6247ab10f4b4b9a965bfeb246a1537c5b9153302)

pkg/solana/logpoller/types/types.go
yashnevatia: I had tried to make it agnostic but had run into issues which I do not clearly remember now.
I will try it again by passing in the idl as a string and see what happens.

PR #1427, opened 2026-01-21
PLEX-1920 fix SolanaService.SubmitTransaction 

Unheilbar:
supports:
- https://github.com/smartcontractkit/capabilities/pull/414
-  E2E: https://github.com/smartcontractkit/chainlink/pull/20856
Contracts are carried out in a separate go module in order to improve artifacts resolution logic in core


pkg/solana/solana_service.go
```diff
@@ -270,6 +271,8 @@ func (ss *solanaService) SubmitTransaction(ctx context.Context, req commonsol.Su
 	if err != nil {
 		return nil, fmt.Errorf("invalid transaction payload: %w", err)
 	}
+	// remove dummy signatures
```
silaslenihan: nit: could you maybe just add a little more on where these dummy signatures come from?

pkg/solana/solana_service.go
```diff
@@ -723,36 +724,68 @@ func convertDataBytesOrJSON(obj *rpc.DataBytesOrJSON, pref commonsol.EncodingTyp
...
 	}
 	if pref == "" {
-		pref = commonsol.EncodingJSON // default
+		pref = commonsol.EncodingBase64
 	}
 
 	txBytes := obj.GetBinary()
 
-	txJSON, jerr := json.Marshal(obj)
+	txJSON, _ := json.Marshal(obj)
```
silaslenihan: why ignore error?

silaslenihan (commented):
Assuming we are killing write target because products will jump directly to mainline CRE

contracts/programs/keystone-forwarder/src/events.rs
```diff
@@ -8,6 +8,12 @@ pub struct ReportProcessed {
     pub result: bool,
 }
 
+#[event]
+pub struct ReportInProgress {
```
amit-momin, line 12: Didn't know a better place to put this comment but I assume we probably have to regenerate the gobinding too right? https://github.com/smartcontractkit/chainlink-solana/tree/develop/contracts/generated/keystone_forwarder

pkg/solana/chainwriter/chain_writer_test.go
```diff
@@ -876,16 +875,13 @@ func TestChainWriter_SubmitTransaction(t *testing.T) {
...
 		require.NoError(t, err)
-		buffer_payload.ProgramID = bufferProgramID
 
 		t.Run("submits as single transaction if tx small enough", func(t *testing.T) {
 			args := BufferArgs{
 				Report: make([]byte, 963),
 				Fail:   false,
 			}
-
-			ix, err := buffer_payload.NewExecuteInstruction(args.Report, args.Fail, admin, solana.SystemProgramID).ValidateAndBuild()
```
amit-momin, line 887: Was there an issue with the gobinding that required this change?

pkg/solana/solana_service.go
```diff
@@ -723,36 +724,71 @@ func convertDataBytesOrJSON(obj *rpc.DataBytesOrJSON, pref commonsol.EncodingTyp
...
+		if err := json.Unmarshal(txJSON, &arr); err != nil {
+			return nil, fmt.Errorf("expected base64 bytes but GetBinary() empty; also failed to parse json: %w json=%s", err, string(txJSON))
+		}
+		if len(arr) != 2 {
+			return nil, fmt.Errorf("expected [data,encoding] json array, got len=%d json=%s", len(arr), string(txJSON))
+		}
 
-			enc = commonsol.EncodingBase64 // fallback
+		s, _ := arr[0].(string)
+		enc, _ := arr[1].(string)
```
amit-momin: For sanity, should we just make that these really are strings by checking the bool?

pkg/solana/chainwriter/chain_writer_test.go
Unheilbar: Yes, I carried out /contracts into a separate go.module, and I ran into circular dependency because of this usage.

contracts/programs/keystone-forwarder/src/events.rs
Unheilbar: I think I did, seems like this change doesnt affect them?

amit-momin (dismissed):

pkg/solana/write_target/write_target.go
```diff
@@ -1,149 +0,0 @@
...
-)
-
-type Chain interface {
-	LatestHead(ctx context.Context) (commontypes.Head, error)
-	//MultiClient() *client.MultiClient
-	ID() string
-	Config() config.Config
-}
-
-func New(chain Chain, reader client.Reader, txm Txm, chainInfo commontypes.ChainInfo, lggr logger.Logger) (capabilities.ExecutableCapability, error) {
```
dhaidashenko, line 32: Are we sure that WriteTarget is not used?

pkg/solana/chain.go
```diff
@@ -614,6 +614,10 @@ func (c *chain) Start(ctx context.Context) error {
 			c.lggr.Debug("Starting multinode")
 			startAll = append(startAll, c.multiNode, c.txSender)
 		}
+		if c.cfg.Workflow.IsEnabled() {
+			c.lggr.Debug("Starting LogPoller")
+			startAll = append(startAll, c.lp)
```
dhaidashenko, line 619: Hm, I wonder how CCIP works without starting the LP

pkg/solana/solana_service.go
```diff
@@ -322,12 +328,7 @@ func (ss *solanaService) SubmitTransaction(ctx context.Context, req commonsol.Su
 		return &commonsol.SubmitTransactionReply{Status: txStatus, IdempotencyKey: transactionID}, nil
 	}
 
-	signature, err := ss.chain.TxManager().GetTransactionSig(transactionID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get transaction signature: %w", err)
-	}
-
-	return &commonsol.SubmitTransactionReply{Status: txStatus, Signature: commonsol.Signature(signature), IdempotencyKey: transactionID}, nil
+	return &commonsol.SubmitTransactionReply{Status: txStatus, IdempotencyKey: transactionID}, nil
```
dhaidashenko, line 331: Isn't it a breaking change that we no longer return the transaction signature? 
Are we sure that noone else uses this method?

pkg/solana/solana_service.go
```diff
@@ -723,36 +724,77 @@ func convertDataBytesOrJSON(obj *rpc.DataBytesOrJSON, pref commonsol.EncodingTyp
...
+		if len(txBytes) != 0 {
+			return &commonsol.DataBytesOrJSON{
+				RawDataEncoding: commonsol.EncodingBase64,
+				AsDecodedBinary: txBytes,
+				AsJSON:          txJSON,
+			}, nil
+		}
+
+		// Fallback: decode ["<base64>", "base64"] manually
+		var arr []any
```
dhaidashenko: nit: why not unmarshal directly to []string?

pkg/solana/write_target/write_target.go
Unheilbar: Yes

pkg/solana/chain.go
Unheilbar: They spin up logPoller during ChainReader creation

pkg/solana/solana_service.go
Unheilbar: 1.  The only consumer of SolanaService currently is solana capability, so noone uses this method
2. TxManager().GetTransactionSig returns only signatures for pending transactions, here call always returns an error since at this point transaction is finalized
3. Solana capability doesn't rely on signature from txm and instead retreives signatures from Logs
However, having signature in reply is potentially missleading so I would either alter GetTransactionSig to support returning signatures for finalized tx (with some cleanup interval) or removed Signature from response. Gonna do this in a follow up PR

dhaidashenko (approved):

dhaidashenko (approved):

PR #1441, opened 2026-02-04
Added GetFiltersNames to SolanaService

silaslenihan:
### Requires
- https://github.com/smartcontractkit/chainlink-common/pull/1801
### Supports
- https://github.com/smartcontractkit/capabilities/pull/337
- https://github.com/smartcontractkit/chainlink/pull/20856


pkg/solana/logpoller/filters.go
```diff
@@ -451,6 +451,22 @@ func (fl *filters) GetFiltersToBackfill() []types.Filter {
 	return result
 }
 
+// GetFilters returns a copy of all currently registered filters, keyed by filter name.
+// Returns an empty map if filters have not been loaded yet.
+func (fl *filters) GetFilters() map[string]types.Filter {
+	if !fl.loadedFilters.Load() {
+		return make(map[string]types.Filter)
```
dhaidashenko: It's not safe to return an empty map if filters are not loaded, as it could lead the caller of GetFilters to act on an empty map, assuming the DB is empty. However, the map is empty because no one called load. 
We should load filters and return an error if we fail

yashnevatia (approved):

yashnevatia (dismissed):
CI Failed and I pressed approved by mistake.
Requesting changes to remove my approval.
Can approve once fixed CI.

Unheilbar (approved):

PR #1465, opened 2026-02-24
Fix LogPoller reprocessing entire block range on partial failure

Krish-vemula:
### Description

Fixes excessive getBlock RPC calls caused by the LogPoller reprocessing already-succeeded blocks when a failure occurs partway through a block range. Previously, lastProcessedSlot was only updated after the entire range completed successfully any error (DB timeout, decoding failure, rate limiting) would cause the next tick to re-fetch and re-process every block from the start of the range. Now lastProcessedSlot advances incrementally after each successfully processed batch, so retries resume from where processing left off.


pkg/solana/logpoller/log_poller.go
```diff
@@ -434,6 +434,13 @@ consumedAllBlocks:
 			if err != nil {
 				return fmt.Errorf("error processing blocks: %w", err)
 			}
+
+			// nolint:gosec
+			// G115: integer overflow conversion uint64 -> int64
+			highestInBatch := int64(batch[len(batch)-1].SlotNumber)
```
amit-momin, line 447: Is there a similar risk doing the update here if the batch has some issues/slowness then the `lastProcessedSlot` update doesn't happen quick enough before a new batch is triggered? Should we go one level deeper and update this after each block process in `processBlocksImpl`?

pkg/solana/logpoller/log_poller.go
amit-momin, line 523: With the new `lastProcessedSlot` update logic, should we remove this from here?

pkg/solana/logpoller/log_poller.go
Krish-vemula: No timing/concurrency risk here i believe - `GoTick` runs `run()` synchronously, so the next tick won't fire until the current one returns. Batches are consumed sequentially in the same goroutine.

Block-level granularity inside `processBlocksImpl` would save slightly more progress (e.g. if block 14 of [11-15] fails, cursor would be at 13 instead of pre-batch). But I'd keep it at batch-level because:

1. `processBlocks` is a swappable function pointer (`NewWithCustomProcessor`) - putting cursor updates inside `processBlocksImpl` wouldn't benefit custom processors
2. Batch size is capped at 16 (`blocksChBuffer`), so worst case we re-process 15 blocks vs the original problem of re-processing entire ranges of thousands

pkg/solana/logpoller/log_poller.go
michelemin: +1

amit-momin (approved):

PR #1475, opened 2026-03-04
return slot instead of blockHeight on LatestBlock

Unheilbar:
[JIRA](https://smartcontract-it.atlassian.net/browse/PLEX-2553)
Report head tracker [progression](https://github.com/smartcontractkit/chainlink-framework/blob/e8678584594270cce2caef08d0669e2fe0dbbe55/multinode/node_lifecycle.go#L312) on blocks instead of slots.

### DIFF
--- a/pkg/solana/client/multinode_client.go
+++ b/pkg/solana/client/multinode_client.go
@@ -19,8 +19,8 @@ import (
 )
 
 type Head struct {
-	BlockHeight *uint64
-	BlockHash   *solana.Hash
+	SlotNumber *uint64
+	BlockHash  *solana.Hash
 }
 
 func (h *Head) BlockNumber() int64 {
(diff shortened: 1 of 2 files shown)


pkg/solana/client/multinode_client.go
```diff
@@ -19,24 +19,24 @@ import (
 )
 
 type Head struct {
-	BlockHeight *uint64
-	BlockHash   *solana.Hash
+	SlotNumber *uint64
+	BlockHash  *solana.Hash
 }
 
 func (h *Head) BlockNumber() int64 {
```
amit-momin, line 26: Would this naming be confusing now since we return the slot number instead? Or is it required to keep this name because of our interfaces?

Unheilbar: Discrepancy comes from chainlink-framework. It expects interface to have LatestBlock, so we have to return LatestSlot as LatestBlock here

amit-momin: Gotcha, was worried about that but guess there's nothing we can do for now

amit-momin (approved):

PR #1482, opened 2026-03-12
LogPoller: process blocks in batches

dhaidashenko:
## Motivation
During prolonged downtime or RPC issues, Chainlink Node may lag behind Solana's latest finalized block. Max lag could be up to 172800 blocks.
LogPoller searches for blocks that contain events defined by Filters and schedules their fetching. 
Block fetching is done by async workers. If the operation fails, the job is retried after a delay; it competes with the rest of the pool for a worker.
Fetched blocks are added to another queue, which sorts them and passes them to the events processor. It's guaranteed that the event processor will observe all blocks with subscribed events, in increasing order.
Thus, if we are scheduled to fetch 1k blocks and we fail to fetch the first block for a long time, all 999 blocks will sit in memory.

Our current assumption is that unreliable RPCs and large lag cause OOM for CL Nodes.
As we do not have access to the profile files for LOOP plugins (WIP), there is no guarantee that our assumption is correct.

 ## Changes
* Process slots in batches. Batch size is defined in config. This way, we can limit the maximum number of blocks loaded into the memory.
* Add a metric to monitor the latest slot processed by LogPoller. Low processing rate or lag accumulation between the latest processed and the latest finalized slots can be used for early alerting on LP health. 

core ref: 78e0a3fb76940b1aa5d700545e29925d5ad90610


pkg/solana/config/toml.go
```diff
@@ -238,6 +238,10 @@ func (c *Chain) SetFrom(f *Chain) {
 	if f.LogPollerCPIEventsEnabled != nil {
 		c.LogPollerCPIEventsEnabled = f.LogPollerCPIEventsEnabled
 	}
+
+	if f.LogPollerSlotsBatchSize != nil {
+		c.LogPollerSlotsBatchSize = f.LogPollerSlotsBatchSize
```
silaslenihan, line 243: probably need to also validate it's > 0

amit-momin: +1

amit-momin (dismissed):

jmank88 (dismissed):

archseer (dismissed):

jadepark-dev (dismissed):

jmank88 (approved):

amit-momin (approved):

jadepark-dev (approved):

archseer (approved):

ogtownsend (approved):

PR #1499, opened 2026-04-06
add unit test for solana service

Unheilbar:


pkg/solana/solana_service_test.go
```diff
@@ -340,6 +350,434 @@ func csig(i byte) commonsol.Signature { return commonsol.Signature(sig(i)) }
...
+	req := commonsol.SubmitTransactionRequest{
+		EncodedTransaction: validBase64Tx(t),
+		Receiver:           cpk(1),
+		Cfg: &commonsol.ComputeConfig{
+			ComputeLimit:    &computeLimit,
+			ComputeMaxPrice: &maxPrice,
+		},
+	}
+
+	reply, err := ss.SubmitTransaction(context.Background(), req)
```
amit-momin: nit: 
```suggestion
	reply, err := ss.SubmitTransaction(t.Context(), req)
```

amit-momin: Noticed you're using `context.Background()` in some other spots in tests too

pkg/solana/solana_service.go
```diff
@@ -369,13 +375,16 @@ func (ss *solanaService) SubmitTransaction(ctx context.Context, req commonsol.Su
 		return nil, fmt.Errorf("failed to enqueue transaction: %w", err)
 	}
 
-	maximumWaitTimeForConfirmation := ss.chain.Config().WF().AcceptanceTimeout()
+	maximumWaitTimeForConfirmation := max(ss.chain.Config().WF().AcceptanceTimeout(), minimumConfirmationTime)
+
 	retryContext, cancel := context.WithTimeout(ctx, maximumWaitTimeForConfirmation)
 	defer cancel()
 
+	var lastStatusErr error
```
amit-momin: I was looking into `retry.Do` and doesn't that already track the `lastKnownError` and return it? Also wouldn't the `lastStatusErr` and `err` essentially be the same thing?

Unheilbar: Oh, actually it didn't used to.. I remember I had to do the same tweak in evm_service because of that. But now I see it was updated to aactually return last known error. Pushing a fix

pkg/solana/solana_service_test.go
Unheilbar: fixed

amit-momin (approved):

PR #1504, opened 2026-04-07
L1 PLEX-2601 LP Fetch Blocks in Batches

dhaidashenko:
## Summary
PLEX-2601 — Batch block fetching inside the log poller loader so slot discovery via `getSignaturesForAddress` runs once per backfill range, instead of restarting for every slot-range batch.

## Problem
Previously, large backfills were split in the log poller into multiple independent slot ranges. Each range called BackfillForAddresses, which re-ran `getSlotsToFetch` from the newest signatures. Pagination (`beforeSig`) did not carry across adjacent batches, so every batch effectively rediscovered recent history and walked backward again. That added redundant RPC load, slowed deep catch-up on hot addresses, and could worsen throttling or recovery time in worst cases. 

## Solution
* Revert batching at the poller layer (`processBlocksRangeInBatches` / per-batch BackfillForAddresses).
* Move batching to EncodedLogCollector: after a single getSlotsToFetch pass for [fromSlot, toSlot], scheduleBlocksFetching processes the resulting slot list in chunks of LogPollerSlotsBatchSize (limits concurrent getBlock work / memory), while signature discovery pagination stays coherent for the whole range.
* Lower the default LogPollerSlotsBatchSize to further limit memory usage, as `getSlotsToFetch` no longer executed multiple times.

### DIFF
--- a/pkg/solana/logpoller/loader.go
+++ b/pkg/solana/logpoller/loader.go
@@ -120,16 +120,12 @@ func (c *EncodedLogCollector) scheduleBlocksFetching(ctx context.Context, slots 
 	}
 
 	c.engine.GoCtx(ctx, func(ctx context.Context) {
-		ctx, cancel := context.WithCancel(ctx)
 		startTime := time.Now()
-		defer cancel()
 		for batchStart := 0; batchStart < len(slots); {
 			batchEnd := min(batchStart+c.slotsBatchSize, len(slots))
 			jobs, err := c.scheduleBlocks(ctx, slots[batchStart:batchEnd], blocks)
 			if err != nil {
 				c.lggr.Errorw("Failed to schedule block fetching jobs for batch.", "err", err)
-				// do not close blocks channel here to avoid signalling end of stream to the consumer - instead, just stop all the jobs and return
-				cancel()
 				return
 			}
 
@@ -145,6 +141,7 @@ func (c *EncodedLogCollector) scheduleBlocksFetching(ctx context.Context, slots 
 		}
 
 		c.lggr.Infow("Finished fetching all blocks.", "from", slots[0], "to", slots[len(slots)-1], "duration", time.Since(startTime))
+		// only close the channel after all jobs are done. Do not close on failures to avoid signaling successful completion to the upstream
 		close(blocks)
 	})
 	return blocks, nil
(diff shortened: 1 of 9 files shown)


pkg/solana/logpoller/loader.go
```diff
@@ -110,31 +114,58 @@ func (c *EncodedLogCollector) getSlotsToFetch(ctx context.Context, addresses []t
...
+		ctx, cancel := context.WithCancel(ctx)
+		startTime := time.Now()
+		defer cancel()
+		for batchStart := 0; batchStart < len(slots); {
+			batchEnd := min(batchStart+c.slotsBatchSize, len(slots))
+			jobs, err := c.scheduleBlocks(ctx, slots[batchStart:batchEnd], blocks)
+			if err != nil {
+				c.lggr.Errorw("Failed to schedule block fetching jobs for batch.", "err", err)
+				// do not close blocks channel here to avoid signalling end of stream to the consumer - instead, just stop all the jobs and return
+				cancel()
```
jmank88: Can omit this since it is also deferred:
```suggestion
```
Although then is creating a new context even necessary? or can 123/125 be dropped too?

archseer (dismissed):

pkg/solana/logpoller/loader_internal_test.go
```diff
@@ -0,0 +1,86 @@
...
+	expectedSlots := make(map[uint64]struct{})
+	for _, slot := range slots {
+		expectedSlots[slot] = struct{}{}
+	}
+
+	prev := uint64(0)
+	for block := range ch {
+		if prev != 0 {
+			require.Equal(t, uint64(1), block.SlotNumber-prev, "expected blocks to be received in consecutive order since batch size is 1")
+		}
```
amit-momin, line 81: I don't think we're setting prev in this loop so this check isn't happening

pkg/solana/logpoller/loader_internal_test.go
```diff
@@ -0,0 +1,86 @@
...
+		require.Contains(t, expectedSlots, block.SlotNumber, "received block for unexpected slot")
+		delete(expectedSlots, block.SlotNumber)
+	}
+	require.Empty(t, expectedSlots)
+}
```
amit-momin, line 87: Could we also write a test that confirms that we aren't refetching the same slots when doing the get slots for addr job? Just want to make sure the reason we're making this change is covered in a test

amit-momin (approved):

jadepark-dev (approved):
lgtm

ogtownsend (approved):

PR #1512, opened 2026-04-08
PLEX-2601 LP fix filters inmemory slot tracking

dhaidashenko:
When filter backfill failed partway through a range, the next poll run restarted backfill from each filter’s original StartingBlock up to the current head. That re-fetched and re-processed slots that had already been handled successfully before the error.

This change tracks per-filter backfill progress in memory, persists partial progress via UpdateBackfillProgress, and uses GetFiltersToBackfill(lastProcessedSlot) to compute the minimum slot from which the next backfill should run. After a partial failure, the poller resumes from the last checkpointed slot instead of replaying the full historical range.

### DIFF
--- a/pkg/solana/logpoller/log_poller.go
+++ b/pkg/solana/logpoller/log_poller.go
@@ -116,7 +116,7 @@ func New(lggr logger.SugaredLogger, orm ORM, cl RPCClient, cfg config.Config, ch
 		Start: lp.start,
 		NewSubServices: func(lggr logger.Logger) []services.Service {
 			lp.filters = newFilters(lggr, orm, lp.cpiEventExtractor)
-			loader := NewEncodedLogCollector(cl, lggr, chainID, lp.metrics, lp.cpiEventExtractor)
+			loader := NewEncodedLogCollector(cl, lggr, chainID, lp.metrics, lp.cpiEventExtractor, cfg.LogPollerSlotsBatchSize())
 			lp.loader = loader
 			return []services.Service{loader}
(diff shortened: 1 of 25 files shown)


jadepark-dev (dismissed):

pkg/solana/logpoller/log_poller.go
```diff
@@ -373,50 +373,74 @@ func (lp *Service) checkForReplayRequest() bool {
...
+	prevCheckpoint := minSlot
+	var markedCompleted bool
+	saveBackfillProgress := func(ctx context.Context, lastProcessed int64) error {
+		if markedCompleted {
+			return nil
+		}
+
+		isComplete := lastProcessed >= to
+		// avoid updating progress too frequently to reduce db load, but still update at least every 100 blocks or once backfill is complete
+		const saveCheckpointInterval = 100
```
amit-momin: I thought the progress update was only in-memory and we only fall down to updating the DB once the backfill is complete ([here](https://github.com/smartcontractkit/chainlink-solana/pull/1512/changes#diff-43599bcdcf8c088e3eac4f93fd79202871475fcdeb2dcc9494a8935b9f991a07R513-R516)). Does this actually reduce load on the DB?

dhaidashenko: You are right. This can be simplified.

amit-momin (dismissed):

ogtownsend (dismissed):

amit-momin (approved):

PR #1517, opened 2026-04-08
Add LogPoller metric to track number of skipped blocks

amit-momin:


pkg/solana/logpoller/metrics.go
```diff
@@ -29,6 +29,11 @@ var promLpLastProcessedSlot = promauto.NewGaugeVec(prometheus.GaugeOpts{
 	Help: "Last processed slot by log poller",
 }, []string{"chainID"})
 
+var promLpBlockSkipped = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
dhaidashenko: It's possible that we skip a block during backfill of a filter, so the value won't always increase. Also due to delay in metrics reporting, it's possible that some values will be skipped.
NOPs should use logs to find all skipped blocks. 
Won't it be more reliable to count the number of skipped blocks? This way, we can create an alert if the value increases, so that NOP can check their RPC.

amit-momin: Even the block fetch aborts may not occur in ascending order of slots so don't think this metric graphed would look like an always increasing line. But someone debugging skipped blocks could take the min of this metric over a range of time to know where the earliest skip occurred. Tbh I prefer the counter you're suggesting but I thought the motive of the metric was to make it easier to figure out what the skipped block is so we don't have to go into the logs. Maybe a counter and your latest processed block metric could be combined to essentially do the same. We just wouldn't have the exact block to start any backfills from

Unheilbar (dismissed):

pkg/solana/logpoller/metrics.go
jadepark-dev: +1 for counter. imo Individual slot visibility isn't really the metric's job anyway. Counter would make alerting on failure rate actually useful

amit-momin: Updated the metric in the latest commit

ogtownsend (approved):

PR #1523, opened 2026-04-13
fix snake to pascal conversion

Unheilbar:

### DIFF
--- a/pkg/solana/codec/v2/solana_test.go
+++ b/pkg/solana/codec/v2/solana_test.go
@@ -1,4 +1,4 @@
-package codecv2_test
+package codecv2
 
 import (
 	_ "embed"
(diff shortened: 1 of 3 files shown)


pkg/solana/codec/v2/solana_test.go
```diff
@@ -94,28 +95,148 @@ func TestExtractEventIDL(t *testing.T) {
...
+			var results []string
+			for _, input := range group.inputs {
+				result, err := snakeToPascal(input)
+				if err != nil {
+					continue
+				}
+				results = append(results, result)
+			}
+			if len(results) > 1 {
+				t.Logf("known collision risk: inputs %v all produce the same output %q", group.inputs, results[0])
```
amit-momin: Think we probably want to make some assertion here right?

amit-momin (approved):

PR #1524, opened 2026-04-13
wrap codec init err wrap for v1/2

Unheilbar:

### DIFF
--- a/pkg/solana/codec/codec.go
+++ b/pkg/solana/codec/codec.go
@@ -18,14 +18,14 @@ import (
 func NewEventCodecEntry(filter types.Filter, includeDiscriminator bool, mod commoncodec.Modifier, builder encodings.Builder) (solcommoncodec.Entry, error) {
 	if filter.ContractIdl != "" {
 		// Try codecv2 first (Anchor 0.30+ IDL format)
-		entry, err := codecv2.NewEventArgsEntryWrapper(filter.EventName, filter.ContractIdl, includeDiscriminator, mod, builder)
-		if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/solana/codec/codec.go
```diff
@@ -18,14 +18,14 @@ import (
 func NewEventCodecEntry(filter types.Filter, includeDiscriminator bool, mod commoncodec.Modifier, builder encodings.Builder) (solcommoncodec.Entry, error) {
 	if filter.ContractIdl != "" {
 		// Try codecv2 first (Anchor 0.30+ IDL format)
-		entry, err := codecv2.NewEventArgsEntryWrapper(filter.EventName, filter.ContractIdl, includeDiscriminator, mod, builder)
-		if err == nil {
+		entry, v2Err := codecv2.NewEventArgsEntryWrapper(filter.EventName, filter.ContractIdl, includeDiscriminator, mod, builder)
+		if v2Err == nil {
```
amit-momin, line 22: I doubt there's a case like this but could a user expect their IDL to succeed v2 parsing but it fails and succeeds with v1? If that can happen, wouldn't we want to log v2 failure regardless? Otherwise, we'll only know this failed v2 parsing if v1 also fails.

silaslenihan (approved):

pkg/solana/codec/codec.go
Unheilbar: I think it's a nice to have  and can help debugging cases where user submitted v1 idl by accident, meaning to decode v2 events.. 
However it will require changes a couple more places to pipe logger here, so I tend to do this in follow up

PR #1528, opened 2026-04-14
fix filter equality check

Unheilbar:

### DIFF
--- a/pkg/solana/logpoller/types/models.go
+++ b/pkg/solana/logpoller/types/models.go
@@ -43,12 +43,8 @@ func (c ExtraFilterConfig) Equal(other ExtraFilterConfig) bool {
 
 func (f Filter) MatchSameLogs(other Filter) bool {
 	status := (f.Address == other.Address) && (f.EventSig == other.EventSig) && (f.EventName == other.EventName) && (f.SubkeyPaths.Equal(other.SubkeyPaths))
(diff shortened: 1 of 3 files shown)


pkg/solana/logpoller/types/models.go
```diff
@@ -43,12 +43,8 @@ func (c ExtraFilterConfig) Equal(other ExtraFilterConfig) bool {
...
 	status := (f.Address == other.Address) && (f.EventSig == other.EventSig) && (f.EventName == other.EventName) && (f.SubkeyPaths.Equal(other.SubkeyPaths))
-	if (f.EventIdl.Event.Name != "") && (other.EventIdl.Event.Name != "") {
-		status = status && f.EventIdl.Equal(other.EventIdl)
-	}
-	if f.ContractIdl != "" && other.ContractIdl != "" {
-		status = status && f.ContractIdl == other.ContractIdl
-	}
+	status = status && f.EventIdl.Equal(other.EventIdl)
+	status = status && f.ContractIdl == other.ContractIdl
 	return status
```
amit-momin, line 48: Sort of related to the fix but I noticed the comment on [RegisterFilter](https://github.com/smartcontractkit/chainlink-solana/blob/9e8a54e33a3783ebd5a9e4e86b213c52fb6c162e/pkg/solana/logpoller/filters.go#L151) doesn't list all of the fields that we try to match on. Could you update it to include EventName and ContractIDL?

amit-momin (approved):

PR #1529, opened 2026-04-14
intoduce resolveSourceForLog helper

Unheilbar:

### DIFF
--- a/pkg/solana/logpoller/cpi_event_extractor.go
+++ b/pkg/solana/logpoller/cpi_event_extractor.go
@@ -140,7 +140,7 @@ func (e *CPIEventExtractor) ExtractCPIEvents(
 			if methodSig == types.AnchorCPIEventDiscriminator() {
 				eventData, ok = extractAnchorCPIEventData(e.lggr, ix.Data)
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/solana/logpoller/cpi_event_extractor.go
```diff
@@ -268,6 +270,13 @@ func extractVecCPIEventData(
 	return data[CPIEventDataOffsetLegacy:], true
 }
 
+func resolveSourceForLog(stackHeight uint16, programAtStackHeight map[uint16]types.PublicKey) string {
+	if stackHeight == 0 {
+		return "unknown (StackHeight=0)"
```
amit-momin: In a method above, it seems like we fall back to the outer instruction's program when stack height is 0. Should we do the same here for the warning log? Silas might be a better person to confirm this though since he helped write this logic. https://github.com/smartcontractkit/chainlink-solana/blob/9e8a54e33a3783ebd5a9e4e86b213c52fb6c162e/pkg/solana/logpoller/cpi_event_extractor.go#L150

silaslenihan: Yeah I think that makes sense

amit-momin (approved):

jadepark-dev (approved):

PR #1530, opened 2026-04-16
bump anchor to 0.31

Unheilbar:
core ref: 4df126be4d3cb13370633d35e52b215c8ae48909


ilija42 (approved):

Fletch153 (approved):

contracts/examples/hello-world/tests/hello-world.ts
```diff
@@ -3,7 +3,8 @@ import * as fs from "fs";
 import { Program, BN } from "@coral-xyz/anchor";
 import { HelloWorld } from "../target/types/hello_world";
 
-const CHAINLINK_PROGRAM_ID = "HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny";
+// Must match `store` in `contracts/Anchor.toml` and `programs/store` `declare_id!`.
+const CHAINLINK_PROGRAM_ID = "9kRNTZmoZSiTBuXC62dzK9E7gC7huYgcmRRhYv3i4osC";
```
augustbleeds, line 7: I think scripts/lib.sh will need to be updated. but like lots of the files , it probably isn't used that often

contracts/programs/ocr_2/src/lib.rs
```diff
@@ -1,10 +1,11 @@
+#![allow(deprecated)] // anchor-lang 0.31 #[program] macro uses AccountInfo::realloc
 use anchor_lang::prelude::*;
 use anchor_spl::token;
 
 use arrayref::{array_ref, array_refs};
 use state::{Billing, Proposal, ProposedOracle, STATE_VERSION};
 
-declare_id!("cjg3oHmg9uuPsP8D6g29NWvhySJkdYdAo9D25PRbKXJ");
```
augustbleeds, line 7: this is also referenced in scripts/lib.sh

sdk/chainlink-solana/src/lib.rs
```diff
@@ -164,26 +169,32 @@ pub mod v2 {
...
-    /// Ensure these values come from the same feed account.
+    /// Ensure feed data and owner come from the same feed account.
     ///
     // DEV: Method does not expose `AccountInfo` to avoid
     // dependency from `anchor-lang` or `solana-program` version
     pub fn read_feed_v2(
         data: Ref<&mut [u8]>,
-        owner: [u8; 32],
+        feed_owner: [u8; 32],
+        store_program_id: &Pubkey,
```
augustbleeds, line 191: Ah yes. So this will be a backwards incompatible change, so the sdk should be bumped to a new major version. I remember that the store program is the same on testnet and mainnet so we kept it hardcoded so the client doesn't need to pass in their instruction args. We just need to let users know a new arg is required.

augustbleeds, line 191: I think JupLend is still the only user for Solana DF?

PR #1543, opened 2026-04-27
remove fromAddress dependency from workflow config

Unheilbar:
core ref: d59bd2dd997f032b2180f9096bc43733c55e023f


pkg/solana/config/testdata/config-full.toml
```diff
@@ -29,9 +29,6 @@ LogPollerSlotsBatchSize = 1000
 
 [Workflow]
 AcceptanceTimeout = '42s'
-ForwarderAddress = '14grJpemFaf88c8tiVb77W7TYg2W3ir6pfkKz3YjhhZ5'
```
dhaidashenko, line 32: I assume that removal of this fields is ok, since we've added them and we do not expect anyone else to use them in prod

dhaidashenko (approved):

amit-momin (approved):

chris-de-leon-cll (approved):

ilija42 (approved):

PR #1555, opened 2026-05-11
Narrow fields returned scope of BHE's GetBlock fetch

ogtownsend:
Narrow BHE `getBlock` payload to cut allocation churn

### DIFF
--- a/pkg/solana/client/client.go
+++ b/pkg/solana/client/client.go
@@ -40,22 +40,6 @@ func HeadMetadataGetBlockOpts(commitment rpc.CommitmentType) *rpc.GetBlockOpts {
 	}
 }
 
-// FeeEstimatorGetBlockOpts returns options for getBlock tuned for the BlockHistoryEstimator.
-// This includes the full transaction list which is needed to inspect compute budget instructions
-// and tx fees, but with rewards excluded and base64 transaction encoding so the response is
-// decoded via the binary path instead of reflective JSON (memory heavy, 1-4MB JSON payload per block).
-func FeeEstimatorGetBlockOpts(commitment rpc.CommitmentType) *rpc.GetBlockOpts {
-	rewards := false
-	v := MaxSupportTransactionVersion
-	return &rpc.GetBlockOpts{
-		Commitment:                     commitment,
-		Encoding:                       solana.EncodingBase64,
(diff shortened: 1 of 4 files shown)


pkg/solana/client/client.go
```diff
@@ -40,6 +40,22 @@ func HeadMetadataGetBlockOpts(commitment rpc.CommitmentType) *rpc.GetBlockOpts {
 	}
 }
 
+// FeeEstimatorGetBlockOpts returns options for getBlock tuned for the BlockHistoryEstimator.
+// This includes the full transaction list which is needed to inspect compute budget instructions
+// and tx fees, but with rewards excluded and base64 transaction encoding so the response is
+// decoded via the binary path instead of reflective JSON (memory heavy, 1-4MB JSON payload per block).
+func FeeEstimatorGetBlockOpts(commitment rpc.CommitmentType) *rpc.GetBlockOpts {
```
amit-momin: Think it would be better to put this in the BHE and just use the existing `GetBlockWithOpts` method already available in the client there.

pkg/solana/client/client.go
```diff
@@ -477,11 +493,11 @@ func (c *Client) GetBlock(ctx context.Context, slot uint64) (result *rpc.GetBloc
...
-		version := MaxSupportTransactionVersion
-		return c.rpc.GetBlockWithOpts(ctx, slot, &rpc.GetBlockOpts{
-			Commitment:                     c.commitment,
-			MaxSupportedTransactionVersion: &version,
-		})
+		// The only production caller of this method (and the GetLatestBlock wrapper above) seems to be the
+		// BlockHistoryEstimator, which only consumes tx.Meta.Fee and tx.Message.Instructions/AccountKeys.
+		// Use the BHE-tuned opts so we don't pull rewards/balances/log messages or pay heavy reflective
+		// JSON decoding cost on every poll. See FeeEstimatorGetBlockOpts for rationale.
+		return c.rpc.GetBlockWithOpts(ctx, slot, FeeEstimatorGetBlockOpts(c.commitment))
```
amit-momin: Think we should revert this even though only the BHE uses it. We already have a `GetBlockWithOpts` method that allows us to pass an opts through

amit-momin (approved):

PR #1563, opened 2026-06-02
PLEX-2773: Support both include reverted states

silaslenihan:
### Requires Dependencies

- https://github.com/smartcontractkit/chainlink-common/pull/2122/

### Resolves Dependencies

- https://github.com/smartcontractkit/capabilities/pull/633


pkg/solana/logpoller/filters.go
```diff
@@ -183,18 +183,6 @@ func (fl *filters) RegisterFilter(ctx context.Context, filter types.Filter) erro
 		fl.removeFilterFromIndexes(*existingFilter)
 	}
 
-	// ensure that the value of IncludeReverted isn't different from any other filters with the same address and event type
-	if contractFilters, okAddr := fl.filtersByAddress[filter.Address]; okAddr {
```
dhaidashenko, line 203: Until CCIP starts passing the filter name we should keep this logic, but hide it behind feature flag that is enabled by default and disabled for CRE

pkg/solana/logpoller/orm.go
```diff
@@ -245,8 +246,20 @@ func (o *DSORM) SelectLogs(ctx context.Context, start, end int64, address types.
 	return logs, nil
 }
 
-func (o *DSORM) FilteredLogs(ctx context.Context, filter []query.Expression, limitAndSort query.LimitAndSort, _ string) ([]types.Log, error) {
-	qs, args, err := (&pgDSLParser{}).buildQuery(o.chainID, filter, limitAndSort)
+// FilteredLogs runs a DSL query against stored logs. When queryName is set, results are
+// scoped to that registered filter (filter_id in SQL, no dedup). An empty queryName is the
+// legacy path: query all filters for the chain and dedupe by (block_number, log_index).
+func (o *DSORM) FilteredLogs(ctx context.Context, filter []query.Expression, limitAndSort query.LimitAndSort, queryName string) ([]types.Log, error) {
```
dhaidashenko, line 252: This is a breaking change that occurs at runtime, leading to hard-to-track bugs. 
You are re-proposing a previously existing function argument. Previously it was queryName, and now it's filterName.
If someone imported DSORM and passed a non-empty queryName, now it'll be treated as filterName

pkg/solana/logpoller/parser.go
```diff
@@ -41,6 +42,9 @@ var (
 	logsFields = [...]string{"chain_id", "log_index", "block_hash", "block_number", "block_timestamp", "address",
 		"event_sig", "tx_hash", "data", "error"}
 
+	scopedLogsFields = [...]string{"filter_id", "chain_id", "log_index", "block_hash", "block_number", "block_timestamp", "address",
```
dhaidashenko, line 45: Do we use filter_id from the Logs structure after select? I'm worried that the field's presence depends on the filter value.

dhaidashenko (changes requested):

PR #1565, opened 2026-06-05
Use cl-framework node_balance metric for balance tracking

ogtownsend:
- use `RecordNodeBalance()` instead of `metrics.NodeBalance.WithLabelValues` to log `node_balance` to both beholder and prom for solana
- keep `solana_balance` prom metric for now until we release this new metric, confirm it's working, and NOPs can migrate over to using the new metric


pkg/solana/monitor/prom_test.go
```diff
@@ -14,8 +15,14 @@ func TestPromSolBalance(t *testing.T) {
...
+	require.NoError(t, err)
+
+	monitor := balanceMonitor{
+		chainID:        "test-chain",
+		balanceMetrics: balanceMetrics,
+	}
+	monitor.updateProm(t.Context(), key, balance)
 
 	// happy path test
 	promBalance := testutil.ToFloat64(metrics.NodeBalance.WithLabelValues(key.String(), monitor.chainID, metrics.Solana))
```
amit-momin: Can't remember if there's an easy way to do it currently but can we also check with the beholder metric recorded the right value here too? Just so we have coverage on both prom and beholder? If there isn't a way, no worries.

amit-momin (approved):

PR #1572, opened 2026-06-16
Skip LogPoller filter errors

silaslenihan:


Unheilbar (approved):

pkg/solana/solana_service.go
amit-momin: Is there any risk in this gap between unregister and register here?

silaslenihan: There would be. And looking at this further I don't think we need to update this logic. For CRE, this scenario is only applicable if someone redeploys a workflow and changes a filter. But when a filter gets re-deployed, the filters are unregistered anyway, so this logic is redundant. I will revert the logic

amit-momin (approved):

PR #1578, opened 2026-06-26
Deprecate eventDetail for BlockData for CPI event extraction

ogtownsend:
- We need to be able to call `ExtractCPIEvents` from o11y
- `eventDetail` was private and was preventing us from calling this
- Upon looking more, it seems like `eventDetail` and `BlockData` are basically the same


silaslenihan (approved):
LGTM

PR #1611, opened 2026-08-25
Return err on db failure

yashnevatia:
Something to note here:

Earlier, db error was returned as filter does not exist and we would proceed to RegisterFilter
Now, in `pkg/solana/ccip/chainaccessor/event.go` and `pkg/solana/solana_service.go` we return an err instead of proceeding to RegisterFilter.


silaslenihan (dismissed):
Lgtm

pkg/solana/chainreader/filter.go
```diff
@@ -83,7 +83,15 @@ func (r *syncedFilter) Unregister(ctx context.Context, registrar filterRegistrar
 }
 
 func (r *syncedFilter) unregister(ctx context.Context, registrar filterRegistrar, name string) error {
-	if !registrar.HasFilter(ctx, name) {
+	exists, err := registrar.HasFilter(ctx, name)
+	if err != nil {
```
amit-momin, line 87: Maybe this doesn't matter so much since the ChainReader isn't really used anymore, but since it hasn't been officially deprecated as far as I know, I just wanted to point out that this may change some upstream behavior. Since we would previously just return nil even on HasFilter failure, unregister loops like [this](https://github.com/smartcontractkit/chainlink-solana/blob/995f1f057c765de62176ffa387a725bf8e64eba2/pkg/solana/chainreader/bindings.go#L418-L420) would continue and call unregister on all readers. With this change, we'll exit early so we may need to adjust that loop to gather all errors and return only after the loop finishes.

Unheilbar (dismissed):

silaslenihan (approved):

amit-momin (approved):

PR #1624, opened 2026-09-02
Bound Slice and array length in idl codecs

Unheilbar:
### Description

The IDL codecs build every vec and bytes field as encodings.NewSlice(codec, builder.Int(4)). The upstream slice decoder reads that 4-byte count and passes it straight to reflect.MakeSlice, checking only that it is non-negative — so it allocates for the declared count before discovering the data isn't there. A few bytes of untrusted input allocate 2 GB for vec<u8>, up to 64 GB for vec<publicKey>.

Adds commoncodec.NewBoundedSlice, which wraps the length-prefix codec so a count whose implied footprint exceeds the 10 MiB Solana account limit is rejected while decoding the prefix, before MakeSlice runs. The bound uses the larger of the element's on-wire and in-memory size — a nested vec is 4 bytes on the wire but a 24-byte slice header in memory.

Wire format is unchanged and asserted byte-identical against the unbounded codec. The cap equals the whole account limit, so legitimate data — which must be smaller than its containing account or log — is never rejected.

### DIFF
--- a/pkg/solana/codec/common/bounded_size.go
+++ b/pkg/solana/codec/common/bounded_size.go
@@ -48,8 +48,12 @@ func CheckElementCount(count int, element encodings.TypeCodec) error {
 		return fmt.Errorf("%w: element codec must be non-nil", types.ErrInvalidConfig)
 	}
 
-	if limit := maxElementCount(element); count < 0 || count > limit {
+	limit := maxElementCount(element)
+	if count > limit {
 		return fmt.Errorf("%w: element count %d exceeds the maximum of %d for solana data", types.ErrInvalidConfig, count, limit)
+	}
+	if count < 0 {
+		return fmt.Errorf("%w: element count %d cannot be less than zero", types.ErrInvalidConfig, count)
 	}
 
 	return nil
--- a/pkg/solana/codec/common/bounded_size.go
+++ b/pkg/solana/codec/common/bounded_size.go
@@ -99,8 +99,10 @@ func (b *boundedSize) Decode(encoded []byte) (any, []byte, error) {
 	if !ok {
 		return nil, nil, fmt.Errorf("%w: %T is not an int indicating the size", types.ErrInternal, value)
 	}
-
-	if count < 0 || count > b.maxCount {
+	if count < 0 {
+		return nil, nil, fmt.Errorf("%w: element count %d cannot be less than zero", types.ErrInvalidConfig, count)
+	}
+	if count > b.maxCount {
 		return nil, nil, fmt.Errorf("%w: element count %d exceeds the maximum of %d for solana data", types.ErrInvalidEncoding, count, b.maxCount)
 	}
 
--- a/pkg/solana/codec/common/bounded_size.go
+++ b/pkg/solana/codec/common/bounded_size.go
@@ -1,0 +1,134 @@
+package commoncodec
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/codec/encodings"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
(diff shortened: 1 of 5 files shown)


dhaidashenko (dismissed):

silaslenihan (dismissed):

ilija42 (dismissed):

pkg/solana/codec/common/bounded_size.go
```diff
@@ -0,0 +1,134 @@
...
+// can fit within MaxAccountBytes.
+//
+// The bound is taken over both the on-wire and the in-memory footprint of an
+// element, whichever is larger. They diverge for elements without a fixed size:
+// a nested vector occupies only a 4 byte length prefix on the wire but a 24 byte
+// slice header in memory, so sizing on the wire footprint alone would allow an
+// allocation several times over MaxAccountBytes.
+func maxElementCount(element encodings.TypeCodec) int {
+	elementBytes := minVariableElementBytes
+	if fixed, err := element.FixedSize(); err == nil && fixed > 0 {
```
amit-momin, line 31: fixed could never be less than minVariableElementBytes right? Just making sure `fixed > 0` shouldn't be `fixed > elementBytes` here

pkg/solana/codec/common/bounded_size.go
```diff
@@ -0,0 +1,134 @@
...
+// MaxAccountBytes. Use it where the count is known at construction time, such as
+// a fixed array length taken from an IDL; slices bound their wire length through
+// NewBoundedSize instead.
+func CheckElementCount(count int, element encodings.TypeCodec) error {
+	if element == nil {
+		return fmt.Errorf("%w: element codec must be non-nil", types.ErrInvalidConfig)
+	}
+
+	if limit := maxElementCount(element); count < 0 || count > limit {
+		return fmt.Errorf("%w: element count %d exceeds the maximum of %d for solana data", types.ErrInvalidConfig, count, limit)
```
amit-momin, line 53: Think this error message isn't completely accurate to the clause. It could also fall here if count is negative

Unheilbar: true

pkg/solana/codec/common/bounded_size.go
Unheilbar: It can be less than minVariableElementBytes for 1-2 bytes primitives/ enums/ and fixed arrays

pkg/solana/codec/common/bounded_size.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, nil, err
+	}
+
+	count, ok := value.(int)
+	if !ok {
+		return nil, nil, fmt.Errorf("%w: %T is not an int indicating the size", types.ErrInternal, value)
+	}
+
+	if count < 0 || count > b.maxCount {
+		return nil, nil, fmt.Errorf("%w: element count %d exceeds the maximum of %d for solana data", types.ErrInvalidEncoding, count, b.maxCount)
```
amit-momin, line 106: Same here for the error message

pkg/solana/codec/v2/solana.go
```diff
@@ -380,6 +376,13 @@ func asArray(parentTypeName string, idlArray *idltype.Array, refs *codecRefs) (c
 		return nil, err
 	}
 
+	// The array length comes straight from the IDL, which is caller supplied on the
+	// log poller path, and encodings.array.Decode materialises the whole array. Bound
+	// it by the Solana account limit so an oversized IDL cannot force a huge allocation.
+	if err = solcommoncodec.CheckElementCount(lenInt, codec); err != nil {
+		return nil, fmt.Errorf("array length defined in IDL type is excessively large: %w", err)
```
amit-momin: Think you should just return err here without wrapping. Technically the err could be negative count which wouldn't make sense with this message

amit-momin (approved):

PR #1631, opened 2026-09-04
Bump solana-go to 1.23 and resolve breakings

Unheilbar:
core ref: 1294f58cb0ddeed3535c12339b08dadd7cc2ce3b
### Why
Picks up upstream fixes/features (v1 transaction support, account state helpers, etc.) and lets us drop a local stopgap.

### What had to change
- **`client.go`**: `MaxSupportTransactionVersion` raised to `1` — solana-go can now decode v1 messages natively, so the old `MaxRequestedTransactionVersion` workaround (requesting a higher version than we could decode, to avoid `getBlock` failing with `-32015`) is no longer needed.
- **`logpoller`**: removed that stopgap (`job_get_block.go`, deleted `job_get_block_version_test.go`, trimmed related metrics).
- **`client.go`**: solana-go's `Transaction.MarshalBinary` now silently zero-pads missing signatures instead of erroring, so `SendTx` explicitly calls `VerifySignatures()` first and returns a new `ErrInvalidSignatures`; `classify_errors.go` classifies it as `Fatal`.
- **`client.go`**: replaced `commonhttp.LimitedTransport` with a local `sizeLimitedTransport` — solana-go's JSON-RPC client now decodes with `goccy/go-json`, whose stream reader swallows non-EOF errors from the response body, so the size-limit error from the old lazy (read-time) transport never surfaced. The new transport enforces the limit eagerly in `RoundTrip`, before decoding starts.
- **`chainwriter`**: fixed `TestChainWriter_SubmitTransaction` — solana-go's `AccountMeta.less()` now tie-breaks same signer/writable-tier accounts by raw pubkey bytes instead of preserving insertion order, so `NewTransaction`'s compiled `AccountKeys` are no longer in program-declared order. This doesn't affect execution correctness (each instruction's account references are still resolved by pubkey via `CompiledInstruction.Accounts`), but the test asserted on fixed indices. Updated it to check account membership/order via the compiled instruction indices instead of raw `AccountKeys` position.


pkg/solana/client/client.go
```diff
@@ -26,22 +28,14 @@ const (
...
-// rather than omitting that transaction, which stalls the poller on that slot until the job
-// exhausts its retries. Requesting a higher version returns the block in full; transactions
-// this repo cannot decode are then skipped individually.
-//
-// This is a stopgap. The vendored solana-go only understands legacy and v0 messages: its
-// Message.UnmarshalWithDecoder routes anything with the version bit set to UnmarshalV0, so a
-// newer transaction would be parsed with the v0 layout rather than rejected. Until that gains
-// real support, decoding stays capped at MaxSupportTransactionVersion.
-const MaxRequestedTransactionVersion = uint64(1)
+// Bump this once solana-go supports decoding a newer message version.
```
amit-momin: nit: think this comment is stale now since this PR is adding support

pkg/solana/client/client.go
```diff
@@ -494,6 +528,14 @@ func (c *Client) SendTx(ctx context.Context, tx *solana.Transaction) (sig solana
 	done := c.latency("send_tx")
 	defer func() { done(err) }()
 
+	// solana-go's MarshalBinary pads any missing signatures with zero bytes rather than
+	// erroring, so an under-signed transaction would otherwise wire-encode as if it were
+	// fully signed and only fail asynchronously once the cluster processes it.
+	if verr := tx.VerifySignatures(); verr != nil {
+		err = fmt.Errorf("%w: %w", ErrInvalidSignatures, verr)
+		return solana.Signature{}, err
+	}
```
toblich, line 537: Is this related to the sdk bump, or is it a gap that we always had? It looks correct to have this anyway, but I wonder if it's safe to add and what downstream effects it might have.

pkg/solana/logpoller/job_get_block_test.go
```diff
@@ -375,6 +376,56 @@ func TestGetBlockJob(t *testing.T) {
...
+		require.NoError(t, err)
+		job := newGetBlockJob(nil, client, make(chan types.Block, 1), lggr, slotNumber, metrics, nil)
+
+		err = job.Run(t.Context())
+		require.NoError(t, err)
+
+		result := <-job.blocks
+		require.Equal(t, slotNumber, result.SlotNumber)
+		require.Equal(t, &block.Blockhash, result.BlockHash)
+		require.Empty(t, result.Events) // no "Program log:"/"Program data:" lines in this tx, so no events expected
```
toblich, line 421: I'll try to look for (or craft) a v1 tx with some events and inner cross-program invocations, so we can test the event parsing and traversal of inner instructions too

toblich (commented):
Looking good, just a couple comments

pkg/solana/client/client.go
Unheilbar: yeah this gap has always been here, somehow got revealed with solana airdrop constantly failing after sdk bump

toblich (approved):

amit-momin (approved):

PR #1636, opened 2026-09-08
add v1 support to fee estimator

Unheilbar:
core ref: 1294f58cb0ddeed3535c12339b08dadd7cc2ce3b


pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,66 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
+	// filter out consensus vote transactions
+	// consensus messages are included as txs within blocks
+	// validate AccountKeys has enough elements to index into ProgramIDIndex
+	if len(baseTx.Message.Instructions) == 1 &&
+		len(baseTx.Message.AccountKeys) > int(baseTx.Message.Instructions[0].ProgramIDIndex) &&
+		baseTx.Message.AccountKeys[baseTx.Message.Instructions[0].ProgramIDIndex] == solana.VoteProgramID {
+		return true
+	}
+
+	return false
```
toblich: Nitpick: you can `return` the condition directly
```suggestion
	return len(baseTx.Message.Instructions) == 1 &&
		len(baseTx.Message.AccountKeys) > int(baseTx.Message.Instructions[0].ProgramIDIndex) &&
		baseTx.Message.AccountKeys[baseTx.Message.Instructions[0].ProgramIDIndex] == solana.VoteProgramID
```

pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,66 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
+func parsePriceFromTransactionV1(baseTx *solana.Transaction) (price ComputeUnitPrice) {
+	computeUnitLimit := baseTx.Message.TransactionConfig.ComputeUnitLimit
+	priorityFee := baseTx.Message.TransactionConfig.PriorityFee
+	if computeUnitLimit == nil || *computeUnitLimit == 0 || priorityFee == nil {
+		return price
+	}
+
+	price = ComputeUnitPrice(*priorityFee * 1_000_000 / uint64(*computeUnitLimit))
+	return
+}
```
toblich: Nitpick: This reads a bit odd with the usage of `price` as a named return - something like the following seems clearer IMHO
```suggestion
func parsePriceFromTransactionV1(baseTx *solana.Transaction) ComputeUnitPrice {
	computeUnitLimit := baseTx.Message.TransactionConfig.ComputeUnitLimit
	priorityFee := baseTx.Message.TransactionConfig.PriorityFee
	if computeUnitLimit == nil || *computeUnitLimit == 0 || priorityFee == nil {
		return ComputeUnitPrice{}
	}

	return ComputeUnitPrice(*priorityFee * 1_000_000 / uint64(*computeUnitLimit))
}
```

pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,66 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
+			// err returned if not SetComputeUnitPrice instruction
+			if parseErr == nil {
+				price = parsed
+				break
+			}
+		}
+	}
+
+	return
+}
```
toblich: Nitpick: same thing here about the named return, you can simplify by just removing the named value
```suggestion
func parsePriceFromTransactionV0(baseTx *solana.Transaction) ComputeUnitPrice {
	for _, instruction := range baseTx.Message.Instructions {
		// find instructions for compute budget program
		// validate AccountKeys has enough elements to index into ProgramIDIndex
		if len(baseTx.Message.AccountKeys) > int(instruction.ProgramIDIndex) &&
			baseTx.Message.AccountKeys[instruction.ProgramIDIndex] == ComputeBudgetProgram {
			parsed, parseErr := ParseComputeUnitPrice(instruction.Data)
			// if compute unit price found, break instruction loop
			// only one compute unit price tx is allowed
			// err returned if not SetComputeUnitPrice instruction
			if parseErr == nil {
				return parsed
			}
		}
	}

	return ComputeUnitPrice{}
}
```

pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,66 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
+	// consensus messages are included as txs within blocks
+	// validate AccountKeys has enough elements to index into ProgramIDIndex
+	if len(baseTx.Message.Instructions) == 1 &&
+		len(baseTx.Message.AccountKeys) > int(baseTx.Message.Instructions[0].ProgramIDIndex) &&
+		baseTx.Message.AccountKeys[baseTx.Message.Instructions[0].ProgramIDIndex] == solana.VoteProgramID {
+		return true
+	}
+
+	return false
+}
```
toblich, line 91: Nitpick: you can just return the condition
```suggestion
func isConsensusVoteTX(baseTx *solana.Transaction) bool {
	// filter out consensus vote transactions
	// consensus messages are included as txs within blocks
	// validate AccountKeys has enough elements to index into ProgramIDIndex
	return len(baseTx.Message.Instructions) == 1 &&
		len(baseTx.Message.AccountKeys) > int(baseTx.Message.Instructions[0].ProgramIDIndex) &&
		baseTx.Message.AccountKeys[baseTx.Message.Instructions[0].ProgramIDIndex] == solana.VoteProgramID
}
```

toblich (commented):
LGTM, though some coding nitpicks only tiny things to simplify. Also, would you mind adding one unit test for each method? E.g. using the base64 of a real tx from devnet, check that each price parser returns the right price, and that vote txs are correctly identified as such.

amit-momin (dismissed):

pkg/solana/fees/utils.go
toblich: This nitpick is still valid (though not a reason to block/delay the PR if you'd rather not do it)

toblich (dismissed):
LGTM

pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,64 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
-				// err returned if not SetComputeUnitPrice instruction
-				if parseErr == nil {
-					price = parsed
-					break
-				}
-			}
+		var price ComputeUnitPrice
+		switch baseTx.Message.GetVersion() {
+		case solana.MessageVersionLegacy, solana.MessageVersionV0:
+			//handle here
```
cll-gg: Nitpick: drop the comment or make it more helpful

pkg/solana/fees/utils.go
```diff
@@ -61,33 +61,64 @@ func ParseBlock(res *rpc.GetBlockResult) (out BlockData, err error) {
...
+func parsePriceFromTransactionV0(baseTx *solana.Transaction) ComputeUnitPrice {
+	for _, instruction := range baseTx.Message.Instructions {
+		// find instructions for compute budget program
+		// validate AccountKeys has enough elements to index into ProgramIDIndex
+		if len(baseTx.Message.AccountKeys) > int(instruction.ProgramIDIndex) &&
+			baseTx.Message.AccountKeys[instruction.ProgramIDIndex] == ComputeBudgetProgram {
+			parsed, parseErr := ParseComputeUnitPrice(instruction.Data)
+			// if compute unit price found, break instruction loop
+			// only one compute unit price tx is allowed
+			// err returned if not SetComputeUnitPrice instruction
```
cll-gg, line 102: The comment says `err` is returned but it doesn't look like it is

Unheilbar: so this code 
```go
			parsed, parseErr := ParseComputeUnitPrice(instruction.Data)
			// if compute unit price found, break instruction loop
			// only one compute unit price tx is allowed
			// err returned if not SetComputeUnitPrice instruction
			if parseErr == nil {
```
means that ParseComputeUnitPrice returns error if no SetcomputeUnitPrice instructions found. then we just continue iteration. It doesn't imply that the outer func body returns error

cll-gg: Ah I see, I read that the other way around 👍

cll-gg (approved):

